PR #9, opened 2019-06-04
Extension controller

Bouh:
## Describe the extension
Add events for connect multiples controllers inside GD5

## Checklist

- [x] Extension has a proper name and description.
- [x] Extension has tags (for example: "platform, brick, breakable").
- [x] All behaviors have a description.
- [x] All functions (actions, conditions, expressions) have descriptions.
- [x] I confirm that this extension can be intergrated to this GitHub repository, distributed and MIT licensed.

## Extension file
See PR


Bouh:
Hello !
Here is the extension for the controllers, the most basic events work. 
The code can be improved if you know how to do it dont hesitate !

Look the functions, test its, stress the code, i authorise you to torture my code ^^

There's still work to be done, but I'm gonna need some help.

I want to add an expression that returns the last button pressed for a specific controller. 
But as you can see I record the last pressed key that is listed in the events. 
In fact, if I press X and it doesn't matter in the game, the key will not be the last key pressed.

Wend1go:
I don't even get a "plugged" signal.
The Steam Controller worked in the old controller demo though.

Bouh:
Oh i've a problem with global variables i fix it now :| 
I will upload a small project example too for fast testing, sorry

Bouh:
I've fixed the freezes.
You can download this project for test.
[My project27.zip](https://github.com/4ian/GDevelop-extensions/files/3254351/My.project27.zip)

4ian:
Great, let me know when it's ready for review :)
First thing that I just saw, initialize a not found `gamepads` with null:
`var gamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads : null);`

because otherwise, if you do `if (gamepads) { ... }`, an empty array (`[]`) will be considered as true. Always use `null` to represent "objects that are not valid and should not be used".

Also nitpicking but important for consistency, prefer to rename gamepads to "getGamepads" (because that's the name of the original function):

`var getGamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads : null); `

not a big deal, but make things clearer in the rest of the code.

Bouh:
I don't know if he's ready for the review, but you can go.

I had trouble declaring a global variable, I created an object as I could.

I wanted to make an expression to know the last button used, but for that I would need a behavior and it makes it all less easy for user. Currently the expression returns the key only if it is used in events.

I have a switch conditionnal too, it seems to me that people don't like this way of doing things, but that's all I know in this case.

4ian:
I'll check tonight :)

> I wanted to make an expression to know the last button used

Yeah, you would need some kind of functions that are automatically run at
every frame. Will think about this.

> I have a switch conditionnal too

It's fine :) It's a good use case for a switch what you did here from what
I see.
`switch` is just less used and sometimes people argue against it but it's
mostly when it's over-abused in object oriented programming.

4ian:
I took a look, nothing bad but can you check these things (sorry that's a long list ;) But I want this to become an official extension):
- `var getGamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads : []);` should be `var gamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads() : []);`
  - `var getGamepads` should be renamed `var gamepads` because it's the result (i.e: an array), not a function.
  - You forgot to actually call webkitGetGamepads: `navigator.webkitGetGamepads()` (with parenthesis) to call the function, otherwise you will be storing the function inside `gamepads` (so that would not work on browsers implementing `webkitGetGamepads`).

- This is useless:
```
if (getGamepads == null) {
    //console.log("Gamepad is not supported in this browser");
    return;
}
```

because getGamepads will ALWAYS be an array according to the previous line. An array is always evaluated to true, even when empty. Too solutions:
  - You keep this condition and replace `[]` by `null` (i.e: you put null in the variable in gamepads can't be loaded). It's clearer
  - You keep `[]` and remove entirely the `if getGamepads == null`. This is shorter.

Up to you :)

- Don't do `var something = new Object({ blabla: thing })`, use directly an object literal: `var something = { blabla: thing }`.
  - In the same way, instead of `= new Object()`, just do `= {}`.
  - This has almost the same meaning in JS and object litteral is shorter and more idiomatic.
- GDevelop guarantees that if you ask for a number in an argument, then you'll get one. Same for strings. So:
  - `eventsFunctionContext.getArgument("stick").toString()` can just be `eventsFunctionContext.getArgument("stick")`
  - `Number(eventsFunctionContext.getArgument("player_ID") - 1)` => the call to number is useless, just do `eventsFunctionContext.getArgument("player_ID") - 1`
- Optionnal => Optional
- In E_Axis_factor, you do:
```
v /= (1.0 - DEADZONE);
if (v == null) {
            return 0;
        } else {
            return v;
        }
```

but v can never be null, because you assign it to a number. So just do:
```
  return v / (1 - DEADZONE);
```

(1.0 is unnecessary is JS, and you don't need to mutate v, you can directly return the result of the division).
- In `A_Set_deadzone`, validate the input of the user:
```
var newDeadzone = eventsFunctionContext.getArgument("param_deadzone");
if (newDeadzone < 0 || newDeadzone >= 1) return;

window._extensionController.deadzone = eventsFunctionContext.getArgument("param_deadzone");
```
- Would be better to attach _extensionController to `gdjs`: `gdjs._extensionController =` because
  - `window` is a concept only existing in browsers. If we use a standalone JS engine without `window` concept (like in Node.js), this will crash
  - Polluting the global namespace is dangerous. A future library/rendering engine/game engine/whatever could define its own _extensionController and this would break the whole thing. Prefer to attach things to gdjs, so that all the GDevelop game engine is "scoped" to `gdjs`.

- `var setDeadzone = function (v)`
  - The name setDeadzone implies that you're mutating something, while it's actually returning a number. Rename it `getDeadzone` or something more descriptive: `getNormalizedAxisValue` (the operation you're doing is called in mathematics normalizing a value from between [x..y] to [a..b])

> // Remap v from
> //    DEADZONE..1 to 0..(1-DEADZONE)
> // or from
> //    -1..-DEADZONE to -(1-DEADZONE)..0

This is very useful, it helped me to understand what you were doing 👍

- You can slightly simply this:
  ```
Object.keys(getGamepads).forEach(key => {
    if (getGamepads[key]) {
        eventsFunctionContext.returnValue = Number(key) + 1;
    }
});
  ```

into: `eventsFunctionContext.returnValue = Object.keys(gamepads).filter(key => !!gamepads[key]).length;`
(explanation: first I've renamed `getGamepads` to `gamepads`, because it does not make sense to iterate on something that is called like a function ;) 
keys allow to get the gamepad keys in an array. Then for each key, we apply a filter: if the `gamepads[key]` is defined, this will return true and keep the key in the array. Otherwise, if `gamepads[key]` is not defined, this will return false and remove the key from the array. At the end we count the number of keys in the array.

This is actually safer to do so, because if gamepads contain controllers that do not start from 0:
```js
gamepads = { // This is just for illustration
"1": { ...some controller },
"4": { ...some other controller }
}
```

then your solution will return 5, instead of 2.

- You can simplify:

```
if (getGamepads[player_ID] != null) {
    if (getGamepads[player_ID].id != null) {
```

into `if (gamepads[player_ID] && gamepads[player_ID].id) {`


If you don't understand some of these, still see if you can fix by yourself what you can, because I have no gamepads to test. I'll fix the rest myself. But again as I don't have a gamepad, I don't want to break anything and prefer to have you start changing these before I do some changes :)

4ian:
I know it's a lot of feedback and can take a bit of time to fix but this will make the code shorter in most places, and more robust - so less users complaining about potential game crashes ;)
It's also nice for you as if you have these pitfalls, you'll have a nice level in JavaScript :)

Bouh:
- Oops the () of webkitGetGamepads, it's fixed!
- I replaced [] by null because it's clearer, and I also deleted the following condition, I was afraid it would block the game if gamepads were empty.

[...]
Few minutes later.

- I've made the corrections, on the whole I understood all the explanations, it's pretty clear 👍 

- Just to make sure I understood what you said with this code : 
`
eventsFunctionContext.returnValue = Object.keys(gamepads).filter(key =>!!gamepads[key]).length;
`

`gamepads[key] `
returns the object of one of the controllers, as it is an array it returns true to filter(). Because an array even empty always returns true

`!gamepads[key] `
If the object does not exist, it returns false 

`!!gamepads[key]`
Reverse the whole thing so in the end everything is always true, unless the object does not exist

Bouh:
- For the vibrations I can't have them because I can't find vibrationActuator in electron (see my function).

- A user who had tested told he wish a condition to know when a key was released, but for that you would have to put everything in a behavior on an object, wouldn't you?
I would like to avoid a behavior so that the extension is only in events.

4ian:
> as it is an array it returns true to filter()

Precisely, an empty array is still an array, so filter will be called on nothing, so the result is an empty array (because nothing was filtered):
`[].filter(() => whatever)` will returns `[]`

in the worst case, filter returns an empty array: `myArray.filter(() => false)` will always return `[]`.

> Reverse the whole thing so in the end everything is always true, unless the object does not exist

Yes basically I'm filtering the array to keep only the object that are existing. Then I count (with length) the number of these objects.

> A user who had tested told he wish a condition to know when a key was released, but for that you would have to put everything in a behavior on an object, wouldn't you?
> I would like to avoid a behavior so that the extension is only in events.

Yes for that I guess we would need some special kind of functions that would be called after (or before) events - like behaviors. I'll think about adding that.

4ian:
I've made some updates:

* You actually have to use `[]` and not null for gamepads, otherwise you'll get a crash.
* Use gdjs._extensionController = gdjs._extensionController || { some_default: value } to initialize things
* Beware, when you do this:
```
var a = 1;
var b = {
  a: 2
};
```

the resulting object is `b = {"a": 2}`, not `b = {1:2}`. Use `b[a] = 2` instead. I fixed that.

Make sure to re-try everything to be sure I've not broken anything ;)

Bouh:
- E_Count_connected_controller
Error : gamepads.filter is not a function

- I shortened the  gamepad=null  condition on a line.

- C_Controller_type
I've added 
`/** @type {Gamepad[]} */`
and [] for Gamepads

I prefer delete A_Vibrate_controller because we can't use atm, because electron have not vibrationActuator.

I will let you fix gamepads.filter, and delete A_Vibrate_controller if you're agree with me.
And it's good for me ! I can't wait to get feedback from users :D

4ian:
> Error : gamepads.filter is not a function
>I've replace with the previous code

Indeed, I got mistaken by the description that "navigator.getGamepads returns an array [...]" while actually it is an object that *looks like* an array (damned strange APIs!). So your solution is perfectly right :)

I've:

* Updated most descriptions
  * Notably moved the descriptions of buttons on Xbox, PS4, other to the description of the condition, where multiple lines can be entered.
* Removed the useless parameter of E_Get_Deadzone
* Renamed in sentences "player XXX" by "gamepad XXX". For example: "The gamepad identifier: 1, 2, 3 or 4" for the description of the parameters asking for the id of a gamepad.
  * The idea is that the metaphor we should use is the gamepad. A player is generally associated to a gamepad, but we can imagine multiple gamepads being used by only one player (or players changing gamepads). So I think it's better/clearer to always talk about gamepads in the extension.
* I've renamed the expression so that they have a name that is nicer to read in the events sheet than "E_something_something". 
  * The idea is that you would name them like a "get" function: "GetGamepadTriggerPressure"
  * By convention, for expression we don't put "Get" because it's implicit.
  * As the name of the extension is always put before the expression name, it's safe to omit Gamepad: "TriggerPressure". Pressure is at the end because it's what you get as a result.

> I prefer delete A_Vibrate_controller because we can't use atm, because electron have not vibrationActuator.

It's actually implemented in other game engines so let's have it. I'll add a comment that this might not work on all platforms.

Bouh:
Yeahh it's good !
A_Vibrate_controller work on Chrome, but not in Edge, and not in Firefox, Chrome is, so far, the only browser that supports vibration, tested and approved.

4ian:
Merging :) I'll set version to 0.0.1
Note that we should be free to do some updates that are not backward compatible, because extensions anyway are copied into games when installed.
Also as this will become used a lot probably, we might think about a help page for it on the wiki?

4ian:
Might be a good idea too to update the existing gamepad/controller example to remove everything and use the extension instead :)

Bouh:
I will update the example, and start a page on wiki

4ian:
Sounds good! Examples are very useful, and after the page is created I'll add a field to link an extension to a help page (so that help button is showing like other actions/conditions).

Issue #11, opened 2019-06-13
Mobile keyboard


Bouh:
## Describe the extension
This extension add a behavior for allow to use keyboard mobile in your games !
I've created a example, you can play it on[ itch.io here.](https://bouh.itch.io/keyboard-mobile-gd5)
All is downloadable bellow 👇

Event added : 
- Open keyboard
- Close keyboard

## Checklist

- [x] Extension has a proper name and description.
- [x] Extension has tags (for example: "platform, brick, breakable").
- [x] All behaviors have a description.
- [x] All functions (actions, conditions, expressions) have descriptions.
- [x] I confirm that this extension can be intergrated to this GitHub repository, distributed and MIT licensed.

## Extension file
[Extension_keyboard_mobile_JSON.zip](https://github.com/4ian/GDevelop-extensions/files/3283846/Extension_keyboard_mobile_JSON.zip)

[Example_Keyboard_mobile.zip](https://github.com/4ian/GDevelop-extensions/files/3283842/Keyboard_mobile.zip)

4ian:
Nice! I've just quickly checked, this works well on my phone.
I've not checked the JS entirely but:

- "Always" is not necessary in the conditions.
- If I enter text, dismiss the keyboard and click to open the keyboard again, the text is added at the beginning of the string. Instead, we should move the selection at the end.
- `document.getElementsByTagName("input")[0];` will always select the first input - which might not even be an input from the game. Instead, we should probably rely on an input created with a unique (~random) classname or id - and store this id in the text entry. Then, whenever you want to open/close keyboard, pass the id to the functions so that they can retrieve the input :)
  - If input can't be found (null or undefined), ignore the rest to avoid a crash.

Bouh:
I've two other issues :
- The keyboard in an infinite state of opening and closing very rapidly. Even after closing chrome. Device required restart.  
Mobile Huawei Nova Lite 2, Android 8

- It writes backwards everything
It buts a new letter left side of the old letter so it writes text like.
I've typed : 1234
Text display : 4321
Samsung Galaxy S7
Samsung Galaxy XCover 4, Android 8.1

4ian:
> Opera : keyboard hover game, and game don't fit screen

Not sure if it's worth spending time fixing things in Opera Mobile - its usage is really low, not sure how good it is in standards implementation and we should more generally aim for Chrome on Android and Safari on iOS.

> position: fixed;z-index: -1;top: -30px;

[...]

PR #15, opened 2019-07-17
Add Mobile virtual keyboard for Text Entry behavior

4ian:
Made by @Bouh in https://github.com/4ian/GDevelop-extensions/issues/11

Let's continue the discussion here :)
I've cleaned the extension (double check though that it's still working, because I found something fishy in the closeKeyboardById function => it was refering to the eventsFunctionContext not passed as parameter, which must never be done!).

An example was included in https://github.com/4ian/GDevelop-extensions/issues/11
Ideally, I would like to see how to bundle examples and even documentation - to be sure that it's easy to understand, use and maintain these custom behaviors.

### DIFF
--- a/Extensions/TextEntryVirtualKeyboard-header.json
+++ b/Extensions/TextEntryVirtualKeyboard-header.json
@@ -1,0 +1,13 @@
+{
+  "shortDescription": "Allow to open and close the virtual keyboard on mobile (iOS, Android). To be used with Text Entry objects.",
+  "extensionNamespace": "",
+  "fullName": "Mobile virtual keyboard for Text Entry",
+  "name": "TextEntryVirtualKeyboard",
+  "version": "0.0.1",
+  "url": "Extensions/TextEntryVirtualKeyboard.json",
+  "headerUrl": "Extensions/TextEntryVirtualKeyboard-header.json",
+  "tags": "javascript, virtual keyboard, mobile",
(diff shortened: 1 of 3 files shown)


Bouh:
When we switch between two scenes the input HTML is created again and again. Otherwise it works well

4ian:
This might be due to the ownerRemovedFromScene not being called, I'll investigate.

4ian:
FYI, I'll fix this in GDevelop by making sure that onOwnerRemovedFromScene is properly called when a scene is unloaded.

4ian:
In the next version, onDestroy (which is the new name of `onOwnerRemovedFromScene` - but existing behavior with onOwnerRemovedFromScene will continue to work) will be properly called when 1) The object is removed from the scene or 2) The scene is unloaded :)

After this, we should be able to merge this, without issue with HTML inputs being recreated without being destroyed! :)

4ian:
This should be better now in beta 76

Bouh:
> This should be better now in beta 76

Yes, perfect. Here new files ready for bundle.
[Keyboard_mobile.zip](https://github.com/4ian/GDevelop-extensions/files/3449163/Keyboard_mobile.zip)

4ian:
Merged! I did some changes/clean up before merging, though this was done before your latest zip file. Can you check the version that is available online and tell me if there are changes to be made? 

Thanks for making this! Will be super useful :)

PR #23, opened 2019-10-15
Fix freeze with vibration on Gamepads extensions

Bouh:


4ian:
Good catch, always check for the validity of objects after getting them from somewhere :)

PR #28, opened 2019-12-15
Add the Debug Console extension

arthuro555:
By @arthuro555.
Permits to inject a text entry object into any scene that executes js functions or gd actions when a predefined keyword (command) is entered, to help debugging. Almost Entierly in JS.

### DIFF
--- a/Extensions/Gamepads-header.json
+++ b/Extensions/Gamepads-header.json
@@ -3,7 +3,7 @@
   "extensionNamespace": "",
   "fullName": "Gamepads (controllers)",
   "name": "Gamepads",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "url": "Extensions/Gamepads.json",
   "headerUrl": "Extensions/Gamepads-header.json",
   "tags": "controllers, gamepads, joysticks, axis, xbox, ps4",
--- a/Extensions/ConsoleExtension-header.json
+++ b/Extensions/ConsoleExtension-header.json
@@ -1,0 +1,13 @@
+{
+  "shortDescription": "An Extension that permits to create some kind of dev console",
+  "extensionNamespace": "",
+  "fullName": "Console Extension",
+  "name": "ConsoleExtension",
+  "version": "1.0-beta",
+  "url": "Extensions/ConsoleExtension.json",
+  "headerUrl": "Extensions/ConsoleExtension-header.json",
+  "tags": "command, console, debugging",
+  "eventsBasedBehaviorsCount": 0,
+  "eventsFunctionsCount": 8,
+  "description": "An Extension that permits to create some kind of dev console. It also contains some default commands."
+}
--- a/Extensions/DebugConsole.json
+++ b/Extensions/DebugConsole.json
@@ -1,0 +1,299 @@
+{
+  "author": "arthuro555, contact through the forum or through discord. You can always ping me ;-)",
+  "description": "An Extension that permits to create some kind of dev console. It also contains some default commands.",
+  "extensionNamespace": "",
+  "fullName": "Console Extension",
+  "name": "ConsoleExtension",
+  "shortDescription": "An Extension that permits to create some kind of dev console",
+  "tags": "command, console, debugging",
+  "version": "1.0-beta",
+  "eventsFunctions": [
+    {
+      "description": "Activate the console. Use this just before the user begins to input commads.",
+      "fullName": "Activate Console",
+      "functionType": "Action",
+      "name": "ActivateConsole",
+      "sentence": "Activate console.",
+      "events": [
+        {
+          "disabled": false,
+          "folded": false,
+          "type": "BuiltinCommonInstructions::Comment",
+          "color": {
+            "b": 109,
+            "g": 230,
+            "r": 255,
+            "textB": 0,
+            "textG": 0,
+            "textR": 0
+          },
+          "comment": "This is only to add the text input extension",
+          "comment2": ""
+        },
+        {
+          "disabled": false,
+          "folded": false,
+          "type": "BuiltinCommonInstructions::Standard",
+          "conditions": [],
+          "actions": [
+            {
+              "type": {
+                "inverted": false,
+                "value": "TextEntryObject::Activate"
+              },
+              "parameters": [
+                "kk",
+                "yes"
+              ],
+              "subInstructions": []
+            }
+          ],
+          "events": []
+        },
+        {
+          "disabled": false,
+          "folded": false,
+          "type": "BuiltinCommonInstructions::JsCode",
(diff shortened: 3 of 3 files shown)


arthuro555:
apparently also updates headers for another extension 👀

4ian:
Nice, let me take a look tomorrow 👍

Extensions/Gamepads-header.json
```diff
@@ -3,7 +3,7 @@
   "extensionNamespace": "",
   "fullName": "Gamepads (controllers)",
   "name": "Gamepads",
-  "version": "0.0.1",
```
arthuro555: That was not intentional and Auto-Updated by the script

4ian: No worries that's fine!

4ian:
Interesting extension :) Thanks for you work on this so far!

I've browsed the extension and listed my thoughts so far. 

* The most important thing: I've seen that you're creating dynamically an object and a behavior. 
This is risky because relying currently on internals (`_behaviors`) that might change without warning, breaking the extension. Also, I'm worried things won't work properly across scene (you're creating a singleton object for all the game, but what happen if I try to switch to another scene).

Instead, shouldn't this extension provides:
* A behavior, that you can attach to one (or more? :)) TextEntry objects.
* A set of actions/conditions/expressions, either on the behavior or (if really really reaaaally necessary) action/condition that are not tied to the behavior but reading/storing information into a global object (`gdjs.evtTools.console`).

Other feedbacks in no particular order:

* The description of the extension needs to be at least describing how to use this because it's quite a complex extension to start with :) Give an example: "To get started, add a Text Entry object and blabla. Then use the action blabla. Finally, use condition blabla and blabla".

* Avoid incomplete words in expression names:
* `AllDefltCommnds` => `AllDefaultCommands`
* Start expression name with a capital:
  * `getTextBuffer` => `GetTextBuffer`
  * `getArg` => `GetArgument`
  * `getArgs` => `GetArguments`
* Prefer to start the name of conditions with a verb: 
  * EventSheetCommand => * `IsCommandInputted` or: `IsCommandEntered`

I've found some mistakes in the text/names/descriptions:
* "Use this just before the user begins to input commads." => commands
* "Command inputed" => "Command inputted" (or "Command entered").
* "Save the args from the last call of a command to a struct" (and others) => "structure". Struct is a C keyword, in GDevelop we prefer structure to make it clear.

Some things that could also be reworded:
* "Activate Console" => "Activate the Console" (same for deactivate)
* "Get an argument from the last call of a command" => "Get an argument passed to the last command"
* "Get one of the arguments given in the last call of a command" => "Get one of the arguments passed to the last command".
* "If a command is inputed, do..." would be better written with the do, like other conditions: "Check if a command is inputted".

* Use booleans to store data rather than numbers:
  * `window.extErrorDefaultCommandJSON = 1;` and similar => put `true` instead.

* I've seen some console.logs, can you remove them if not useful?
image

* Can you fix the `Behavior type "undefined" was not found.` warning? The Debugger console should stay blank, extensions should in general not be displaying anything apart from fatal errors.

Some JS stuff:

* `var match = /\r|\n/.exec(this._str);` => This creates a regex everytime this function is creating. I think we can avoid this (regex can be complex to create) by just checking `this._str.indexOf("\r") !== -1 || this._str.indexOf("\n") !== -1`.

* For:
``` 
if (match){
        return true;
    } else {return false};
```

write instead: `return !!match;`

* `let variab = runtimeScene.getVariables().get(eventsFunctionContext.getArgument("var"));` => Don't use abbreviation. Use plain, simple and descriptive words. Make a code that is simple and stupid :)
`const variable = runtimeScene.getVariables().get(eventsFunctionContext.getArgument("variable"));`

Also in 90% of the case you can use `const` instead of `let` if you're not re-assigning the variable - which you probably almost never want to do.

* `window.extErrorArgs` and co: that's a lot of global variable. Murphy's Law tell use that this *will* at some point create issue for some obscure reasons. Instead, I would suggest storing them in `gdjs.evtTools.console` (you can do `gdjs.evtTools.consoleUserErrors = gdjs.evtTools.consoleUserErrors || {};` first thing in functions to make a "global" object on gdjs.evtTools and use it to store your errors (a bit dirty.. but less than using window).

arthuro555:
Wow! I am happy to see you really did go through all of it and are interested! The fact that a behavior and an object are "injected" is on purpose. I did think of a behavior but I think it is easier to use and more convenient when you don't have to create an object and everything, and It helps to keep the scene of the visual editor "cleaner". It's also partly because this entire project was a test to try some stuff to see if it is possible (this being the main one). Excuse all misconceptions, they are due to the fact that this was originally little challenge (coding it in a day) for me and I didn't thought things throughly before implementing them. For the internal variables, I will check the GDevelop version (if it changes some day) and use the correct attribute, but if a getter is added I will use it. I am sorry for the DEbUg output I forgot to remove, I will do it for the next commit. For the error, I actually have no idea where it came from. I don't have my code in front of me right now to check what might cause that but I don't remember putting more than one addObject call. And I only use JS since a month or so so excuse my mistakes commig from my inexperience in JS specific things.

4ian:
> Probably not as you are against changing internal variables so more object "injection" would be bad.

I'm worried of things that touch internal methods/variables because, while JS allow you to do this, you should not be able in theory to access them. This reduces the agility of the code and increase the risk of me (or someone else unaware of your extension) doing changes to some code, which will break the extension (creating frustrated user, frustrated developer, and confusion for everyone "why did is this software so bugged?!?").

Everything that is private (`_something` in JS) should be avoided and dynamically creating objects is already a bit of "dark magic" ;) 

> I had an idea to select an object to use a command like "getObjects (object name)" to get an array with all Instances, then a next command that changes a Sprite "selector" (I imagine it as a red circle) to the x,y coords of each objects by updating an index and if it becomes out of range comes back to 0. It would be an easy way to select an object through commands

For this, I would recommend taking a look at making an action/condition that takes objects as parameter.
Having objects as parameter allow to make flexible extensions, because you can compose the objects rather than relying on the extension to create objects or "find" them (which is implicit, so error prone and hard to discover).
Instead, having an action "Update the selectors using objects XXX" (XXX being objects that you pass as parameter) is surely more robust, let the user choose which objects it want to be used, more explicit (it's written in the events sheet). A bit more of set up sure, but extensions are there to be composed together :) 

>  I did think of a behavior but I think it is easier to use and more convenient when you don't have to create an object and everything, and It helps to keep the scene of the visual editor "cleaner"

A bit the same comment, reconsider strongly behaviors. In the future, we might have "prefabs" objects coming with behaviors or things like this. Having your extension being a "standard" extension as any other will be helpful and avoid lots of headache and confusion.

arthuro555:
I'll potentially rework that I e day but not anytime soon. Closing this old PR.

PR #29, opened 2020-01-08
Add condition "Any button is pressed"

Bouh:
Add condition "Any button is pressed"
Tested and operational
Gamepads version now is 0.0.3


4ian:
Could you explicitly set `eventsFunctionContext.returnValue = false` before the `return`, so that it's extra clear that the return value is false when no button is found :) 

Also, avoid doing `if (!gamepad.buttons[btn_id])` if btn_id is undefined. This can create subtle bugs because you're accessing the member called "undefined":

image

Typescript/Flow would yell at you for trying this ;) JavaScript is not erroring but that's a mistake.

Instead, just do `if (!btn_id) { eventsFunctionContext.returnValue = false; return;  }`. In other words: if a variable is not guaranteed to be defined (or can be null, which is almost the same), always test it before using it.

Bouh:
if i do 
`if (!btn_id) { eventsFunctionContext.returnValue = false; return; }`
Because btn_id can be 0 the condition return false. 0 is the key for A or Cross

4ian:
You're right, I forgot 0 is a valid value! In this case better be extra explicit and check for `undefined` then:
if (btn_id !== `undefined`)

(always use triple equal)

`if (typeof btn_id != 'number')` is also right. I just prefer to avoid checking the type because this give the impression to the reader of the code that this thing (btn_id) can for some reasons be something else than a number, so that we need a check for it... while we know precisely what it can be: either a number or undefined. So we check for undefined, and after we're sure that it's a number (Flow/Typescript would help by checking this for us).

Bouh:
not 
`if (btn_id !== "undefined")`
but
if (btn_id === "undefined")

We wish return false if the value is undefined, not return false if value is not undefined ^^

4ian:
Almost there, I still made a mistake 😬
`if (btn_id === undefined)`

No quotes, undefined is a keyword.

4ian:
There we go 👍

PR #32, opened 2020-01-19
Add Gamepad button released

Bouh:
Add Gamepad button released.
How are manage the actual release for keyboard / mouse.
Silver-Streak gave me feedback :

Here are the states on keyboard:
Key never pressed during scene: Not treated as released.
Key pressed with "Trigger once": Not treated as released
Key released with "Trigger once": Treated as released one frame
Two frames after key released with Trigger Once: Not treated as released
Key pressed again: Not treated as released

Here's how it works currently on the extension:
Button never pressed during scene: Not treated as released.
Button pressed with "Trigger once": Not treated as released
Button released with "Trigger once": Treated as released
Two frames after button released with Trigger Once: Treated as released
Button pressed again: Not treated as released


4ian:
Thanks for the PR :)

> Two frames after button released with Trigger Once: Treated as released

That's strange 🧐 I've not looked at the code for now though!

4ian:
Changes look good! Though I've not seen the "Gamepad button released" condition itself - did I miss something?
Still merging because the other changes look good 😊

PR #33, opened 2020-01-23
[Gamepads] Upgrade the extension with more instructions and rework the structure.

Bouh:
Sorry my previous PR didn't include the condition button released.
Here the new version :)

-------
Add Gamepad button released.
How are manage the actual release for keyboard / mouse.
Silver-Streak gave me feedback :

Here are the states on keyboard:
Key never pressed during scene: Not treated as released.
Key pressed with "Trigger once": Not treated as released
Key released with "Trigger once": Treated as released one frame
Two frames after key released with Trigger Once: Not treated as released
Key pressed again: Not treated as released

Here's how it works currently on the extension:
Button never pressed during scene: Not treated as released.
Button pressed with "Trigger once": Not treated as released
Button released with "Trigger once": Treated as released
Two frames after button released with Trigger Once: Treated as released
Button pressed again: Not treated as released


4ian:
Can you explain the logic behind `gamepad.buttons[btn_id].pressed === false && gamepad.buttons[btn_id].value === 0`?

I'm wondering if for this kind of case, we would rather need to have the state of the button at the previous frame (was the button pressed at the previous frame?) and the compare it with the state now (is the button pressed now?)
The logic would be "condition is true if button pressed at the previous frame and the button is not pressed now".

This would imply that I add a support for a special kind of function that would be always executed after or before events (or both) - a bit like behaviors but for events.
That I'm a bit afraid that could be abused to implement events there that would better go in a behavior or in a function. But hopefully not because you would not have access to any object or parameter - so it would be only useful for these kind of state checking.

What do you think? Let me know how it works now so that we can decide if this is reliable as a first step :)

Bouh:
- GDJS haven't a method for know if we are in a other frame that thre previous ?

- If this case require a feature only for this extension, (because we haven't not other case) we can forget this.
Maybe we are at some point where this extension should be include as extension in the engine?

- `gamepad.buttons[btn_id].pressed === false && gamepad.buttons[btn_id].value === 0` 
Are almost same, buttons.pressed is explicit, and buttons.value is the value property is a floating point value used to enable representing analog buttons, such as the triggers on many modern gamepads. The values are normalized to the range 0.0 representing a button that is not pressed, and 1.0 representing a button that is fully pressed.

Bouh:
-New thing in this commit! You can use the expression for get string of last button pressed.
This can work only if an condition use also same input. I should make a behavior for make this function but it's not logic to add a behavior on an object for use gamepad extension.

So yeah maybe Functions need a cycle function called each frame like behavior?

- I've also fixed the issue with state of button released. I take core of last state on previous frame for see what changed and if button was pressed at last frame and is not press at current frame i return true for this condition!

4ian:
> So yeah maybe Functions need a cycle function called each frame like behavior?

Yes that might be something to add.

> I take core of last state on previous frame for see what changed and if button was pressed at last frame and is not press at current frame i return true for this condition!

That sounds like a proper solution, let me take a look at this tonight :)

4ian:
I've seen usage of `getElapsedTime` - why is the time used here? The extension should probably not rely on this at all.

Judging by the number of time I see:
image

I think we'll have to actually had special functions that are executed at the beginning of the game, beginning of scene, scene unload, scene pause, scene resume and before and after events.

I'm worried that if not, what should be a very simple code is actually being hard to read because there are tons of initialization done again and again, leading to code that is not elegant and hard to maintain.

4ian:
This being said, appreciate a lot your work on the extension :) One of the feature often asked for GDevelop. Hopefully these new lifecycle functions for extensions will make your life easier.

Bouh:
Code need to be cleaned. I was thinking use getElapsedTime() for see if i'm in same frame, but i've found something easier :)

I will read again the functions for see what can be simplify. 
And if you work on this card i will wait before improve other functions.

4ian:
I'll try to come up with something in the next days for the lifecycle functions :) 

JS events might need more work to be more comfortable yeah, I won't have time to look at that though for now.

4ian:
@Bouh Do you have time to update this to use the new lifecycle functions? Should clean a lot the code.

Bouh:
I've started but not finished :/
Need to clean code.
Need to finish WIP functions actions and expression for remap the inputs.

All other thing not tagged wip should work, i've tested this two months ago.
I've rework deadzone for each players instead global deadzone for all players. And yes used lifecycle 
 functions.

I can take back on this, I'm tired of coding my shader for merge two textures. basic are present but i need rest,  uvs and matrix are new and complex for me.
This extension will be more simple and it's almost ready to use.

Here the current state
[Gamepad_0.0.4.2b-Extension.zip](https://github.com/4ian/GDevelop-extensions/files/4434538/Gamepad_0.0.4.2b-Extension.zip)

4ian:
Sure, I think we're not too far from having something here that is ready and really useful. I'm thinking of making some extensions like this one highlighted in GDevelop so that they are easy to discover.

Bouh:
- I use now lifecycle function for define the object used in the extension and catch the last key pressed and more stuff.
- All functions have been reworked for working with the new object define in onFirstSceneLoaded.
- New instruction and expressions:
  - Key is released.
  - Any key pressed.
  - Any key released.
  - Last pressed button (string) 
  - Deadzone expression and setter have a new parameter (i've invert the order, this break the compatibility on previous version.)

- Gamepad vibration work only on web export, not in preview, not in windows exports. Not working on electron in general (this was also the case in previous version, not tested on electron v8 ).
- I've added some functions in WIP, it is some ideas, let me know what you think.


[gamepad_benchmark_project_file_with_extension.zip](https://github.com/4ian/GDevelop-extensions/files/4465644/gamepad_benchmark.zip)

4ian:
> All functions have been reworked for working with the new object define in onFirstSceneLoaded.

Sounds good.

> I've added some functions in WIP, it is some ideas, let me know what you think.

I see it's a lot about mapping. It's interesting but I would recommend to keep stuff simple, maybe we could create another extension "Input Mapper" that could work to map inputs to arbitrary strings, and could work with touch/mouse, keyboard and even the Gamepads extension - if detected :)

> If needed i can write a migration guide for include it in the description of the extension when the user download this new version. Like a patchnote for extension.

That would be interesting in the future. For now, I think it's ok because it's probably not yet used a lot, so it's fine to "break compatibility". People can also just not update and stay and the previous version.

To distribute the new version, I think we should remove the WIP expressions though (and be sure they don't start with E_). That's another improvement: being able to hide expressions/actions/conditions that would not be ready or deprecated.

Bouh:
> That's another improvement: being able to hide expressions/actions/conditions that would not be ready or deprecated.

Dont waste your time on it. Clean the extension is simplier.

> That would be interesting in the future. For now, I think it's ok because it's probably not yet used a lot, so it's fine to "break compatibility". People can also just not update and stay and the previous version.

Ok then i'll rename all my functions, change the order in parameters, avoid to see gamepad::E_Last_button_ID()
It's ugly i can write a beautiful, gamepad::getLastButtonID().


> I see it's a lot about mapping. It's interesting but I would recommend to keep stuff simple, maybe we could create another extension "Input Mapper" that could work to map inputs to arbitrary strings, and could work with touch/mouse, keyboard and even the Gamepads extension - if detected :)

Oh yeah a other extension for that is a good idea!

4ian:
> Clean the extension is simplier. [...]
> Ok then i'll rename all my functions, change the order in parameters

Sounds good 👍

> It's ugly i can write a beautiful, gamepad::getLastButtonID().

Yep! Speaking of this, as all expressions are getter, the convention in GD is to not put "Get" also, everything in PascalCase (`Gamepad::LastButtonID()`).

Look forward to having this extension updated 👍

Bouh:
Ready to review!

4ian:
Looks mostly good :)

A few nits:

* In TriggerPressure: `//and if we delete this condition the game freeze` => not useful to have this comment (it's understood that if the condition to check `gamepad` for nullity is removed, then the rest of the code using it like `gamepad.buttons[6].value` will crash).
* `in expression: Value of a gamepad axis` and other `in condition: blabla`. Put quotes around the conditions or expression name: `in expression "Value of a gamepad axis".`, ``in condition " blabla"`
* `Return the index of the last button used on a gamepad` => `Return the index of the last pressed button of a gamepad`
* LastButtonString: 
  * `Mapping arrays for ech type of controllers` => `Map associating controller button ids to button names`
  * `map_of_controllers` => you redefine this (`const map_of_controllers = {`) at each call of `LastButtonString`. It's creating a new object for each call, which will increase pressure on the JavaScript garbage collector (especially on low end devices). Prefer to set up a variable in your extension object in `onFirstSceneLoaded`:
  ```
  gdjs._extensionController = {
    players: { ... },
    controllerButtonNames: { "XBOX": {...}, ... }
  }; 
  ```

  (also note the name: "controllerButtonNames". It's an object containing the names of the buttons of the controllers).

* C_Controller_type: 

```
if (gamepads[player_ID] != null) {
    if (gamepads[player_ID].id.toUpperCase().indexOf(controller_type) != -1) {
        eventsFunctionContext.returnValue = true;
    } else {
        eventsFunctionContext.returnValue = false;
    }
}
```

You can slightly simplify the code by avoiding branching (ifs) with ternary and assignement.
=> 

```
const gamepad = gamepads[player_ID];
eventsFunctionContext.returnValue = gamepad ? gamepad.id.toUpperCase().indexOf(controller_type) !== -1 : false;
```

Might be a matter of taste, but using an intermediary variable "gamepad" allow to avoid risky repetition of "player_ID". Use a ternary (`gamepad ? .... : false`) to put the value to use in case of gamepad being null.

* Test if the specified gamepad => Check if the specified gamepad
* `Useful for know if gamepad is Xbox or PS4 controller.` => `Useful to know if gamepad is a Xbox or PS4 controller.`. Use "to" if the next word is a verb.

* Gamepad informations => Gamepad information. Information has no plural in English (https://www.grammar-monster.com/plurals/plural_of_information.htm), no s.

We should be good to go ahead with this extension then! :)
Thanks for working on this will be super useful 👍

Bouh:
- In which case i should use `Check ` instead `Test`or `Verify `in the description of the conditions ?

- Using ternary and assignement in the same line it's cool, very compact, it's a habit to take, i keep it in my mind.

- I've found a bug with the condition `Any gamepad button released`, the state of the button at last frame stay true indefinitely.

Bouh:
I've done what you suggest and change these:

- I've fixed all conditions and sentences for console.error(). (One is an warn, deadzone)
- Deadzone cannot have scientific notation anymore.
- Some if are reduce with the conditionnal ternary.  (♥‿♥) 
- "Any gamepad button released" is now finish, WIP tag is removed.

4ian:
Great!! 👍 We should be not too far from having this ready. A bunch of comments still :) 

Potential issue in A_Set_deadzone:
* `newDeadzone = Number.parseFloat(gdjs.evtTools.common.clamp(newDeadzone, 0, 1)).toFixed(3);`
  * You're parsing a number - the return value of clamp is already a number. But then you're using toFixed() to convert it back to a string... knowing that newDeadzone is previously a number ;) Be careful about the types of your variables - changing them at runtime is potential a performance issue and is only allowed because JavaScript is loosely typed. Let's inspect things:
    1. `newDeadzone` is a number. This is guaranteed by the parameter configuration: 
image
    2. This means that in theory `if (newDeadzone == null) return;` can never be true, newDeadzone is always a number.
    3. This means that `newDeadzone = Number.parseFloat(gdjs.evtTools.common.clamp(newDeadzone, 0, 1)).toFixed(3);` is clamping it (unnecessary because you had a check before), parsing it (but it's a number ;)), converting it to a string (`toFixed`) and storing it in `gdjs._extensionController.players[playerId].deadzone` 

This means that we can greatly simplify the whole code, from:

```
//Get function parameter
const playerId = eventsFunctionContext.getArgument("player_ID") - 1;
var newDeadzone = eventsFunctionContext.getArgument("deadzone");

if (playerId < 0 || playerId > 4) {
    console.error('Parameter gamepad identifier in action: "Set gamepad deadzone for sticks", is not a valid, must be between 0 and 4.');
    return;
}
if (newDeadzone < 0 || newDeadzone > 1) {
    console.warn('Parameter deadzone in action: "Set gamepad deadzone for sticks", is not valid, the value has been automatically revamp beetween 0 and 1.');
    return;
}

//Deadzone is not define
if (newDeadzone == null) return;

//Clamp the newDeadzone 0-1, parseFloat and toFixed avoid to have scientific notation. like 5e-7 = 0.0000005
newDeadzone = Number.parseFloat(gdjs.evtTools.common.clamp(newDeadzone, 0, 1)).toFixed(3);

//Save the deadzone value in the appropriate object for the player
gdjs._extensionController.players[playerId].deadzone = newDeadzone;
```

to:
```
const playerId = eventsFunctionContext.getArgument("player_ID") - 1;
const newDeadzone = eventsFunctionContext.getArgument("deadzone");

if (playerId < 0 || playerId > 4) {
    console.error('Parameter gamepad identifier in action "Set gamepad deadzone for sticks", is not valid, must be between 0 and 4.');
    return;
}

gdjs._extensionController.players[playerId].deadzone = gdjs.evtTools.common.clamp(newDeadzone, 0, 1);
```

Nitpicking/descriptions:
* `Test if a button is pressed on a gamepad` => `Check if a button is pressed on a gamepad`. Prefer "check" when it's a boolean.
* Everywhere: `is not a valid` => `is not valid`.
* `Return the number of gamepads connected` => `Return the number of connected gamepads`
* GamepadType. Full name `Gamepad information` can just be `Gamepad type`? (not sure). Otherwise consider renaming this to `GamepadInformation`.
* C_Controller_Type: 
  * Full name `Gamepad information` can just be `Gamepad type`? (not sure)
  * `Useful for know if gamepad is Xbox or PS4 controller.` => `Useful to know if the gamepad is a Xbox or PS4 controller.`
* LastButtonString: careful about your descriptions, you're missing a period. `Return the last button pressed buttons for Xbox and PS4 can be:` => `Return the last button pressed. Buttons for Xbox and PS4 can be:`
* LastButtonString: You're calling `gdjs._extensionController.getInputString(controller_type, gdjs._extensionController.controllerButtonNames, lastButtonUsedID)` passing `gdjs._extensionController.controllerButtonNames`. Don't pass this argument, put it in the function directly:
```
gdjs._extensionController.getInputString = function (type, buttonId) {
    const controllerButtonNames = gdjs._extensionController.controllerButtonNames;
    if (controllerButtonNames[type] !== undefined) {
        return controllerButtonNames[type][buttonId];
    } 

    return "UNKNOWN_BUTTON";
}
```
  * nit: prefer camelCase for variables.
  * `Mapping not exist` => `UNKNOWN_BUTTON` (don't return sentences if the developer is expecting the name of a button, that is usually in UPPER_CASE. Instead, return a string describing a fake button named "UNKNOWN_BUTTON".
  * In the same way, `"Not on Xbox"` => `NONE` (a button called "NONE" should be fairly explicit that it's a button that is not existing on the controller).

* C_Axis_pushed: super nitpicking: `let getNormalizedAxisValue = gdjs._extensionController.getNormalizedAxisValue` => `const getNormalizedAxisValue = gdjs._extensionController.getNormalizedAxisValue`

Bouh:
```
gdjs._extensionController.players[playerId].deadzone = gdjs.evtTools.common.clamp(newDeadzone, 0, 1);
```

Then the scientific notation still exist. 
The value can be 5e-7 = 0.0000005, and worst with more zero i've already seen 27 zero. We dont need this precision. 2 or 3 are enough.
Then i can do:

```
newDeadzone = Number(gdjs.evtTools.common.clamp(newDeadzone, 0, 1).toFixed(3));
```

>Test if a button is pressed on a gamepad => Check if a button is pressed on a gamepad. Prefer "check" when it's a boolean.

Keyboard condition use Test ^^, but ok.

4ian:
> Then the scientific notation still exist.
> The value can be 5e-7 = 0.0000005, and worst with more zero i've already seen 27 zero. We dont need this precision. 2 or 3 are enough.

Where do you see it? There is no problem for such a number to exist in memory right? 

> newDeadzone = Number(gdjs.evtTools.common.clamp(newDeadzone, 0, 1).toFixed(3));

You can truncate the number, in which case you can do `(gdjs.evtTools.common.clamp(newDeadzone, 0, 1) * 1000 | 0) / 1000`. (1000 for 3 decimal places, `| 0` for truncating to an integer).
This avoids converting to a string and then having to parse a number, which is time consuming, while operation on numbers are very fast.

But again not sure what's the issue. If the user wants to set a dead zone of 0.000003, well it's weird but we should store this in memory, no?

> Keyboard condition use Test ^^, but ok.

Yes you're right that there are still a lot of conditions using Test :) I think that "Check" sounds better and is more natural english than test. I also prefer "Compare" when the test is done a number or strings.

Bouh:
mmmmm

Look the event under the preview, and the keys pressed.

Bouh:
It's a strange behavior from Javascript 
image

Bouh:
with your code i can get 
image
After 6 up key pressed.

4ian:
> It's a strange behavior from Javascript

Exactly, to be very precise it's a problem with floating point number representation that is not limited to JS but to most languages as this is how numbers are represented in memory. See this interesting article: https://medium.com/better-programming/why-is-0-1-0-2-not-equal-to-0-3-in-most-programming-languages-99432310d476

Hence why I think it's better to not to any rounding :) It sadly won't solve everything (on the other hand, it can add confusion: "why is my deadzone truncated, did I do anything wrong??") and I'm not sure it can be solved nicely apart from giving ways to display a number nicely as a text, i.e: by truncating it after some decimal places (something that also should be done in GDevelop UI).

Bouh:
Interesting!
Then i have removed the code.

4ian:
I merged by myself in #33! (did a few typo fixes and re-run the script to update the registry).
Thanks for doing this update and for your patience, it's getting really good 👍

PR #36, opened 2020-04-20
Fix bullet not firing on first shot

paulera:
Added an onCreated function to reset the timer "FireCooldown" of the
Object as soon as it is created.

Without this, the first attempt to shoot creates the timer, but doesn't
shot because "FireCooldown" seconds haven't passed yet.


arthuro555:
Consider bumping the version and regenerating the header ;)

4ian:
Looks good! Thanks for opening the PR. I bumped the version, ran `node update-extensions-registry.js` and applied the changes.

Merging now 👍👍

PR #37, opened 2020-04-30
fix gamepad

Bouh:
const player_ID ....
was declare but playerId was used.
So this fix use playerId like in other functions.


4ian:
Thanks! Can you run the node script in `scripts` folder?

PR #39, opened 2020-06-04
Add numeric expression for get angle and force on sticks

Bouh:
Full name in the instruction editor:
Value of a stick rotation

Description:
Get the rotation value of a gamepad stick.
If the deadzone value is high, the angle value is rounded to main axes, left, left, up, down.
An zero deadzone value give a total freedom on the angle value.

Expression used: 
Gamepads::StickRotationValue(1, "LEFT")
1 is player gamepad
left is stick, can be right too.


Bouh:
- Added a new expression for get the force on sticks.
- Update header, version and new features.

Bouh:
There is no issue broken change.
Ready for review.

In next days i'll update all examples using my gamepad extension for use this new version 0.0.6

4ian:
Nice! :) 
I'll have a look.

4ian:
Merged! Thanks a lot 👍👍

I think we could show this extension as a "recommended" extension or highlighted one in GDevelop :)

PR #51, opened 2020-09-20
[Sine extention] Support negative value in amplitude for inverse the sin wave.

Bouh:
The amplitude X and Y can now be negative for inverse the sin wave.


4ian:
I guess why not! :)

PR #53, opened 2020-10-14
added the calc extention

Add00:
This is first pull request, so sorry if I did something wrong.

### DIFF
--- a/Extensions/calc.json
+++ b/Extensions/calc.json
@@ -1,0 +1,333 @@
+{
+  "author": "add_",
+  "description": "A collection of 7 calculations functions to help find slope, gcf and many other values.\n\nEnjoy :)",
+  "extensionNamespace": "",
+  "fullName": "calculations",
+  "name": "calc",
+  "shortDescription": "A collection of 7 calculations functions to help find slope, gcf and many other values.",
+  "tags": "calculations, math",
+  "version": "1.1.0",
+  "eventsFunctions": [
+    {
+      "description": "Calculates the steepness of a line between two points",
+      "fullName": "Slope",
+      "functionType": "Expression",
+      "name": "slope",
+      "sentence": "",
+      "events": [
+        {
+          "disabled": false,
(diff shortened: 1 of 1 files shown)


Extensions/calc.json
```diff
@@ -0,0 +1,333 @@
+{
+  "author": "add_",
+  "description": "A collection of 7 calculations functions to help find slope, gcf and many other values.\n\nEnjoy :)",
```
arthuro555, line 3: I would stay formal, and remove the enjoy (GDevelop is for everyone, also professionals ;) )

Extensions/calc.json
```diff
@@ -0,0 +1,333 @@
...
+  "name": "calc",
+  "shortDescription": "A collection of 7 calculations functions to help find slope, gcf and many other values.",
+  "tags": "calculations, math",
+  "version": "1.1.0",
+  "eventsFunctions": [
+    {
+      "description": "Calculates the steepness of a line between two points",
+      "fullName": "Slope",
+      "functionType": "Expression",
+      "name": "slope",
```
arthuro555, line 15: I think names should be capitalized, not 100% sure though (wait for bouh or 4ian to confirm)

Add00: No your right they told me the same thing when I submitted the noise extension. So I'll go fix that.

Bouh:
One by one the extensions please.
Create separate pull requests for each extensions.

The main issue is in the [wiki page.](http://wiki.compilgames.net/doku.php/gdevelop5/tutorials/procedural-generation/implementing-procedural-generation) about _procedural extension_ you use a **map** function.
This one isn't existing in the extension, so you have two choices.

- Add the function in the existing _procedural extension_ 
- Edit the current wiki and add the function is another extension called **Extended math support**.

This way the unfinished **Constants Extension** will be rename **Extended math support**, and you can include your constant math, and function math.

**Create separate pull requests for each extensions.**

Issue #1, opened 2019-05-18
New extension


4ian:
> Please note that this extension was originally created by @ddabrahim (thanks!) - creating this issue myself to give an example of how you can submit extensions :)

## Describe the extension
An action to move the camera by dragging with the mouse (or scrolling with the touch), on vertical, horizontal or both directions.

## Checklist

- [x] Extension has a proper name and description.
- [x] Extension has tags (for example: "platform, brick, breakable").
- [x] All behaviors have a description.
- [x] All functions (actions, conditions, expressions) have descriptions.
- [x] I confirm that this extension can be integrated to this GitHub repository, distributed and MIT licensed.

## Extension file

[DragCameraWithPointer.json.zip](https://github.com/4ian/GDevelop-extensions/files/3194273/DragCameraWithPointer.json.zip)

PR #62, opened 2020-10-27
Add extension Auto Typing

Bouh:
Continue here @4ian @westboy31.
@4ian the extension is good to me,  if you want check :)

Here an example running the extension.
[exemple_autotyping.zip](https://github.com/4ian/GDevelop-extensions/files/5448107/exemple_autotyping.zip)

### DIFF
--- a/Extensions/AutoTyping-header.json
+++ b/Extensions/AutoTyping-header.json
@@ -1,0 +1,13 @@
+{
+  "shortDescription": "This extention simulate typing text (useful for dialogue scenes or visual novals).",
+  "extensionNamespace": "",
+  "fullName": "Auto typing",
+  "name": "AutoTyping",
+  "version": "1.0.0",
+  "url": "Extensions/AutoTyping.json",
+  "headerUrl": "Extensions/AutoTyping-header.json",
+  "tags": "Text, BBText, Dialogue, Visual noval, Autotyping, typing",
+  "eventsBasedBehaviorsCount": 0,
+  "eventsFunctionsCount": 7,
+  "description": "* Apply auto typing action to a text object to start typing and chose interval between characters.\n* It automatically auto type again when the text content changes (but don't forget to trigger once).\n* In condition you can catch the event typing end.\n* You can also make pause to text typing and resume it after."
+}
--- a/Extensions/AutoTyping.json
+++ b/Extensions/AutoTyping.json
@@ -1,0 +1,853 @@
+{
+  "author": "westboy, daliyoucefmedakram@gmail.com",
+  "description": "* Apply auto typing action to a text object to start typing and chose interval between characters.\n* It automatically auto type again when the text content changes (but don't forget to trigger once).\n* In condition you can catch the event typing end.\n* You can also make pause to text typing and resume it after.",
+  "extensionNamespace": "",
+  "fullName": "Auto typing",
+  "name": "AutoTyping",
+  "shortDescription": "This extention simulate typing text (useful for dialogue scenes or visual novals).",
+  "tags": "Text, BBText, Dialogue, Visual noval, Autotyping, typing",
+  "version": "1.0.0",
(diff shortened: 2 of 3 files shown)


westboy31:
this is the same extention with correction of sentences :
[AutoTyping.zip](https://github.com/4ian/GDevelop-extensions/files/5459554/AutoTyping.zip)

I agree that it's better to use it as behavior , this is how it's look like : 
[AutoTyping_behavior.zip](https://github.com/4ian/GDevelop-extensions/files/5459602/AutoTyping_behavior.zip)

but we have the functions implemented two times (two behavior one for BBtext and the other for text) , the only way i see to avoid this is to use any_object in parameters but it could occasion bugs if user choses sprite or something else .

4ian:
> but we have the functions implemented two times (two behavior one for BBtext and the other for text) , the only way i see to avoid this is to use any_object in parameters but it could occasion bugs if user choses sprite or something else .

Yeah it's unfortunate to have to write the behavior twice, but it's no big deal and better this way :) 
I think having the behavior is a bit more intuitive (what do you think @Bouh?), thanks for doing it!

There are two thing that I'd like to understand:
1. When I change the text with a trigger once, it works ok :) But if I try to change the text *while* it's being typed, it starts to glitch, displaying the full text followed by some part of the text being typed.
  - See this example I made: [Example.json.zip](https://github.com/4ian/GDevelop-extensions/files/5464148/Example.json.zip)
  - This is because you're adding a character one by one. I think this is overcomplicated things here:
    image
    Instead, use `SubStr` to choose a portion of the text to write (which is _txt_buffer), starting at position 0 and finishing just after the write index (the `1 +` is important): 
    image

2. You're using a scene timer with the object name. This has the problem that this won't work with more than one object.
  - => Can you use a timer of the object rather? I think you overcomplicated this by using a scene timer. Unless I missed something, use an object timer called "Write timer" or something like "Autotyping Write Timer" (so you're sure no one else will use your timer :)) and it will work for multiple object independently.

3. Finally, even with the fix of `SubStr` to avoid glitches while changing the text, if we try to change the text while it's written, we have potential the text being flashed, before being replaced by the auto typed text the next time the timer reaches the next character to be added. 
  - How to solve this? If you're in the state of writing (i.e: _write_index is less than the length of the text to type), then always set the text. Then increment the _write_index only when the timer reaches the interval. Like this:
    image

With this, you have a behavior that:
- works even without a trigger once! In my example, I used a click/touch condition.
- the text won't stop or glitch before it's fully finished. If you try to change the text, it will only work once the typing is finished.
- with the timer fix, it should work for multiple objects

Let me know if you can do these three changes in both behaviors. This is my example with the extension modified to have the fix #1 and #3 done on the text object: 
[Example partially fixed.json.zip](https://github.com/4ian/GDevelop-extensions/files/5464298/Example.partially.fixed.json.zip)

If you can do the timer fixes, and do the equivalent changes in the BBText object, we should then be ready to publish this very nice extension! 👍👍

Bouh:
> I think having the behavior is a bit more intuitive (what do you think @Bouh?)

This is better indeed, the users will have no surprises with the startup action at least ^^
I still see people asking why the actions don't work, because the actions are in an sub-events with conditions doesn't work hehe.

westboy31:
Well seen it's way better like that:
[AutoTyping.zip](https://github.com/4ian/GDevelop-extensions/files/5465496/AutoTyping.zip)

exemple to check the whole extension:
[exemple.zip](https://github.com/4ian/GDevelop-extensions/files/5465527/exemple.zip)

4ian:
Thanks! I did a final update of some descriptions, it's now added (ac1da141ceb919bbc6e0beb4f5523b137162e669) and available :)

4ian:
Closing this now, thanks @westboy31 for this nice extension and @Bouh for your help!

Will be very very useful for a lot of games!

PR #69, opened 2020-11-22
Rework cursor extension

arthuro555:
- Adds a string selector as parameter
- Move some descriptions to long descriptions
- Add a behavior that applies the cursor appearance when hovering an object that has the Behavior. 

I am not sure how the "clean way" to access a behavior property via js is, but the way I used seems to work well so I guess it is the right one, though it uses a method with an underscore so I am not very sure about it.


arthuro555:
🤔 It seems I was using an old GDevelop version and it removed the icon now that I look at the diff... I'll fix that in a few minutes I got to go for now, don't merge yet though!

arthuro555:
Ok @Bouh and @4ian this is now ready for review!

Bouh:
Tested and approved.
And link in long description works too, very cool!

But before merging an extension with a string selector we should support expression in the feild, if we didn't support it and user has already added a expression this will not working again.

That's what I deduced from my test on the nightly build.
I would wait for this support too before updating the gamepad extension.

Extensions/CursorType.json
```diff
@@ -36,26 +36,125 @@
...
+          "type": "stringWithSelector"
         }
       ],
       "objectGroups": []
     }
   ],
-  "eventsBasedBehaviors": []
+  "eventsBasedBehaviors": [
+    {
+      "description": "Sets the cursor appearence on hover of the object.",
```
4ian, line 61: `appearance when the object is hovered`

Extensions/CursorType.json
```diff
@@ -36,26 +36,125 @@
...
+          "fullName": "",
+          "functionType": "Action",
+          "name": "onCreated",
+          "sentence": "",
+          "events": [
+            {
+              "disabled": false,
+              "folded": false,
+              "type": "BuiltinCommonInstructions::JsCode",
+              "inlineCode": "const renderer = objects[0].getRendererObject();\nrenderer.interactive = true; // Activate necessary pixi plugin\n",
```
4ian, line 117: Usually in the game engine we're never using PIXI `interactive` because I think it runs a separate timer and logic to check for hovering/collisions. More generally, we use PIXI only as a rendering engine. The exception is for the IDE where we use the interaction capabilities.

Could you see if you can rework the behavior to use `isCollidingWithPoint` or the equivalent condition checking if a point is inside an object?
You'll need to get the cursor position with MouseX, MouseY - don't forget to pass the layer of the object.

I would like to centralize everything that is related to collisions through the game engine - because later this can be reworked to use optimizations like spatial partitioning.

Extensions/CursorType.json
```diff
@@ -36,26 +36,125 @@
...
+          "fullName": "",
+          "functionType": "Action",
+          "name": "doStepPreEvents",
+          "sentence": "",
+          "events": [
+            {
+              "disabled": false,
+              "folded": false,
+              "type": "BuiltinCommonInstructions::JsCode",
+              "inlineCode": "objects[0]\n    .getRendererObject()\n    .cursor = objects[0]\n        .getBehavior(\n            eventsFunctionContext.getBehaviorName(\"Behavior\")\n        )\n        ._getcursor_type();\n",
```
4ian, line 77: It's ok to use `_getcursor_type`  here because the underscore is here to avoid external code, i.e: JS code of the scenes, to get access to the properties (if you want to expose a property, you can make an expression for it).
Here, we're inside the behavior, so we're allowed to access this "private" property.

4ian (commented):
Thanks for doing this :)
I've posted a comment about avoiding using PIXI capabilities in the behavior, I hope we can do without PIXI as this is something that should be handled by the game engine.

4ian:
> But before merging an extension with a string selector we should support expression in the feild, if we didn't support it and user has already added a expression this will not working again.

It will be more painful to edit indeed, though the expression will still work. We should definitely add support for editing with the selector button and maybe autocompletions... Actually we could make the choices to be autocompletions themselves! :)

Extensions/CursorType.json
arthuro555: > don't forget to pass the layer of the object.

I don't see where I should pass the object layer? Neither MouseX, MouseY or isCollidingWithPoint seem to require it. It seems to work anyways.

arthuro555:
I probably should use GDevelop global variables for the default and the zindex instead of js globals, and also put in in readable events... Or should I stay "consistent" by using JS everywhere?

4ian:
Having a mix of events and JS is not a problem at all - on the contrary :) I would not care as long as the "interface" (i.e: the action/condition) are clean for the end user :)

Might be a good thing to use global variables rather than putting stuff on `gdjs`, or at least put it in a "namespace" (`gdjs._yourExtensionWithALongNameToAvoidAnyConflicts.xxx`) (if you use global variables, might be a good idea to also use a structure to avoid clashes with other extension global variables... in the future we should really make extension scoped variables!).

Extensions/CursorType.json
4ian: The `gdjs.InputManager` getMouseX/getMouseY returns the position in the coordinates of the canvas.
`isCollidingWithPoint` is working with what we usually call "world coordinates", i.e: the real position of the objects.

The canvas coordinates and the world coordinates are only the same if your camera is not moved. As soon as your camera is moved/rotated/zoomed, they are not the same at all anymore :) 
Fortunately, we offer a way to translate from one to the other: [`convertCoords` on `gdjs.Layer`](https://docs.gdevelop-app.com/GDJS%20Runtime%20Documentation/gdjs.Layer.html#convertCoords)

Get the layer name of the object, get the layer from the runtimescene, call this with the mouse x/y position (and camera id 0) and send the returned coordinates to isCollidingWithPoint and you should be good. You can check that by moving the camera in your game and see that the object with the behavior still has a cursor :)

arthuro555:
Ok, I now use global variables and interchangeably use javascript and normal events.

Extensions/CursorType.json
4ian: ping on this

arthuro555: huh weird I could have sworn I had already fixed this

4ian: No big deal, I merged and made a few fixes in another commit :)

PR #72, opened 2020-12-01
Add extension Swipe

e1e5en-gd:
Added extensions that include Swipe behavior. This behavior attaches to any object and “handles the swipe event”.

Behavior settings:
- `Time (s)` - duration of the swipe (default value _0.2 s_);
- `Min Length (px)` - the minimum length that the swipe must “pass” (default value _100 px_).

Output values / functions:
- `IsDone` (_Condition_) - swipe done or not;
- `StartPointX`, `StartPointY` (_Expression_) - coordinates of the point where the swipe starts;
- `EndPointX`, `EndPointY` (_Expression_) - coordinates of the swipe end point;
- `Length` (_Expression_) - the length of the swipe;
- `DirectionX`, `DirectionY` (_Expression_) - values of the swipe direction vector;
- `DirectionNormX`, `DirectionNormY` (_Expression_) - values of the normalized vector of the swipe direction;
- `Angle` (_Expression_) - the value of the angle of the direction vector to the zero angle [-180, 180];
- `Average4Direction` (_String Expression_) - the direction value for 4 sides (_UP, DOWN, LEFT, RIGHT_) is displayed as a string. In this case, the “circle” is divided into 4 segments of 90 degrees;
- `Average8Direction`(_String Expression_) - the direction value for 8 sides (_UP, DOWN, LEFT, RIGHT, UP-LEFT, UP-RIGHT, DOWN-LEFT, DOWN-RIGHT_) is displayed as a string. In this case, the “circle” is divided into 8 segments of 45 degrees.

_Demo_: [web](https://e1e5en-gd.github.io/gdev-blog.github.io/swipe_ext/), [android](https://github.com/e1e5en-gd/gdev-extension-swipe/releases/download/1/swipe_ext.apk).
_Screenshot_:
image

_Usage example_:
image

### DIFF
--- a/Extensions/Swipe.json
+++ b/Extensions/Swipe.json
@@ -1,6 +1,6 @@
 {
   "author": "@e1e5en",
-  "description": "A behavior that detects a swipe being done on the screen, and that gives you information about it, including the direction and the length.\n\nTo begin, attach this behavior to *any object*.\n\nSettings are :\n- Time (s) - duration of the swipe (default value is 0.2s);\n- Min Length (px) - the minimum length that the swipe must have to be considered (default value 100 px).\n\nUse **the condition to check if a swipe** was done. Then you can use **the expressions** to get the swipe values:\n- `PointStartX`, `PointStartY` (Expression) - coordinates of the point where the swipe starts,\n- `PointEndX`, `PointEndY` (Expression) - coordinates of the swipe end point,\n- `Length` (Expression) - the length of the swipe, in pixels,\n- `DirectionX`, `DirectionY` (Expression) - values ​​of the swipe direction vector,\n- `DirectionNormX`, `DirectionNormY` (Expression) - values ​​of the normalized vector of the swipe direction,\n- `Angle` (Expression) - the value of the angle of the direction vector to the zero angle (0-360),\n- `GetDirectionsFor4Parties` (String Expression) - the direction value for 4 sides (UP, DOWN, LEFT, RIGHT). In this case, the “circle” is divided into 4 segments of 90 degrees,\n- `GetDirectionsFor8Parties` (String Expression) - the direction value for 8 sides (UP, DOWN, LEFT, RIGHT, UP-LEFT, UP-RIGHT, DOWN-LEFT, DOWN-RIGHT). In this case, the “circle” is divided into 8 segments of 45 degrees.",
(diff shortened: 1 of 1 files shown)


4ian:
Thanks! Looks pretty good! :) I've made some changes to the description and names of expressions and the behavior :)

For the Angle expression, this looks very complicated: 
image

I think it can be replaced by:
image

(Internally, AngleBetweenPositions is doing an `atan2(-y, x)`). What do you think? Can you try this? Thanks!

e1e5en-gd:
In your case, the angle is not returned in the range [0, 360]. That is why I then determine which quarter is "looking at the corner".

4ian:
Sure, it's returned in the range [-180, 180], which is equivalent :) You just need to check if the angle is between -45 and 45 (for right), or -135 and -45 (for up), 45 and 135 (for down) or greater than 135 or less than -135 (for left).
(i.e: 4 slices of 90 degrees)

It's simple and should work very well ;) Same for the 8 directions: instead of dividing [0, 360] in 8 slices, you divide [-180, 180] in 8 slices of 45 degrees each.

4ian:
Let me know if it make sense to you :) I think this can greatly simplify the angle calculation :) The range should not be a problem, and this will be very efficient.

e1e5en-gd:
I understand it :)
It's more convenient and easier for me, that's why I did it.
I was guided by the picture from [wiki](http://wiki.compilgames.net/lib/exe/detail.php/gdevelop5/behaviors/gdevelop_covaliantcuriosity.jpg?id=gdevelop5%3Abehaviors%3Atopdown).
I don't see much difference ... :)

e1e5en-gd:
I use the same formula to calculate the angle, the only difference is the rounding and conditions for such a range. This does not affect performance.
It's not difficult for me to correct it, I just don't see the difference :)

4ian:
Speaking of atan2, I have a question: in your current formula, I see this:
image

Why the multiplying between the coordinates (do you need the square of the direction?) and why multiplying X by Y in the first argument of atan2? Is this an error?

> It's not difficult for me to correct it, I just don't see the difference :)

There is no real difference overall :) I just think it can drastically simplify the `Angle` expression, because you just return the result of `atan2` (or `AngleBetweenPositions`) without having more work to do. 

Speaking of the formula again, I think we can also drop the `ceil` - someone could want to use the Angle of the swipe with the exact precision in the angle. What do you think? Users can always use `round` (or`floor` or `ceil`) if they need to have an angle without more precision than 1 degree.

e1e5en-gd:
> Is this an error?
Yes.

As a result, I will correct for the range [-180, 180] and remove the rounding. I will do it as time appears.

4ian:
Cool, ping me here when you have this updated 👍

e1e5en-gd:
Updated extension

e1e5en-gd:
Thanks!
The old range "(0-360)" remains in the description of the angle function.
Is it a separate PR to fix or will it be easier for you to do it?

4ian:
Nice catch, feel free to open a PR for it :)

tristanbob:
@e1e5en-gd great work!  This is a very useful extension that I am intending to use on an upcoming project.

PR #73, opened 2020-12-02
Update description

e1e5en-gd:
1. Changed the name in the function description:
- StartPointX, StartPointY;
- EndPointX, EndPointY.
2. Changed the range of angles in the Angle function description.


e1e5en-gd:
Updated the description to reflect the change in the extension in the "header" file

4ian:
Excellent 👍

PR #77, opened 2020-12-11
Fix swipe detector extension 

TheGemDev:
The description of the swipe detector extension displays a clickable text due to the use of  square brackets 
Screenshot (84)
The current solution is to change the brackets to a round bracket (if the brackets are necessary )
also clicking on the text displays the gdevelop quit alert  as seen below
Screenshot (85)


4ian:
Good catch! This is because this is parsed as markdown :) I've also fixed the issue of GDevelop being reloaded when clicked for next version.

PR #79, opened 2020-12-15
Add MQTT extension

arthuro555:
Fixes https://github.com/4ian/GDevelop/issues/2135

This is an MQTT client extension that adds pubs, subs, messages and connection to broker server.
This was tested with `test.mosquitto.org:8081` as broker.


4ian:
That was fast! :) Looks fair enough to me, I'll merge this now and will let you maintain/update/fix things as needed (as it's a very "niche" extension).

PR #86, opened 2020-12-20
Update clipboard extension

arthuro555:
Those changes should make the clipboard extension work on more platforms, by supporting the [browser APIs for clipboard](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard). 
As those are async, I had to make the read clipboard expression an action. I left the old expression for compatibility.


Extensions/Clipboard.json
```diff
@@ -1,34 +1,111 @@
...
+      "name": "WriteText",
+      "sentence": "Write _PARAM1_ to clipboard",
+      "events": [
+        {
+          "disabled": false,
+          "folded": false,
+          "type": "BuiltinCommonInstructions::JsCode",
+          "inlineCode": "const electron = runtimeScene.getGame().getRenderer().getElectron();\nconst text = eventsFunctionContext.getArgument(\"text\");\n\nif (electron !== null && electron.clipboard)\n  electron.clipboard.writeText(eventsFunctionContext.getArgument(\"text\"));\nif (\n  typeof navigator !== \"undefined\" &&\n  navigator.clipboard &&\n  navigator.clipboard.writeText\n) navigator.clipboard.writeText(text)\n  .catch(e => console.error(\"Error while writing clipboard: \", e));\nelse {\n  // TODO cache this element to not create elements on each call\n  const tempElement = document.createElement(\"textarea\");\n  tempElement.id = \"Clipboard_copy_temporary\";\n  tempElement.value = text;\n  tempElement.setAttribute('readonly', '');\n  tempElement.position = 'absolute';\n  tempElement.style.left = '-9999px';\n  document.body.appendChild(tempElement);\n  tempElement.select();\n  document.execCommand(\"copy\");\n  tempElement.remove();\n}; \n",
```
4ian, line 90: Is `document.execCommand` useful for a particular browser? Seems `navigator.clipboard` is supported by most browsers (https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) - apart from IE but we're not compatible with IE anymore.
If we can, let's not have the `document.execCommand` branch at all - simpler and better.

arthuro555: For me on google chrome, pasting requires execCommand, even though it said [here](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText) that `navigator.clipboard` works on my chrome version. I'd therefore keep it just in case.

4ian: It's supposed to work since Chrome 66. Can you double check? No way this thing is not working. You might have to accept a permission, which is something we won't be able to avoid (and that must be explained too in the action).

arthuro555: Nevermind, there was two factors that caused me to believe this:
1. The extension had a typo in `navigator`🤦‍♂️ 
2. For some security reason, on preview over wifi (which I use for testing in browser with hot reload), `navigator.clipboard` is undefined. As soon as I exported it and hosted it locally though, it worked.

Extensions/Clipboard.json
```diff
@@ -1,34 +1,111 @@
...
+      "fullName": "Write text to the clipboard (Windows, macOS, Linux)",
+      "functionType": "Action",
+      "name": "WriteText",
+      "sentence": "Write _PARAM1_ to clipboard",
+      "events": [
+        {
+          "disabled": false,
+          "folded": false,
+          "type": "BuiltinCommonInstructions::JsCode",
+          "inlineCode": "const electron = runtimeScene.getGame().getRenderer().getElectron();\nconst text = eventsFunctionContext.getArgument(\"text\");\n\nif (electron !== null && electron.clipboard)\n  electron.clipboard.writeText(text);\nif (\n  typeof navigator !== \"undefined\" &&\n  navigator.clipboard &&\n  navigator.clipboard.writeText\n) navigator.clipboard\n  .writeText(text)\n  .catch(e => console.error(\"Error while writing clipboard: \", e));\nelse console.error(\"Unable to write clipboard: No compatible method found!\"); \n",
```
4ian, line 90: `Unable to write to the clipboard: no method found for this platform.`

Extensions/Clipboard.json
```diff
@@ -1,34 +1,111 @@
...
+      "fullName": "Get text from the clipboard",
+      "functionType": "Action",
+      "name": "readTextCrossPlaform",
+      "sentence": "Read clipboard and store text in _PARAM1_",
+      "events": [
+        {
+          "disabled": false,
+          "folded": false,
+          "type": "BuiltinCommonInstructions::JsCode",
+          "inlineCode": "const electron = runtimeScene.getGame().getRenderer().getElectron();\nconst callback =\n    runtimeScene\n        .getVariables()\n        .get(eventsFunctionContext.getArgument(\"callback\"));\n\nif (electron !== null && electron.clipboard)\n    callback.setString(electron.clipboard.readText());\nelse if (\n    typeof navigator !== \"undefined\" &&\n    navigator.clipboard &&\n    navigator.clipboard.readText\n) {\n    navigator.clipboard.readText()\n        .then(text => callback.setString(text))\n        .catch(err =>\n            console.error(\"Error occured while getting clipboard content: \", err.message)\n        );\n} else console.error(\"Unable to read clipboard: No compatible method found!\")\n",
```
4ian, line 24: `Unable to read from the clipboard: no method found for this platform.`

4ian:
Alright we're almost good to go! :) 

The description of the action is I think not explicit enough about the fact that the variable won't be containing the clipboard text immediately after it's called.
Sure you wrote `asynchronously`, but for 95% of users this notion is not known. Instead, let's write:

`Read the text from the clipboard asynchronously (works on all platforms)` => `Read the text from the clipboard asynchronously. As this is "asynchronous", this means that the variable won't be immediately filled with the text from the clipboard. Instead, it will be filled a few milliseconds later. 

Note also that on some platforms (like on web browsers), the user will be asked for permissions to read from the clipboard.`

Finally, will this work on Cordova? Remember that half of the games are created for Android. If not, we should be clear in the description and the action name - and possibly keep the "(Windows, macOS, Linux)" everywhere in the extension.

4ian:
(I think this will work on Android/iOS in a browser but not when exported to Cordova, as this would require a plugin?... Or will the permission work? Not entirely sure, I don't think so because these native apps will need permissions at the OS level, but we should definitely search this information instead of risking users to be confused).

arthuro555:
So, after testing, it indeed doesn't work on cordova. The Web API is present and apparently somehow implemented, but it errors out saying the permission isn't granted. Considering that I cannot find any permissions for the clipboard on the android docs, I guess that it is just their way to say "quit it and use a plugin"?

For plugins, the only maintained one I could find is [this one](https://www.npmjs.com/package/cordova-clipboard), but it wasn't updated for a year. Then, I guess the clipboard API doesn't change that much. 

Indeed for now there is no mobile support possible as events functions extensions cannot have plugins. Should I rewrite it as JsExtension? Or should we add extension dependencies to events functions extensions? Leaving it like that when mobile is a very demanded platform doesn't sound good.

4ian:
> Considering that I cannot find any permissions for the clipboard on the android docs, I guess that it is just their way to say "quit it and use a plugin"?

Yes I think so. Most of the Web apis that are requiring a permission appeared after Cordova, and so Cordova is not "translating" web permissions to native permissions or vice versa. That's a bit sad but that's how it works.

> For plugins, the only maintained one I could find is this one, but it wasn't updated for a year. Then, I guess the clipboard API doesn't change that much.

Looks stable and yes the clipboard api does not need regular update - which is a good thing I guess :) 

>  Should I rewrite it as JsExtension? Or should we add extension dependencies to events functions extensions? Leaving it like that when mobile is a very demanded platform doesn't sound good.

That's a good question.
I would be in favor of seeing if we acn add support for the dependencies that you added to events functions extensions. That could be a fun little project. It would consist in:
1. Adding `gd::DependencyMetadata` to the `gd::EventsFunctionsExtension`
2. Update `declareExtension` in `MetadataDeclarationHelpers.js` to copy the dependencies from the `gd::EventsFunctionsExtension` to the "real" extension (`gd::PlatformExtension`).
3. Make a new dialog/tab in the `OptionsEditorDialog` to allow to specify dependencies.
4. ...and that's it?...
5. Profit!

There is already a way to specify a minimum GDevelop version for an extension, so that the user will be warned that this extension won't work if they try to use it on an old GDevelop version - so this part should be covered and safe already :) 

I would favor this because:
- it's likely that this is not the only extension that will need a Cordova plugin in the future.
- this is going in the direction of decreasing the gap between events functions extensions and traditional extensions, which is a good thing! Not sure if one day both will be exactly the same so much that we could merge the concept/classes into a single one, but at least it's good that we have one less reason to have to rewrite something as a JsExtension :) 

The clipboard is also one of these feature that is just between the "important, but maybe not important enough to be a core extension, but well still important enough" - so it's good if we can keep it in the extension registry :)

arthuro555:
Should we already merge this as a PC only extension and add the support for mobile later (after GDevelop starts supporting dependencies)?

4ian:
As this extension was already desktop only, yes we can do so, by making extra explicit that this will work only on "Windows, macOS, Linux and web browsers".

arthuro555:
Ok, done. I've also put the action name in PascalCase to be compliant with the best practices.

Extensions/Clipboard-header.json
```diff
@@ -10,6 +10,6 @@
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/clipboard-text-multiple-outline.svg",
   "eventsBasedBehaviorsCount": 0,
   "eventsFunctionsCount": 3,
-  "description": "This extension adds tools to access the clipboard.",
+  "description": "This extension adds tools to access the clipboard. It only work on Windows, Mac. Linux and browsers.",
```
4ian, line 13: `It only works on Windows, macOS, Linux and web browsers.`

Extensions/Clipboard.json
```diff
@@ -11,10 +11,10 @@
   "version": "0.0.2",
   "eventsFunctions": [
     {
-      "description": "Read the text from the clipboard asynchronously (works on all platforms)",
-      "fullName": "Get text from the clipboard",
+      "description": "Read the text from the clipboard asynchronously. As this is \"asynchronous\", this means that the variable won't be immediately filled with the text from the clipboard. Instead, it will be filled a few milliseconds later.\n\nNote also that on some platforms (like on web browsers), the user will be asked for permissions to read from the clipboard.\n\nThis only works on Windows, macOS, Linux and the browser",
```
4ian, line 14: `This only works on Windows, macOS, Linux and web browsers`

Extensions/Clipboard.json
```diff
@@ -11,10 +11,10 @@
...
+      "fullName": "Get text from the clipboard (Windows, macOS, Linux, browser)",
```
4ian, line 15: "web browsers"

Extensions/Clipboard-header.json
```diff
@@ -1,15 +1,15 @@
...
   "eventsBasedBehaviorsCount": 0,
-  "eventsFunctionsCount": 1,
-  "description": "Expression returning the content of the clipboard. Works only on Windows, macOS and Linux (not on iOS/Android or web games).",
-  "iconUrl": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz48IURPQ1RZUEUgc3ZnIFBVQkxJQyAiLS8vVzNDLy9EVEQgU1ZHIDEuMS8vRU4iICJodHRwOi8vd3d3LnczLm9yZy9HcmFwaGljcy9TVkcvMS4xL0RURC9zdmcxMS5kdGQiPjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0ibWRpLWNsaXBib2FyZC10ZXh0LW91dGxpbmUiIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNMTksM0gxNC44MkMxNC4yNSwxLjQ0IDEyLjUzLDAuNjQgMTEsMS4yQzEwLjE0LDEuNSA5LjUsMi4xNiA5LjE4LDNINUEyLDIgMCAwLDAgMyw1VjE5QTIsMiAwIDAsMCA1LDIxSDE5QTIsMiAwIDAsMCAyMSwxOVY1QTIsMiAwIDAsMCAxOSwzTTEyLDNBMSwxIDAgMCwxIDEzLDRBMSwxIDAgMCwxIDEyLDVBMSwxIDAgMCwxIDExLDRBMSwxIDAgMCwxIDEyLDNNNyw3SDE3VjVIMTlWMTlINVY1SDdWN00xNywxMUg3VjlIMTdWMTFNMTUsMTVIN1YxM0gxNVYxNVoiIC8+PC9zdmc+"
+  "eventsFunctionsCount": 3,
+  "description": "This extension adds tools to access the clipboard. Itonly works on Windows, macOS, Linux and web browsers.",
```
4ian, line 13: Typo

4ian:
Merged, thanks! 👍

PR #87, opened 2020-12-26
All tags to lower case

Bouh:
In reference to double **Camera** tags in the future extension search dialog.
https://github.com/4ian/GDevelop/pull/2140#issuecomment-751380418


4ian:
Sounds fair. Note that we could have tags that are different then in the extension json files... but I think it's ok, the search is only based on tags in the "extension short headers".

PR #93, opened 2021-01-10
Add BackButton extension

arthuro555:
Adds the possibility to control the back button on mobile and browsers. Doesn't work on electron because electron doesn't have a back button.

### DIFF
--- a/Extensions/BackButton.json
+++ b/Extensions/BackButton.json
@@ -1,0 +1,101 @@
+{
+  "author": "Arthur Pacaud (arthuro555)",
+  "description": "This extension prevents the back button from quitting your game, and let you decide what it does instead.",
+  "extensionNamespace": "",
+  "fullName": "Back button",
+  "helpPath": "",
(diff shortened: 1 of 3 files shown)


4ian:
Thanks for the PR! 👍 I'll check the code. 
Should this be called BrowserBackButton? I'm afraid this could be confused with an extension to check if the Android "Back" button was pressed.

arthuro555:
> to check if the Android "Back" button

It is for that as well

Extensions/BackButton.json
```diff
@@ -0,0 +1,101 @@
+{
+  "author": "Arthur Pacaud (arthuro555)",
+  "description": "This extension prevents the back button from quitting the game and adds a condition to decide what it does instead.",
```
4ian, line 3: "Prevents the back button from quitting the game and add a condition to check when it's pressed (to allow customising its behavior)."

(I avoid saying "This extension", as it's implied when shown on screen)

Extensions/BackButton.json
```diff
@@ -0,0 +1,101 @@
...
+          "parameterObjects": "",
+          "useStrict": true,
+          "eventsSheetExpanded": false
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "This simulate the normal action of the back button. This action will quit the app on mobile, and go back to the previous page on the web.",
```
4ian, line 60: "This simulate" => "This simulates"
"quit the app on mobile": "quit the app when in a mobile app"
"on the web": "when in a web browser"

4ian (approved):
Looks great overall! Added a few nitpickings/typos to fix

arthuro555:
I've fixed the wording issues you mentioned.

4ian:
Thanks! 👍

PR #94, opened 2021-01-10
Add compression extension

arthuro555:
Adds expression to inflate and deflate strings via the zlib port [pako](https://github.com/nodeca/pako).


4ian:
`// Load Pako: https://github.com/nodeca/pako` => `This is Pako vx.x.x, download from https://github.com/nodeca/pako/releases/...`

> if you have a JSON save file you might want to prevent editing it easily. 

We should probably avoid telling this because while it's true against a script kiddie, it can give a false sense of security

> # Compressor 

No need to put a header I think?

4ian:
Note about this: this could override another version of pako, like the one being used for Tilemaps. We'll consider this fine as both are using recent versions and the interface seems stable.

PR #100, opened 2021-01-25
Add extension for advanced P2P event handling

arthuro555:
This adds a way to handle all events in one frame using a while loop:
Screenshot of the way to use the extension


Extensions/AdvancedP2PEventHandling-header.json
```diff
@@ -0,0 +1,15 @@
+{
+  "shortDescription": "Adds a way to handle events the advanced way. It is more complex but also potentially more performant.",
```
4ian, line 2: "events" can be confusing for a new user, let's be more specific:

`Allows handling all received P2P events at once instead of one per frame`

Extensions/AdvancedP2PEventHandling-header.json
```diff
@@ -0,0 +1,15 @@
...
+  "extensionNamespace": "",
+  "fullName": "Advanced p2p event handling",
+  "name": "AdvancedP2PEventHandling",
+  "version": "1.0.0",
+  "url": "Extensions/AdvancedP2PEventHandling.json",
+  "headerUrl": "Extensions/AdvancedP2PEventHandling-header.json",
+  "tags": "p2p,performance",
```
4ian, line 9: let's added `advanced`, so we can maybe later differentiate these extensions in another category :)

4ian (approved):
Thanks, sounds good to go after minor edits :)

arthuro555:
I did the changes, sorry for forgetting this for 2 days 😬

4ian:
👍

PR #104, opened 2021-02-04
Gamepad - Avoid crash

Bouh:
Currently, PS5 controllers are not supported on our current version of Electron 8.2.5 (based on chrome 80.0.3987.165, Apr 21 2020 )

Support for PS5 was [added Nov 04 2020 in chrome](https://chromium-review.googlesource.com/c/chromium/src/+/2515099).
And chrome build 87.0.4280 for PS5 controller released 2020-11-17.

These mean the PS5 controller are not supported but the extension shouldn't crash.
Here the error

bug

Pressed is the first write bellow:

```JS
//When a button is pressed, save the button in lastButtonUsed for each players
if (gamepad.buttons[buttonId].pressed) gdjs._extensionController.players[playerId].lastButtonUsed = buttonId;
eventsFunctionContext.returnValue = gamepad.buttons[buttonId].pressed;
```

For avoid crash i think add 

```JS
if (gamepad.buttons == null || gamepad.buttons[buttonId] == null) {
    eventsFunctionContext.returnValue = false;
    return;
}
```

Without being able to test with this new controller I would like to make sure it's ok.
What you think?
I have to edit other functions too.
And I'll edit for use const and let, var are still present.


4ian:
I missed this @Bouh!

Yes I think it's a good idea to protect against undefined/null, like you mentioned.
Basically, we must be cautious that the `Gamepad` object might not have certain properties when it's not supported.

Have you implemented this fix in all functions? If so I think we can go ahead with this PR. Let me know!

Bouh:
Should be good now!
I've seen errors in expressions in the example, i'll update it in few minutes, and add this extension.

4ian:
Let's go ahead, thanks!

Issue #107, opened 2021-03-09
Update - Extended variables support


Bouh:
Here an update for update my **Extended variables support** extension.

- [x] Use const, let instead var.
- [x] Improve descriptions.
- [x] Align the extension on the best pratice.
- New expressions:
   - [x]  `ExtendedVariables::Variable(name)` Return the value of a scene variable.
   - [x] `ExtendedVariables::GlobalVariable(name)` Return the value of a global variable.
   - [x]  `ExtendedVariables::VariableString(name)` Return the text of a scene variable.
   - [x]  `ExtendedVariables::GlobalVariableString(name)` Return the text of a global variable.
- New actions:
   - [x]  Modify a global variable with a value.
   - [x]  Modify a scene variable with a value.
   - [x]  Modify a global variable with a text.
   - [x]  Modify a scene variable with a text.
   
Why these new expressions and actions?
Because in an variable field we can't write expressions.
For example we want create a variable named "Countdown123", and 123 is the value from the parameters in an action extension.
Here how we allow the user to create their variable, and below how it's not possible.

![image](https://user-images.githubusercontent.com/1670670/110402262-63491a00-807b-11eb-882f-2062e703f645.png)

These new expressions and actions use the GDJS API, this mean all is compatible with current variables, there is no differences.

If theses new expression and actions are not approved because this it should be supported in the engine i'm ok with that. In this case i'll only update the other points of the checklist.

Here the current state of the extension in a project, with the example shown on the screen.
(There is also another extension about timers in this project, forget it)
[countdownTimer.zip](https://github.com/4ian/GDevelop-extensions/files/6105127/countdownTimer.zip)

4ian:
> Here how we allow the user to create their variable, and below how it's not possible.

It's ok to do that in this extension :) I prefer to keep it in a separate extension because in general I want to make this kind of "dynamic" handling discouraged for "top level variables" (i.e: you can't just write `"My" + "Variable"` for a variable, but you can write it for a child of a structure: `MyStructure["My" + "Child"]`) - this way dynamic constructions are possible but we can have in the future error handling (You've used a variable called MyVariable, but there is one named myvariable in the list - did you make a mistake?).
This kind of "static checking" is not possible with fully dynamic variable names.
It's ok as you long as you know that you lose this capacity in the future, which is why I'm ok having this but in an "opt-in" extension.

PR #108, opened 2021-03-09
Update to 0.0.2

Bouh:
See #107
Need someone to review the english descriptions.

### DIFF
--- a/Extensions/ExtendedVariables-header.json
+++ b/Extensions/ExtendedVariables-header.json
@@ -1,15 +1,15 @@
 {
-  "shortDescription": "Conditions and actions to check for the existence of a variable, or delete existing ones from memory.",
+  "shortDescription": "Add some conditions, actions and expressions to check for the existence of a variable, delete existing ones from memory, and create dynamic variables.",
   "extensionNamespace": "",
   "fullName": "Extended variables support",
   "name": "ExtendedVariables",
-  "version": "0.0.1",
+  "version": "0.0.2",
   "url": "Extensions/ExtendedVariables.json",
   "headerUrl": "Extensions/ExtendedVariables-header.json",
-  "tags": "variables, scene variables, global variables, object variables",
+  "tags": "variable,scene,global,object,remove,expression",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/variable-box.svg",
   "eventsBasedBehaviorsCount": 0,
-  "eventsFunctionsCount": 5,
-  "description": "This extension provides conditions to check if a global/scene/object variable exists and actions to delete an existing global/scene variable from memory.",
+  "eventsFunctionsCount": 13,
+  "description": "This extension provides: \n- Conditions to check if a global/scene/object variable exists.\n- Actions to delete an existing global/scene variable from memory.\n- Actions for create a global/scene variable with a text input instead a variable input.\n- Expressions for get the value or text from a global/scene variable with a text input instead a variable input.\n\nText input instead of variable input allow the user to use or create dynamics variable with expressions like: \n`\"MyVar\"+ ToString(Variable(Health))`",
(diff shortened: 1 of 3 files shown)


Extensions/ExtendedVariables-header.json
```diff
@@ -1,15 +1,15 @@
...
   "url": "Extensions/ExtendedVariables.json",
   "headerUrl": "Extensions/ExtendedVariables-header.json",
-  "tags": "variables, scene variables, global variables, object variables",
+  "tags": "variable,scene,global,object,remove,expression",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/variable-box.svg",
   "eventsBasedBehaviorsCount": 0,
-  "eventsFunctionsCount": 5,
-  "description": "This extension provides conditions to check if a global/scene/object variable exists and actions to delete an existing global/scene variable from memory.",
+  "eventsFunctionsCount": 13,
+  "description": "This extension provides: \n- Conditions to check if a global/scene/object variable exists.\n- Actions to delete an existing global/scene variable from memory.\n- Actions for create a global/scene variable with a text input instead a variable input.\n- Expressions for get the value or text from a global/scene variable with a text input instead a variable input.\n\nText input instead of variable input allow the user to use or create dynamics variable with expressions like: \n`\"MyVar\"+ ToString(Variable(Health))`",
```
4ian, line 13: * "Actions for create a global/scene variable with a text input instead a variable input"
  =>
  Actions to create a global/scene variable using a text expression, rather than a variable name.

  Remember: "Pour" is only translated to "for" for nouns (and verb+ing). If there is a verb, it's always always always "to".

*  "Expressions for get"
  => "Expressions to get" and same fixes

* "Text input instead of variable input allow"
  => Don't forget the "s" to the verb for the 3rd person.

* "dynamics variable"

  Never add a "s" to adjectives in english: dynamic variables.

* But let's avoid speaking about "dynamic" variables, all variables are already dynamic. Let's also avoid talking about "Text Input" because people will confuse it by a widget to enter a text.
* Let's also avoid saying "the user". This description is already read by the user, no need to talk about them.

  Instead: "Text expressions allows to enter the name of the variable with an expression like `\"MyVar\"+ ToString(Variable(Health))`"

Extensions/ExtendedVariables.json
```diff
@@ -1,17 +1,25 @@
...
+    "scene",
+    "global",
+    "object",
+    "remove",
+    "expression"
+  ],
   "eventsFunctions": [
     {
-      "description": "Global variable exist",
+      "description": "A global variable exist.",
```
4ian, line 22: Check if a global variable exist**s**.

(s to the 3rd person)

Extensions/ExtendedVariables.json
```diff
@@ -73,29 +81,20 @@
       "objectGroups": []
     },
     {
-      "description": "Object variable exist",
+      "description": "An object variable exist.",
```
4ian, line 84: Same (check + s)

Extensions/ExtendedVariables.json
```diff
@@ -123,11 +122,11 @@
       "objectGroups": []
     },
     {
-      "description": "Delete global variable",
+      "description": "Delete a global variable, the variable will be deleted from the memory.",
```
4ian, line 125: Delete a global variable, removing it from memory.

Extensions/ExtendedVariables.json
```diff
@@ -123,11 +122,11 @@
...
       "fullName": "Delete global variable",
       "functionType": "Action",
       "name": "A_Delete_global_variable",
-      "sentence": "Delete global variable _PARAM1_",
+      "sentence": "Delete the global variable _PARAM1_",
```
4ian, line 129: No "the" because you're giving the name after in the parameter.

Extensions/ExtendedVariables.json
```diff
@@ -19,7 +19,7 @@
   ],
   "eventsFunctions": [
     {
-      "description": "A global variable exist.",
+      "description": "Check if a global variable exists.",
       "fullName": "Global variable exist",
```
4ian, line 23: I think this should be `Global variable exists` actually for the name too (and the scene/object variable).

4ian (approved):
Feel free to go ahead with this :)

PR #125, opened 2021-04-20
Add mobile support to clipboard extension

arthuro555:
Adds support via cordova plugins for mobile clipboard, hide deprecated ReadText expression, and bump to version 1.0.0 as I think it can be considered stable and complete as now that all platforms are supported, changes to the API are unlikely. 

Tested on my FairPhone 3.


Bouh:
👀dependencies from extension are used ?
I haven't yet read how it works, is used even on online building ?

arthuro555:
Yep I tested with online build. It uses extension dependencies to inject cordova plugins in the config.xml.

4ian:
Looks great! 👍

PR #126, opened 2021-04-20
Add console behavior

arthuro555:
I remade my first extension, it is a very simple behavior that basically splits by whitespace all words in a text input when pressing enter and allowing to process them via events. I use it for example to test extensions: 
image


Extensions/Console.json
```diff
@@ -0,0 +1,289 @@
+{
+  "author": "Arthur Pacaud (arthuro555)",
+  "description": "A console behavior that allows parsing simple commands inputed via a Text Input object.",
```
4ian, line 3: Could you use this description for the short description? And maybe add a bit more details to this description?

Extensions/Console.json
```diff
@@ -0,0 +1,289 @@
...
+  "extensionNamespace": "",
+  "fullName": "Console",
```
4ian, line 5: Maybe something like "Console (commands handler for Text Entry objects)" (a bit long, but more precise)

Extensions/Console.json
4ian: Text Entry, not Text Input.
(A text input would be a widget)

4ian:
May I still ask for another change: `ConsoleExtension` -> `TextEntryConsole`
No "Extension" (because by its nature, it is an extension), and "TextEntry" to make it very clear this is linked to this object. This might leave the room for another more generic "Console" extension in the future (no plans for this, but still - and we can always move this back to "Console" later).

4ian (approved):

PR #128, opened 2021-04-20
Add microphone extension

arthuro555:
Adds an extension to request microphone access and obtain the gain of the microphone.


4ian:
You're unstoppable recently! :) 
Can you check the compatibility so that we explicitely say in the extension in which platforms this will work?

arthuro555:
> You're unstoppable recently! :)

Idk, I had a lot of motivation today I guess :p

> Can you check the compatibility so that we explicitely say in the extension in which platforms this will work?

Getting the gain only works on electron, though I am not sure why as the permission works on chrome and use the same API 🤔 . I am not really sure how the WebAudio API works in depth so I am unsure how to fix it or add cordova support, I'll look into it later.

4ian:
Let’s add at least some info about compatibility in the extension
name/description then (something like “(only on desktop exports and
preview)” etc...)

Bouh:
I've taken a quick look at the extension.

- In **Gain** function you get the value it by **Volume**, in audio Gain and Volume are two different values.
 `eventsFunctionContext.returnValue = gdjs.microphone.meter.volume;`

In short, the gain value is used to expand or contract the waveform.
Then the volume is a post setting that will amplify the waveform on final output.

Both can be saturated. BUT the difference is here.
If a gain is saturated you can still adapt the volume the sound will be always saturated.
This is not the case if the gain is correct, and the volume too loud, this is correctable, reduce the volume to get a proper sound.
Gain has an influence on the output volume, this isn't true for the inverse.

- `gdjs.microphone` don't follow the [best practices](http://wiki.compilgames.net/doku.php/gdevelop5/extensions/best-practices#javascript), should be `gdjs._extensionMicrophone`.
(With time now I think change for `gdjs._extension.microphone` to get the possibility to easy reuse and see all extensions.)

- About compatibility, you know more than me (mobile/Cordova/deps).

- Some extensions have in descriptions what provides the extension, like which expressions, conditions, actions.
It's a great thing to understand what composes the extension. It's not just a description for say, "This allow to get the gain", you have also two other conditions to describe.
image

- `Returns the microphone gain (how loud the sound is, higher is louder).`
=> `Returns the microphone gain (a normalized value between 0 and 1, higher is louder).`

arthuro555:
> In audio Gain and Volume are two different values.

I know that :p Originally I named it volume everywhere but a while ago, after discussing it on discord, multiple persons told me that this is actually the gain not the volume, so I changed it to that.

> gdjs.microphone don't follow the best practices, should be gdjs._extensionMicrophone.
It is true that I named my namespace poorly here, but I still don't get where this best practice comes from. The convention for JavaScript extension was always to use as namespace `gdjs.evtTools.extensionName` afaik, I don't see why it shouldn't be put there for all extensions.

Bouh:
>  I still don't get where this best practice comes from. The convention for JavaScript extension was always to use as namespace gdjs.evtTools.extensionName afaik, I don't see why it shouldn't be put there for all extensions.

It's for separate and knowing where comes from the extension. Is it an internal extension from the engine, or is a user extension?
If all extensions have the same syntax how knowing without seeing it in the codebase.


> I know that :p Originally I named it volume everywhere but a while ago, after discussing it on discord, multiple persons told me that this is actually the gain not the volume, so I changed it to that.

It's not a big deal then we ca go ahead with volume :)

arthuro555:
I still have to do tests on mobile but I doubt it will work out of the box

arthuro555:
> It's for separate and knowing where comes from the extension. Is it an internal extension from the engine, or is a user extension?

What's the point in that though? 
1. There is no point in knowing if the extension is a JsExtenion or event based.
2. People won't look at the javascript namespace used in JsEvents to know what type of extension it is.
3. Looking at the namespace requires looking at the js code which require to know already what type of extension it is.

Bouh (commented):
Then make it just private, like 4ian wrote in the first review on an extension.

image

https://github.com/4ian/GDevelop-extensions/pull/9#issuecomment-499625167

arthuro555:
I applied Bouh's review instructions, just need to check mobile compat now

arthuro555:
Well it doesn't work on mobile. I tried using https://github.com/edimuj/cordova-plugin-audioinput#readme but didn't manage to make it work.

Bouh:
Thx for this extension!

4ian:
But is this working on mobile? I'm really worried of extension that would not because this is deceptive for users and give a bad view of GDevelop.

arthuro555:
No, I didn't get it to work on mobile.

4ian:
Let's make sure it's in the title of the extension and the description then (Windows/macOS/Linux).

If someone tries the webapp on an iPad/Android, if they say microphone, it's *obvious* for them that it will work. Same if anyone create a mobile game => of course Microphone means that it works on mobile, that's the majority of games developed today.

Imagine the frustration when it turns out it does not, without any indication. 
That's directly a bad comment "this game engine is bugged, 0/5, do not recommend" 😱

4ian:
As discussed on Discord, I've removed it temporarily, I know it's frustrating but we can't afford having confused users over this.

PR #150, opened 2021-05-14
HexagonGrid: Factorize the bubble snap

D8H:
I thought I couldn't factorized it but I actually did a typo.
Tested with the BimBam example


Extensions/HexagonalGrid.json
```diff
@@ -197,21 +197,12 @@
...
-          "parameterObjects": "Object",
-          "useStrict": true,
-          "eventsSheetExpanded": true
-        },
-        {
-          "disabled": true,
-          "folded": false,
```
Bouh, line 214: @D8H You are sure you don't use this part anymore?
I'am asking before merging.

D8H: I'm sure. The same JS code is in another action that is called in place.

Bouh:
Thank for your feedback, the extension is now updated!

4ian:
Thanks both 👍

PR #151, opened 2021-05-14
Add FlexBox extension

arthuro555:
Adds a behavior to use flexbox layout inside GDevelop.  
[Example project](https://trello-attachments.s3.amazonaws.com/5fd532f01e99c97130e99fe9/609ec3db127ab987a25fe16e/e53fd9e0bd2b32b399d3948a7420c831/layout.zip).

### DIFF
--- a/Extensions/FlexBox-header.json
+++ b/Extensions/FlexBox-header.json
@@ -1,0 +1,15 @@
+{
+  "shortDescription": "Allows using flexbox for positioning objects",
+  "extensionNamespace": "",
+  "fullName": "Flexbox",
+  "name": "FlexBox",
+  "version": "0.0.1",
+  "url": "Extensions/FlexBox.json",
+  "headerUrl": "Extensions/FlexBox-header.json",
+  "tags": "flexbox,layout,css,items,container,html",
+  "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/page-layout-body.svg",
+  "eventsBasedBehaviorsCount": 2,
+  "eventsFunctionsCount": 0,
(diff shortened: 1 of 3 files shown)


4ian:
Interesting. You're using hidden DOM elements to compute this?

arthuro555:
Yep. I had at first the problem that not attaching those to the DOM wouldn't calculate the CSS, and attaching it to the DOM would mess with the document and have the other elements mess with the positioning, but I worked around that by making an IFrame that I hide via CSS and put everything inside that IFrame.

4ian:
I see! How do you do for synchronisation, when the window is resized or objects resize?

arthuro555:
> when the window is resized

I do not handle that, to do so the anchor behavior has to be used on the container object.

> or objects resize?

I am only handling that for containers in onScenePreEvents, not for items as flexbox needs the size of the items to work, and will resize them afterwards if flex-grow/flex-shrink is not 0. I can't know if the resizing has been done by the user events or flexbox itself (not without prototype hacks that would deoptimize the game at least), so I always feed flexbox the size the object had when created.

4ian:
Ok! I'm always worried of using DOM apis because it ties the extension/game to running in the browser forever (unless you switch to a flexbox implementation like Yoga) and the DOM is not known for its speed/performance.
I think we should clearly mark this in the extension tags and maybe even an "experimental" in the extension name?

arthuro555:
I originally looked into Yoga but couldn't get the browser bundle to work, and it also looked like it wasn't actively maintained i think. 

> I think we should clearly mark this in the extension tags

What exactly should be in the tags? Low performance? DOM APIs? Not working on node?

4ian:
Let's add a tag `DOM apis` so that we can follow what extensions are using it :) I'll update the best practices about this.

PR #161, opened 2021-06-08
Add Repeat every X seconds extension

arthuro555:
Adds the "Repeat every X seconds" by @VegeTato, an extension that adds convenience abstractions for timers.  
Reviewed and modified to support both objects and scene timers by me.

### DIFF
--- a/Extensions/RepeatEveryXSeconds-header.json
+++ b/Extensions/RepeatEveryXSeconds-header.json
@@ -10,6 +10,6 @@
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/repeat.svg",
   "eventsBasedBehaviorsCount": 1,
   "eventsFunctionsCount": 2,
-  "description": "Convenience functions for timers to trigger a condition every X seconds.",
+  "description": "Convenience conditions, actions and behaviors for timers to trigger a condition every X seconds.",
   "iconUrl": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz48IURPQ1RZUEUgc3ZnIFBVQkxJQyAiLS8vVzNDLy9EVEQgU1ZHIDEuMS8vRU4iICJodHRwOi8vd3d3LnczLm9yZy9HcmFwaGljcy9TVkcvMS4xL0RURC9zdmcxMS5kdGQiPjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0ibWRpLXJlcGVhdCIgd2lkdGg9IjI0IiBoZWlnaHQ9IjI0IiB2aWV3Qm94PSIwIDAgMjQgMjQiPjxwYXRoIGQ9Ik0xNywxN0g3VjE0TDMsMThMNywyMlYxOUgxOVYxM0gxN003LDdIMTdWMTBMMjEsNkwxNywyVjVINVYxMUg3VjdaIiAvPjwvc3ZnPg=="
 }
--- a/Extensions/RepeatEveryXSeconds-header.json
+++ b/Extensions/RepeatEveryXSeconds-header.json
@@ -1,15 +1,15 @@
 {
-  "shortDescription": "This allows to trigger an event every X seconds",
+  "shortDescription": "This allows to trigger an event every X seconds.",
   "extensionNamespace": "",
   "fullName": "Repeat every X seconds",
   "name": "RepeatEveryXSeconds",
   "version": "0.1.0",
   "url": "Extensions/RepeatEveryXSeconds.json",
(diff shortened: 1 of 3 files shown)


Bouh:
The description is empty, it should explain what is for the extension, and how to use it.
See for **Camera shake** extension, there is a great description!
Personally I list what contain the extension, like with **Extended variable support**.

Descriptions for functions must finish with a dot, also function in behavior.

**RepeatXTimes** have a parameter#2 **time** with label: **The time in seconds between each trigger**
**Repeat** have a parameter#2 **timer** with same description, it's ok just unify the parameter name accross the function when the parameters are same.

I hope you have tested the extension first, I trust you ;)

arthuro555:
> The description is empty.  
> unify the parameter name across the function when the parameters are same

Whoops, those are points i actually noticed while reviewing the extension and had fixed, but i think the changes must have been lost without me noticing as GDevelop crashed multiple times while I fixed the issues.

> For authors, use a comma rather than an esperluette &

🤔 Why does this matter?

> I hope you have tested the extension first, I trust you ;)

Well yes that's part of reviewing the extension :p

Extensions/RepeatEveryXSeconds.json
```diff
@@ -0,0 +1,684 @@
...
+              "name": "Behavior",
+              "optional": false,
+              "supplementaryInformation": "RepeatEveryXSeconds::RepeatTimer",
+              "type": "behavior"
+            }
+          ],
+          "objectGroups": []
+        },
+        {
+          "description": "Allows to trigger the recurrent timer X times again.",
```
4ian, line 589: "recurrent timer" => "recurring timer" (here and everywhere ;))

Extensions/RepeatEveryXSeconds.json
```diff
@@ -0,0 +1,684 @@
...
+          "label": "The time between each trigger",
+          "description": "",
+          "extraInformation": [],
+          "hidden": false,
+          "name": "TimerLength"
+        },
+        {
+          "value": "-1",
+          "type": "Number",
+          "label": "How many times is the timer allowed to trigger? -1 for forever.",
```
4ian, line 666: "How many times should the timer trigger? -1 for forever."

Extensions/RepeatEveryXSeconds.json
```diff
@@ -0,0 +1,684 @@
...
+          "label": "The name of the timer to repeat",
+          "description": "",
+          "extraInformation": [],
+          "hidden": false,
+          "name": "TimerName"
+        },
+        {
+          "value": "1",
+          "type": "Number",
+          "label": "The time between each trigger",
```
4ian, line 657: Specify the unit: "The time between each trigger (in seconds)"

Extensions/RepeatEveryXSeconds.json
```diff
@@ -0,0 +1,684 @@
...
+          "optional": false,
+          "supplementaryInformation": "",
+          "type": "expression"
+        }
+      ],
+      "objectGroups": []
+    }
+  ],
+  "eventsBasedBehaviors": [
+    {
```
4ian, line 299: Can you explain the advantage of the behavior compared to a set of actions? This avoids to repeat the name of the timer I guess?

Extensions/RepeatEveryXSeconds-header.json
```diff
@@ -0,0 +1,15 @@
...
+  "fullName": "Repeat every X seconds",
+  "name": "RepeatEveryXSeconds",
+  "version": "0.1.0",
+  "url": "Extensions/RepeatEveryXSeconds.json",
+  "headerUrl": "Extensions/RepeatEveryXSeconds-header.json",
+  "tags": "repeat,timer,time,convenience",
+  "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/repeat.svg",
+  "eventsBasedBehaviorsCount": 1,
+  "eventsFunctionsCount": 2,
+  "description": "Convenience functions for timers to trigger a condition every X seconds.",
```
4ian, line 13: Not "functions" => "conditions, actions and behaviors".

Remember, the term function should not be surfaced outside of an extension, for users the concepts to understand are "conditions", "actions" and "behaviors" :)

4ian (approved):
Added a few comments ;) 
Looks good otherwise!

Extensions/RepeatEveryXSeconds.json
arthuro555: > This avoids to repeat the name of the timer I guess?

Yes, that's it. If that is not a good idea though i can make those regular actions.

4ian: I'm not sure tbh, the only real risk for now is that behaviors can't be easily renamed in objects.

4ian:
Thanks @arthuro555 and @VegeTato!

Extensions/RepeatEveryXSeconds.json
```diff
@@ -0,0 +1,684 @@
...
+          "optional": false,
+          "supplementaryInformation": "",
+          "type": "expression"
+        }
+      ],
+      "objectGroups": []
+    },
+    {
+      "description": "Triggers every X seconds X amount of times.",
+      "fullName": "Repeat every X seconds in X number",
```
4ian, line 127: @arthuro555  I merged this a bit fast, this is not proper english.  Can you fix that?

PR #168, opened 2021-06-26
Add Render to Sprite extension

arthuro555:
## Describe the extension
This extension allows rendering objects, a layer or a scene into a sprite object.

Allows rendering (creating an image) of objects, layers and the scene to display in a sprite. 
This allows to make optimizations by transforming the resulting image of a shape painter into a faster sprite, or to make viewports/cameras.

## Checklist
- [x] Extension has a proper name and description.
- [x] Extension has tags (for example: "platform, brick, breakable").
- [x] All behaviors have a description.
- [x] All functions (actions, conditions, expressions) have descriptions and proper sentences.
- [x] I've followed the best practices listed on http://wiki.compilgames.net/doku.php/gdevelop5/extensions/best-practices
- [x] I confirm that this extension can be intergrated to this GitHub repository, distributed and MIT licensed.


4ian:
Interesting, I guess this is stretching the limits of what the engine/extensions allow to do :) 
Could you describe the overall way this works? Any internal API or things that look a bit hacky in the way it's done?

arthuro555:
Basically:

1. If the target sprite texture is not a render texture, replace it with a new render texture.  
2. Resize the render texture of the target sprite to the size of the source.

If the source is not an object {

3. If there is no temporary render texture create one and cache it
4. If there is no temporary sprite create one that uses the temporary render texture and cache it
5. Resize the temporary render texture to the size of the source.
6. Render the source render object using the PIXI `Renderer` into the temporary render texture ([needed to prevent a WebGL crash](https://www.html5gamedevs.com/topic/45423-why-is-this-not-allowed/)).
7. Render the temporary sprite using the PIXI `Renderer` into the render texture of the target object.

} If it is an object {

3. Store the objects position in a variable
4. Set its position to the origin to make sure it fits perfectly inside the render frame
5. Render the source render object using the PIXI `Renderer` into the render texture of the target object.
6. Restore the object position

}

Bouh:
I edited your first post with the information found in the extension.
It's easier with the description to know what the extension is for.

About javascript events, there is a way to define methods in a lifecycle function once instead of checking or recreate the methods each time the functions are called.

Can you add a comment on the process for creating and resize the render texture In **RenderLayer**? 
The rest of the code is commented but not the beginning. Copy/paste comment in **RenderScene** too.

`gdjs._renderToSprite.rt` is checked twice `gdjs._renderToSprite.rt`

arthuro555:
> About javascript events, there is a way to define methods in a lifecycle function once instead of checking or recreate the methods each time the functions are called.

Not sure what you are talking about, I am not redefining anything.

> Can you add a comment on the process for creating and resize the render texture In RenderLayer?
> The rest of the code is commented but not the beginning.

I never copy the begin of the code because I only define constants and get function arguments here which is something pretty clear. I am someone who thinks commenting too much is more harmful than too little. If I think something needs to be explained I comment it but if it is self-explanatory I don't.

arthuro555:
If there are no other review or issues I'll merge this in a few days

arthuro555:
Since there have been no objections, I'll go ahead and merge this now.

4ian:
As this is fairly linked to the renderer/PixiJS logic, is there a chance this breaks if we upgrade PixiJS? I would be in favor of adding a "(experimental)" tag at the end of the name maybe.

arthuro555:
It is possible, in fact, that API had a breaking change from v5 to v6 I believe.

Bouh:
Adding a PixiJS tag is welcome for the day where the update will be done ;)

4ian:
Yep good idea! 
Basically, we can think of extensions as in 5 categories (technically speaking):

- Pure events based (super stable)
- JS based, using the public interface of the game engine (fairly stable)
- JS based, using PixiJS stuff outside of the game engine (can break if we upgrade Pixi, so better be sure about what we're using/mark experimental/use with caution)
- JS based, using DOM apis for non visual stuff (fairly stable, but still caution here, because some browser might not support everything, and a non browser JS environment like Node might not have these).
- JS based, using DOM apis for visual stuff (extreme caution, as this is stepping outside of the traditional game engine/render loop - might only work for stuff like text input).

Tags will be useful to keep this visible.

Bouh:
Yes, internally, it will help us maintain the extension and identify those that may be broken 👍
I've added PixiJS tag ✅

PR #170, opened 2021-06-29
Add autoclose for issues with a default template

Bouh:
Adding a Github token can be required in settings of this repo.


4ian:
👍

PR #174, opened 2021-07-05
AutoTyping add BitmapText support

Bouh:
Ready to merge, it's just a copy paste of a new behaviors adapted for bitmap text object.
It works like other behaviors for text and bbtext.


4ian:
👍

PR #183, opened 2021-07-15
Add ScriptEngine extension

arthuro555:
Adds a script engine extension, which allows executing JavaScript safely by isolating it in a web worker and exposing an events sheet controlled JavaScript API to that script. This allows to execute arbitrary scripts to allow extensions/mods/custom stuff to be made for games without exposing all of the game internals.  
Electron/node APIs seem to not be accessible from there as well, so this keeps players machines integrity safe (tested on preview and build).

Example game: [scripting.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/6819760/scripting.zip)

### DIFF
--- a/Extensions/ScriptEngine.json
+++ b/Extensions/ScriptEngine.json
@@ -1,0 +1,427 @@
+{
+  "author": "Arthur Pacaud (arthuro555)",
+  "description": "Adds a mechanism to execute arbitrary JavaScript from a GDevelop game, while not exposing a direct access to the game by running the code in another isolated thread (via WebWorkers).  The code can still interract with the game but only through APIs designed by you.\n\n## Usage\n\nThe first step is defining the API. This is necessary to do **before** executing a script. To define the API, you only need to run the condition that checks if the api has been called once. If you do not want to check if the APIs have been called yet, you can still define the APIs with the \"Define API\" action.  \nOnce you have done that, you can execute a script, call APIs from it and process the calls from events.\n\n## Notes\n\n - When an \"API called\" condition is trigered, the api call linked to it will be picked. This means that the return actions and the get argument expressions will apply to it. \n - Checking if another API has been called implicitely returns the currently picked api call. \n - Once you have returned an API call, it is unpicked, meaning that its return value cannot be changed and its argument not read anymore.\n - Once you have executed a script, new APIs cannot be defined for that script, they will be defined for next scripts though.\n",
+  "extensionNamespace": "",
+  "fullName": "Script Engine",
+  "helpPath": "",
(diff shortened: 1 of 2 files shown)


4ian:
Very interesting :)

Could you make an example of this?
Have you looked precisely at the isolation guarantees offered by a webworker?
I've seen [Realms](https://github.com/agoric/realms-shim/) and also some risks [associated with them](https://www.figma.com/blog/an-update-on-plugin-security/)

I would be ok to go ahead with this extension if we can have a bit of documentation written about its usage on the wiki, and also an "experimental" tag (+ "(experimental)" in the title) to make sure people using it are cautious.

arthuro555:
> Have you looked precisely at the isolation guarantees offered by a webworker?
I've seen Realms and also some risks associated with them

I had taken a look at realms when starting the extension, but since there seems to be no implementer interest for the standard and the shim is basically [evaluating the code](https://github.com/Agoric/realms-shim/blob/a5ae1fca7707d289c5ee817f50f3b5f1edb67a51/src/evaluators.js#L60) (simplifying a lot here it is doing a lot of scope manipulation and overriding and code string manipulation to make it actually secure), which in my opinion is a flawed way to do it as there probably always will be a way to do get access to higher scope (and if not now later through new ECMAScript features), I have decided against using it. This Figma blog post further makes me think it is right not to use it.

Isolation wise, Workers are almost perfect, as everything runs in another isolated scope, and the only way to interact with the  is through postMessage. As long as I don't use eval in the main context message handler, which I don't, the worker is truly limited in terms of interacting with the game. According to [the OWASP](https://cheatsheetseries.owasp.org/cheatsheets/HTML5_Security_Cheat_Sheet.html#web-workers), the only security concerns are Denial of Service attacks (by using excessive CPU power). I don't see how a script engine can avoid this type of attacks generally and I think it's still better to have it in a separate thread than in the UI thread.

Some other isolation concerns I have would be the same site policy and access IndexedDB. I think that it is fine (for now) to not remove the access to the games indexed db as nothing in GDevelop is using it right now so no harm can be done. The same site policy is a bit more annoying, if the worker is on a website then it can send a request with cookies to any endpoint on the same origin as it is identified as of the same origin 😬.
A solution would be dereferencing in the worker runtime any "Dangerous" method like `fetch`, `XMLHTTPRequest`, `indexedDB`... But I would rather not do that (or at least not by default) as it severely limits the possibilities with this extension. I don't think it would be a safe way anyways as who knows when a new "dangerous" API could be added and making all the efforts useless until the extension and games using the extension are updated :/

I think I will update the extension to dereference the worker postMessage and event listener methods to be sure that any communication is indeed done through the runtime only though.

The alternative would be to use an iframe with the sandbox attribute, but I would rather not use DOM APIs when [workers are compatible with node](https://nodejs.org/docs/latest-v14.x/api/worker_threads.html#worker_threads_class_worker). IFrames have as far as I am aware the same Denial of Service vulnerability and in most implementations run in the same thread as the main page meaning blocking code there would block the game 😬. It wouldn't have access to any storage (since the iframe would be on an unique anonym origin after it is terminated the storage will never be accessible again), it would have an overhead due to having a whole DOM and render loop, but it would solve the same origin policy issue. Using a worker in an IFrame is not a viable solution as a worker cannot run on a sandboxed IFrame so the sandboxing would have to be disabled, at which point it makes no difference from running the worker directly except that the IFrame adds an overhead.

The figma blog post also suggested using a wasm compiled JS engine, which is something I had considered too, but I think that in games, performance is critical, so a Worker would be more adapted to prevent any slowdowns from a bad script and because v8 executed directly is probably way faster than a small&easy wasm executed engine.

> I would be ok to go ahead with this extension if we can have a bit of documentation written about its usage on the wiki, and also an "experimental" tag (+ "(experimental)" in the title) to make sure people using it are cautious.

This is fine for me 👍 

> Could you make an example of this?

I have already included a GDevelop project using it in the original post.

Extensions/ScriptEngine.json
```diff
@@ -0,0 +1,427 @@
...
+  "name": "ScriptEngine",
+  "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Glyphster Pack/Master/SVG/Applications and Programming/Applications and Programming_script_code_snippet_download_save.svg",
+  "shortDescription": "Execute arbitrary JavaScript safely in GDevelop games.",
+  "version": "1.0.0",
+  "tags": [
+    "worker",
```
Bouh, line 13:
```suggestion
    "worker",
    "experimental",
```

Extensions/ScriptEngine.json
```diff
@@ -0,0 +1,427 @@
...
+  "extensionNamespace": "",
+  "fullName": "Script Engine",
```
Bouh, line 5:
```suggestion
  "fullName": "Script Engine (experimental)",
```

Bouh (commented):
Would be nice to add the example with this extension, and add a few words about it in the description of the extension to let people find how to use this ScriptEngine ext.

Issue #195, opened 2021-07-28
Magnetic Effect


Entr0py404:
## Describe the extension
With this behavior, you can attract an object to another object, such as coins attracted to a player.
![FgszB1u0PT](https://user-images.githubusercontent.com/75917656/127251841-14c7ca8b-4376-42b7-93d8-e72dec52c0d1.gif)

## Checklist
- [x] Extension has a proper name and description.
- [x] Extension has tags (for example: "platform, brick, breakable").
- [x] All behaviors have a description.
- [x] All functions (actions, conditions, expressions) have descriptions and proper sentences.
- [x] I've followed the best practices listed on http://wiki.compilgames.net/doku.php/gdevelop5/extensions/best-practices
- [x] I confirm that this extension can be intergrated to this GitHub repository, distributed and MIT licensed.

## Example
[Magnetic Effect Extension Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/6889781/Magnetic.Effect.Extension.Example.zip)

## Extension file
[MagneticEffect.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/6889779/MagneticEffect.zip)

4ian:
Looks nice!
A few nitpicking:
- Rate should be named Speed. Also include the unit in the sentence ("a rate of _PARAM4_ " => "a speed of _PARAM4_px/s")
- Distance should be "Minimum distance". In the sentence, also change "within a distance below _PARAM3_" to "(until the minimum distance of _PARAM3_px)"
- In the sentence, add punctuation for "permanently _PARAM5_" =>  "(permanently: _PARAM5_)"

I'll probably also adapt the descriptions of the behaviors/extensions, you can let this to me :)

ClementPasteau:
Thanks for this extension!

I've also noticed that the logic of the behavior has some actions duplicated between "Permanent" and "non-permanent" that could be moved outside of any condition on the permanent boolean :)

I'll take it from here to publish it!

arthuro555:
@ClementPasteau Remember to close the issue after adding the extension ;)

PR #199, opened 2021-08-06
Add Magnetic Effect extension

ClementPasteau:
Add Magnetic Effect extension suggested in #195 
Reviewed, tested and tweaked some variables naming and the logic.

### DIFF
--- a/Extensions/MagneticEffect.json
+++ b/Extensions/MagneticEffect.json
@@ -1,0 +1,749 @@
+{
+  "author": "Entropy",
+  "description": "With this behavior, you can attract an object to another object, such as coins attracted to a player.",
+  "extensionNamespace": "",
+  "fullName": "Magnetic Effect",
+  "helpPath": "",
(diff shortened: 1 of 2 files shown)


arthuro555:
On the extension and examples repository, we handle things differently than on the main repository. On the main one, PRs are expected so that 4ian can review them. Here though, we can push directly on main without having a PR with 4ians approval. Once you have accepted an extension, you can push your commit directly instead of making a PR. This had been decided to allow the workload of 4ian (and probably yours too now) to be reduced.

ClementPasteau:
Thanks for the feedback @arthuro555 - I didn't know that :)

As this is the first time I'm merging on those repositories, I wanted a quick review to make sure I wan't missing something obvious! I'll push directly moving forward

arthuro555:
> I wanted a quick review to make sure I wasn't missing something obvious

Everything looks fine to me:
- [X] Header regenerated
- [X] File name same es internal extension name
- [X] All check passed

And I trust you have checked before making the PR that the extension is working and respects the best practices.

4ian:
Thanks both!

PR #204, opened 2021-08-24
[Slider] Draw only when needed.

D8H:
This uses the new "Clear" action of the ShapePainter indroduced in GDevelop 5.0.0-beta115.
It avoid to rebuild the shapes at every frame.

An example, to test: https://www.dropbox.com/s/pwwtqds56e5048f/MarchingSquares.zip?dl=1
and its build: https://games.gdevelop-app.com/game-b6f7ae43-151b-4cd1-95fe-fc30cadcb56c/index.html


4ian:
Excellent, thanks for taking the time to improve this!

PR #207, opened 2021-08-25
Marching Squares extension

D8H:
This extension allows to build a scalar field and draw contour lines of it.

It can be helpful to draw:
* fog of wars
* liquids effects

## Examples
* 3 short cases
  * Project: https://www.dropbox.com/s/pwwtqds56e5048f/MarchingSquares.zip?dl=1
  * Build (Press Space to switch cases): https://games.gdevelop-app.com/game-70751f50-09b7-4998-b94f-0e26720a7cb8/index.html
* An integration into the RTS example by Tristan
  * Project: https://www.dropbox.com/s/f8jmgeupm5dz8xe/RectangularFloodFill.zip?dl=1
 
GDevelop_5_ZaTG4F44SB
MarchingSquaresWater mp4_snapshot_00 32_ 2021 08 25_17 27 14


Bouh:
This extension is impressive, not sure if I'm able to review this.
Btw I've tried to add an object shader on the shape painter object, but the effect isn't applied, is there limitation in your code?

D8H:
> This extension is impressive, not sure if I'm able to review this.

The effect is impressive but the algo is really simple. Basically, it's a switch of 16 cases and some mean functions to smooth the edges. The 1st image of Wikipedia explains the whole algo: https://en.wikipedia.org/wiki/Marching_squares

But in general, do reviewers need to understand how the code work? If the review is to ensure good practices and try to find bug, I guess it's not necessary.

For instance, it adds prototype functions on the behavior. Maybe it's a bad practice or it should use a prefix to avoid collision or do it on a member instead.

> Btw I've tried to add an object shader on the shape painter object, but the effect isn't applied, is there limitation in your code?

It changes the shape style to fill and outline independently, but this should be ok I guess.
I will check. I used a blurring effect on a layer at some point and it was working.

Bouh:
> I used a blurring effect on a layer at some point and it was working.

For effect on layer yes, for effect on the object no. I've used the nightly build from master ;)

4ian:
Awesome I've done a few quick changes (`gdjs.RuntimeScene.parseOperation` => `gdjs.__marchingSquares.parseOperation` to follow the conventions) and I'll add the game as an example too :)

D8H:
> I'll add the game as an example too :)

I can make a PR if you want. I was thinking of submitting one example for the fog of war and another one for the liquid effect and not submitting the one that let draw a field.

Bouh:
Go ahead, but there is already an example.

https://github.com/GDevelopApp/GDevelop-examples/commit/f11d8b4aebd4d30210a44a73015e351944aac352

4ian:
I went ahead as the same time I submitted the example, but feel free to open a PR with two separate examples and removing the one I put - would be better I think!

4ian:
Just tried with the Kawase Blur on the shape painter objects, it seems to work well! 

image

PR #210, opened 2021-08-27
Add DiscordRichPresence extension

arthuro555:
Adds an extension for discord rich presence.

I was unable to properly test the joining parties feature, as it would require 2 computers with 2 discord account. Should I remove it just in case or have faith in the discord client library I used?

This is unaffected by https://github.com/4ian/GDevelop/issues/1947, since this discord client library is bundled with GDevelop for its own rich presence integration, and will therefore be installed and available for previews to use.


Extensions/DiscordRichPresence.json
```diff
@@ -0,0 +1,372 @@
...
+      "functionType": "Condition",
+      "name": "OnJoinRequest",
+      "private": false,
+      "sentence": "Join request received",
+      "events": [
+        {
+          "disabled": false,
+          "folded": false,
+          "type": "BuiltinCommonInstructions::JsCode",
+          "inlineCode": "eventsFunctionContext.returnValue = gdjs.evtTools.drp.joinRequest !== null;\r\n",
```
4ian, line 266: Nitpicking: I know it's a bit of a pain, but even  on `evtTools`, `drp` is not a name that is unique enough. What if someone wants to make a "dynamic response plugin" in the future? Name your namespace "discordRichPresence" ;)

Extensions/DiscordRichPresence.json
```diff
@@ -0,0 +1,372 @@
...
+          "parameterObjects": "",
+          "useStrict": true,
+          "eventsSheetExpanded": false
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Answers the last recieved join request.",
```
4ian, line 339: received

4ian (approved):
Overall lgtm!
I think it's ok to use the fact that we already have the plugin in the IDE.

4ian:
I think it's ok as long as we can coordinate the release of both :)

arthuro555:
Ok, I think this should be ready to go now. The only thing I am unsure about is the events sheet sentence for the update rich presence action, could someone check it, and if it is good, merge it (or tell me to do it if you do not have the perms)? Thanks!

arthuro555:
Oh also I've added `gdjs.Logger` to the exceptions for all extensions as I think it is a pretty common task to want to log something from JS events

arthuro555:
No one reviewed the changes, so I'll go ahead and merge this as the only thing blocking it is that little sentence that can easily be changed without problems later, and I had already gotten approval before.

4ian:
I think it looks good 👍

PR #211, opened 2021-09-01
Add an extension to stick objects on other ones.

D8H:
This extension can be useful for
* Sticking an accessory to a moving object
* Animating a skeleton

## Examples
* 2 short cases
  * Project: https://www.dropbox.com/s/2d6pwoj3zpykdj2/StickerTest.zip?dl=1
  * Build: https://games.gdevelop-app.com/game-8000ff47-e6ab-4787-964f-623340754cec/index.html

* An integration into the RTS example by Tristan
  * Project: https://www.dropbox.com/s/moqrgz3ginctwf7/RectangularFloodFill.zip?dl=1
    * It avoids LinkedObject searchs for a 3% overall performance gain.
	* It avoids to use complicated concepts like for each loops and linked objects for a trivial thing.

Vines

Boat


4ian:
Nice! Are you able to use `_getOnlyFollowPosition()` instead of `_behaviorData.OnlyFollowPosition`?
I think the property getters are in the future more likely to be stable than the `_behaviorData` object that could be renamed (even if also unlikely).

Bouh:
An action is useless, it works without.
image

The project with boat and vine can be added into GDevelop too, what you think?

In object property,  what did you think of "Lock the angle"? 
Can be cool to make a ["bielle"](https://www.youtube.com/watch?v=1EkTe9fR7_4) ([like this blue bar on video](https://www.youtube.com/watch?v=lrZSd8ALOn8))

D8H:
> An action is useless, it works without.

Without it the crew members size depends on the size of the boat when they are created.
It needs a comment.

> The project with boat and vine can be added into GDevelop too, what you think?

Yes, I was waiting to see if the extension were accepted.
I will probably submit them in the same project as they don't have a clear goal other than showing how to use this extension.

> In object property, what did you think of "Lock the angle"?
> Can be cool to make a ["bielle"](https://www.youtube.com/watch?v=1EkTe9fR7_4) ([like this blue bar on video](https://www.youtube.com/watch?v=lrZSd8ALOn8))

Yes, I hesitated to add it. I couldn't find a good use case (but I'm sure there is). So, I decided to wait and see if someone ask for it.
This kind of movement could be interesting for a moving platform but I guess I would use the action "Put around a position" with the position in object variables to do it.
Do you have a use case in mind?

Bouh:
> > In object property, what did you think of "Lock the angle"?
> > Can be cool to make a ["bielle"](https://www.youtube.com/watch?v=1EkTe9fR7_4) ([like this blue bar on video](https://www.youtube.com/watch?v=lrZSd8ALOn8))
> 
> Yes, I hesitated to add it. I couldn't find a good use case (but I'm sure there is). So, I decided to wait and see if someone ask for it.
> This kind of movement could be interesting for a moving platform but I guess I would use the action "Put around a position" with the position in object variables to do it.
> Do you have a use case in mind?

A valve for open a door?
Or for dosing a liquid in a gauge?
[Or this](https://www.youtube.com/watch?v=WUGQCi98d1M)

tristanbob:
> > This kind of movement could be interesting for a moving platform but I guess I would use the action "Put around a position" with the position in object variables to do it.

I agree, if you are not adjusting rotation then it makes sense to use the simpler built-in tool "put object around position".

> [Or this](https://www.youtube.com/watch?v=WUGQCi98d1M)

We already have this with the "OrbitingObjects" extension.  :)

4ian:
Looking at the extension, what makes me afraid in the code is the need to save a reference to an object in the behavior.
This means that when the object gets deleted (or worse, recycled), the reference will stay there and Very Bad Things will happen. A memory leak in the best case (which can be damaging as it will also point to Pixi objects that can use WebGL stuff).

Two things:
- Either you can try to use `gdjs.registerObjectDeletedFromSceneCallback`. This is guaranteed to be called when an object is deleted. Probably also listen for scenes being unloaded? In both cases, you should somehow drop any reference to these objects.
- Or you store just the id of the object, and find a way to get it back?

This needs more thinking, but basically I'm afraid we're messing with future changes to the scene/objects lifecycle, and this will for sure create bugs in a future version.

4ian:
Thanks for the update! I did a few changes to make this ready:

* Fixed Unstick => no Basis object, so it was not working

* Removed useless code:
```
const pointX = eventsFunctionContext.getArgument("PointX");
const pointY = eventsFunctionContext.getArgument("PointY");
```

* Not need for `for(const object of objects)`

* `@param basisObject {gdjs.RuntimeObject}` => `@param {gdjs.RuntimeObject} basisObject`

* Avoid label "Object" for parameter (give more context: "Object to stick to")

* upateOldCoordinates => updateOldCoordinates

* Need to do:
  ```
  if (!runtimeScene.__allStickers)
      runtimeScene.__allStickers = new Set();
  ```

  in the `onCreated`, because otherwise __allStickers was not created when changing scenes.
  
I think it's ready now, merging it. Thanks for the extension! I'll also add the game as an example :)

D8H:
Thanks for the review. I make similar changes to the ObjectStack extension.

4ian:
Thank you!

PR #218, opened 2021-09-14
ObjectStack extension for card games

D8H:
## Description

An ordered list of objects and a shuffle action

It provides:
* Actions to modify a stack of objects
* Conditions to access the objects of a stack
* A shuffle action

It can be helpful for:
* Card games
* Fair randomness (for instance, to create a stack of predetermined bonus and random the order they appear)

## Card System Example
* Short video: https://user-images.githubusercontent.com/2611977/133291953-b2f49035-17eb-472d-902e-8594d7a2555b.mp4
* Build: https://games.gdevelop-app.com/game-b6751124-53fe-4d08-a180-35aec55de4c1/index.html
* Project: https://www.dropbox.com/s/t7ltigvjzvgq4k6/card-system.zip?dl=1

CardExample


4ian:
Looks awesome! I've not checked the internals yet, but the video is very impressive. Definitely looks like a great example to add :)
Remind me to also review and merge your Sticky extension if not done by tomorrow evening!

4ian:
Finally got the time to review this! :) 

I think the main thing is the usage of `gdjs.evtTools.object.pickObjectsIf`. I've seen it's used in `Contains`, `ContainsBetween`, `HasOnTop`. This function is usually used to implement free conditions (not behavior conditions) because it will *filter* the object list maps that are passed as argument.

In the case of your extension, this means that if for example the `Contains` behavior condition is called for two objects having the behavior (let's imagine we do "Card is in the stack of MyStack", and there are two "MyStack" objects). 
The first MyStack `Contains` method will properly call `gdjs.evtTools.object.pickObjectsIf`, and filter the objects passed as parameters. 
The second MyStack `Contains` method will also call `gdjs.evtTools.object.pickObjectsIf`, but the objects will be already filtered.

I don't have a good solution for now - but I have the feeling we're exposing some strange behaviors here if we have more than one "stack object" having the behavior and used at once. Not sure if we can do better, but I can already see confused users asking themselves why only some cards in their stacks are picked by the condition when they have multiple stacks 🤔

Does this make sense?

4ian:
Actually I do have a good solution! Which is to use the API made to filter two list of objects at once: 
`gdjs.evtTools.object.twoListsTest`

This means that the behavior methods that are using `pickObjectsIf` should be moved to free functions outside the behavior. In these functions, you take a list of objects (as a parameter) with this behavior (as another parameter), and the list of other objects (as another parameter).

You can then call `gdjs.evtTools.object.twoListsTest` and pass the object lists and write a predicate that verifies that the first object passed as parameter (which will be a stack) contains or not the second one. That should be enough and work in all cases.

Let me know if you give this a try. I can also take a try at this if you wish!

D8H:
> Actually I do have a good solution! Which is to use the API made to filter two list of objects at once:
> `gdjs.evtTools.object.twoListsTest`
> 
> This means that the behavior methods that are using `pickObjectsIf` should be moved to free functions outside the behavior. In these functions, you take a list of objects (as a parameter) with this behavior (as another parameter), and the list of other objects (as another parameter).
> 
> You can then call `gdjs.evtTools.object.twoListsTest` and pass the object lists and write a predicate that verifies that the first object passed as parameter (which will be a stack) contains or not the second one. That should be enough and work in all cases.
> 
> Let me know if you give this a try. I can also take a try at this if you wish!

Ok, thanks, I'll try this.

D8H:
I committed to save a version before using `twoListsTest`.

D8H:
I updated the 2 examples to use the new conditions.

4ian:
These new conditions look great! 
Last stuff though which is a bit of a problem: the fact that we're storing `gdjs.RuntimeObject` directly means that we have to handle clearing them from the stack if we destroy one - otherwise we have a "dangling pointer" to one, and the object (and its renderer, and its behaviors, and its variables...) will survive in memory.

I'm not sure what's the best solution, surely to hook again like you did in the other extension to a gdjs callback to listen to object destruction? This might be a pain because you have to inspect all the stacks... or keep a map of "object id to their stack object" - so that whenever an object is deleted, you can quickly get its stack (if any) and remove it?

If this kind of behavior is becoming more common, we could introduce something in GDJS directly, something like a "RuntimeObjectWeakReferenceMap" (sorry for the long name, but you get the idea), that would help doing this for an extension (storing references to objects, and automatically dropping them if the object is deleted).

D8H:
> I'm not sure what's the best solution, surely to hook again like you did in the other extension to a gdjs callback to listen to object destruction? This might be a pain because you have to inspect all the stacks... or keep a map of "object id to their stack object" - so that whenever an object is deleted, you can quickly get its stack (if any) and remove it?

An inverted index will allow to make the conditions more efficient anyway. It will also allow to ensure uniqueness too, because I think allowing to add the same instance in one stack twice is probably not a good idea.

arthuro555:
I wanted to make a semi-related feature request for a long while now actually. A very common pattern with behaviors is to use some of them to "Mark" some objects to be used by another one, similarly to an object group. See for example the LightObstacle, Platform, and Pathfinding obstacle behaviors. Making those with events is possible via the usage of scene variables, but is quite tricky and limited.

To make that easier, my idea is to expose to all functions in extensions for each behavior of the extension a group that references all objects on the scene with that behavior. They would work like normal object groups, except that instead of using `runtimeScene.getObjects`, it would use another function `runtimeScene.getBehaviorObjects(behaviorname: string)` that 
1. Get all objects that have the behavior from the game project data
2. Get them all via `runtimeScene.getObjects`
3. Filter then by behavior activation
4. Return that objects list 😎 

The type of that object group would be the same as the Object parameter in the Behavior functions.  
This eliminates memory leaks risks as objects list are cleared after each frame anyways, and gives events new powerful features.

This would allow for many powerful things: 
- Being able to use that popular pattern of marking objects to be used by an extension easily (and also store object-specific state via the behavior properties ^^)
- Create stuff like the stick extension via events & 2 behaviors Sticker and Stickeable, by linking the objects Sticking to a Stickeable when sticking, then set the position of all Sticking to Stickeable.X() while taking into account linked objects each frame
- Allow making behavior that needs access to all objects that have it for updating it, for example, if I want to make an extension that sorts by the value of the object variable `weight`, I can make an extension wide post events callback that will be able to know the weight of all objects at once instead of using a behavior update method that only knows about its own object instance

4ian:
> An inverted index will allow to make the conditions more efficient anyway. It will also allow to ensure uniqueness too, because I think allowing to add the same instance in one stack twice is probably not a good idea.

Sounds good! 

> I wanted to make a semi-related feature request for a long while now actually. 

Let's talk about this in https://github.com/4ian/GDevelop/discussions/3001

My main concern is that naively iterating on behaviors will lead to bad performances as soon as the number of objects grow.

D8H:
Actually, I only used sets because an inverted index would make it a lot more complex for little value.
* A global Set contains every living objects that has been into a stack allows to check in O(1) deleted objects that never been in a stack.
* A Set for each stack allows to check deleted objects that once been in a stack in O(number of stacks)

4ian:
This problem should be gone because the registry does not need to be rebuilt and pushed anyway now.
Let's see if we can rebase this on main: !rebase

arthuro555:
probably an issue due to the branch to merge from being on a fork, but actions run on the main repository

4ian:
Thank you! 👍👍

Issue #184, opened 2021-07-16
3D Flip


VegeTato:
![3D flip extension](https://user-images.githubusercontent.com/81715237/125967031-9e9cb69e-f6e9-4e38-81fa-9ce387134aad.gif)

This extension is useful to give your game the 3D feeling, yet cute feeling to the players characters.
= = = = = = = = = = = = = = = = = = = = = =
**You can test the extension here:**
https://games.gdevelop-app.com/game-92062a3c-ea9d-4e14-9454-d375dca71ba1/index.html

Version 0.1
-------------------
-User can set the 3D flip speed
-User can set the object Width
-User can set the Right and Left keys
-User can change the ("Origin" point) to object center for the best results

Version 0.2
-------------------
-Added condition to check IF the object is 3D flipped (Helpful to make sound when object/player is 3D flip)

Version 0.3
-------------------
-User don't have to set the width anymore

## Checklist
- [done] Extension has a proper name and description.
- [done] Extension has tags (for example: "platform, brick, breakable").
- [done] All behaviors have a description.
- [done] All functions (actions, conditions, expressions) have descriptions and proper sentences.
- [done] I've followed the best practices listed on http://wiki.compilgames.net/doku.php/gdevelop5/extensions/best-practices
- [done] I confirm that this extension can be integrated to this GitHub repository, distributed and MIT licensed.

**You can download the example here:**
https://trello-attachments.s3.amazonaws.com/5fd532f01e99c97130e99fe9/60d0a6d728e99c7fc00c0d71/d8cccda5672fd75e3fdedf3058a3bfbd/3D_Flip_extension_Example_V0.2.rar

**You can download the extension here:**
https://trello-attachments.s3.amazonaws.com/5fd532f01e99c97130e99fe9/60d0a6d728e99c7fc00c0d71/000cd56bf954a99b488bf63b717d3467/3D_Flip_extension_v0.3.rar

arthuro555:
[3D Flip extension Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/6835364/3D.Flip.extension.Example.zip)
[ThreeDFlip.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/6835365/ThreeDFlip.zip)

I have made some changes:
- Converted to a behavior
- Changed some properties to respect the best practices
- Added some actions/conditions

I have made major changes to the extension so I would appreciate another review, if none comes in a few days I'll add it.

[...]

PR #224, opened 2021-09-22
[Flip3D] Duration instead of speed, trigo, revert at any time.

D8H:
Changes
* The speed property is replaced by a duration parameter to use it like Tweens.
  This is a breaking change so the version is incremented to 1.0.0
* The flipping uses a cosinus to make it more realistic.
* A flipping animation can be revert at any time.

* An example of animation on cards (position tween + flipping)
  https://www.dropbox.com/s/0v113sqm2vi9kds/Klondike.zip?dl=1
* A comparison with the previous version on the original example (the flip duration is long to let see what happens)
  https://www.dropbox.com/s/3xh24b5bnmbgi4d/3D%20Flip%20extension%20Example.zip?dl=1

Extension thread: https://github.com/GDevelopApp/GDevelop-extensions/issues/184

### DIFF
--- a/Extensions/ThreeDFlip.json
+++ b/Extensions/ThreeDFlip.json
@@ -308,7 +308,7 @@
           "objectGroups": []
         },
         {
-          "description": "Start a flipping animation on the object.",
+          "description": "Start a flipping animation on the object",
           "fullName": "Flip the object",
           "functionType": "Action",
           "name": "Flip",
@@ -649,7 +649,7 @@
           "objectGroups": []
         },
         {
-          "description": "Stops flipping the object.",
+          "description": "Stops flipping the object",
           "fullName": "Stop the flip",
           "functionType": "Action",
           "name": "StopFlip",
@@ -779,7 +779,7 @@
           "objectGroups": []
         },
         {
-          "description": "Checks if a flipping animation is currently playing.",
+          "description": "Checks if a flipping animation is currently playing",
           "fullName": "Flipping is playing",
           "functionType": "Condition",
           "name": "IsFlipping",
@@ -843,7 +843,7 @@
           "objectGroups": []
         },
         {
-          "description": "Checks if the object has been flipped.",
+          "description": "Checks if the object has been flipped",
           "fullName": "Is flipped",
           "functionType": "Condition",
           "name": "IsFlipped",
@@ -907,7 +907,7 @@
           "objectGroups": []
         },
         {
-          "description": "Flips the object to one specific side.",
+          "description": "Flips the object to one specific side",
           "fullName": "Flip to a side",
           "functionType": "Action",
           "name": "FlipTo",
@@ -1168,7 +1168,7 @@
             {
               "codeOnly": false,
               "defaultValue": "",
-              "description": "duration (in milliseconds)",
+              "description": "Duration (in milliseconds)",
               "longDescription": "",
               "name": "Duration",
               "optional": false,
--- a/Extensions/ThreeDFlip.json
+++ b/Extensions/ThreeDFlip.json
@@ -1,24 +1,24 @@
 {
-  "author": "VegeTato",
-  "description": "Helpful to give the feeling of 3D to your object character and your game.\nThe origin point must be at the center for the best results.",
+  "author": "VegeTato, D8H",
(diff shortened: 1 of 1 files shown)


VegeTato:
> 
> 
> Changes
> 
>     * The speed property is replaced by a duration parameter to use it like Tweens.
>       This is a breaking change so the version is incremented to 1.0.0
> 
>     * The flipping uses a cosinus to make it more realistic.
> 
>     * A flipping animation can be revert at any time.
> 
>     * An example of animation on cards (position tween + flipping)
>       https://www.dropbox.com/s/0v113sqm2vi9kds/Klondike.zip?dl=1
> 
>     * A comparison with the previous version on the original example (the flip duration is long to let see what happens)
>       https://www.dropbox.com/s/3xh24b5bnmbgi4d/3D%20Flip%20extension%20Example.zip?dl=1
> 
> 
> Extension thread: #184

Perfect !
the extension is now even more flexible ❤️
i would add a long description below the duration parameter that says (Example: 200), so the user understand how the duration works for the best use of extension.

D8H:
> i would add a long description below the duration parameter that says (Example: 200), so the user understand how the duration works for the best use of extension.

I don't think that necessary. I mean "Duration (in milliseconds)" is pretty clear to me.

VegeTato:
> 
> 
> > i would add a long description below the duration parameter that says (Example: 200), so the user understand how the duration works for the best use of extension.
> 
> I don't think that's necessary. I mean "Duration (in milliseconds)" is pretty clear to me.

it is clear for the value yes, but i mean a (Recommend: 190) can let the user set it to 190 because he will think that's the best value in the extension owner opinion (which it is), if he set it to 1k (just like in your example, which is not how i built the extension to be like that) the extension will looks so bad and the user might even not know whats the extension about or might think its broken, for me i always look at the recommendation in the extensions and i always set that value when its just for testing to see whats the extension is about, and the last thing is rather than leaving the user to try the best values till he finds it, we just give him the best value already to test it out, and he can adjust it starting from there, which is 190.

Note: i checked the numbers again and 190 works smoother than 200.

D8H:
> it is clear for the value yes, but i mean a (Recommend: 190) can let the user set it to 190 because he will think that's the best value in the extension owner opinion (which it is), if he set it to 1k (just like in your example, which is not how i built the extension to be like that) the extension will looks so bad and the user might even not know whats the extension about or might think its broken, for me i always look at the recommendation in the extensions and i always set that value when its just for testing to see whats the extension is about, and the last thing is rather than leaving the user to try the best values till he finds it, we just give him the best value already to test it out, and he can adjust it starting from there, which is 190.
> 
> Note: i checked the numbers again and 190 works smoother than 200.

I see. But, it's a good value for the specific case where it's used on a platformer character, isn't it? I'm not sure that it will be the common usage of the extension. I mean I've never seen a platformer do this. I think it will be great to do animations for collectable or things like this and 1k will be a good value then.

VegeTato:
> 
> 
> > it is clear for the value yes, but i mean a (Recommend: 190) can let the user set it to 190 because he will think that's the best value in the extension owner opinion (which it is), if he set it to 1k (just like in your example, which is not how i built the extension to be like that) the extension will looks so bad and the user might even not know whats the extension about or might think its broken, for me i always look at the recommendation in the extensions and i always set that value when its just for testing to see whats the extension is about, and the last thing is rather than leaving the user to try the best values till he finds it, we just give him the best value already to test it out, and he can adjust it starting from there, which is 190.
> > Note: i checked the numbers again and 190 works smoother than 200.
> 
> I see. But, it's a good value for the specific case where it's used on a platformer character, isn't it? I'm not sure that it will be the common usage of the extension. I mean I've never seen a platformer do this. I think it will be great to do animations for collectable or things like this and 1k will be a good value then.


oh no, i mean it can be used for any sprite as animation, but its perfect for platformer, it gives the 3D feeling to the platformer game more like paper character, and 1k is way too slow for that, this extension is made for platformer not animations, but if the user wants to use it as an animation (which i already included in my animation extension) well he can then change the value from 190 to anything else, i posted this on discord WIP before, and the community was supportive as a platformer character, since it gives a new feeling to the game rather than just flip normally.

Bouh (approved):

PR #225, opened 2021-09-22
Cancellable Draggable behavior with tweens

D8H:
This behavior add an action to tween back the object to it's original position.
It requires Tween and Draggable behaviors.
There is not many events but I think it can often be useful when using the Draggable behavior.

Example
* a card games: https://www.dropbox.com/s/0v113sqm2vi9kds/Klondike.zip?dl=1
  Press "Back" to skip the animation.


Bouh:
A little update with the easing parameter :)

style

Bouh (approved):

PR #226, opened 2021-09-22
[Sticker] Avoid a 1 frame delay when chaining objects.

D8H:
A small change to ensure that parents are process before their children.
I'm not really prod about `hasBehavior(behaviorName)`, it finds the behavior on the parent according to the child behavior name. Do you think it should look the behavior type instead?
```JavaScript
    prototype.followBasisObject = function () {
        if (this.followingDoneThisFrame) {
            return;
        }
        this.followingDoneThisFrame = true;
        /** @type {gdjs.RuntimeObject} */
        const basisObject = this.basisObject;
        if (basisObject) {
            if (basisObject.hasBehavior(behaviorName)) {
                // Follow parents 1st to avoid frame delays
                basisObject.getBehavior(behaviorName).followBasisObject();
            }
```

## Examples
* The original example running the new version
  https://www.dropbox.com/s/pmfmoqjzpmpw1w2/stick-objects.zip?dl=1
* A card game that drags several cards at once with an intermediate object (because card <-> card is not possible). Without the fix the top card is a bit shifted.
  https://www.dropbox.com/s/0v113sqm2vi9kds/Klondike.zip?dl=1
  Press "Back" to skip the animation.


D8H:
@arthuro555 It looks like the script forbidding to use `gdjs.evtTools.object.twoListsTest`. It passed on another PR: https://github.com/GDevelopApp/GDevelop-extensions/pull/218 is it new?

arthuro555:
That was not introduced by me but by 4ian. This is to ensure that any extension doing manual object picking is carefully reviewed. Afaiu the goal of the CI is to ultimately pretty much merge stuff without downloading the extension and carefully reviewing it if it passes CI and the changed files don't look odd (for example obfuscated javascript added somewhere), to make reviewing faster and less exhausting.

Bouh:
@D8H [You can add a line here](https://github.com/GDevelopApp/GDevelop-extensions/blob/main/scripts/lib/ExtensionsBestPracticesExceptions.js#L159) to bypass this limitation.
As Arthuro555 said we're now a bit more carefully, it also to ensure people use as possible GDJS API.

Bouh:
Lgtm!

PR #228, opened 2021-09-23
[Slider] Value range and tick spacing

D8H:
This add 3 properties (it doesn't break compatibility)
* Minimum value (0 by default)
* Maximum value (1 by default)
* Tick spacing (0 = none by default)

Properties, actions and parameters name now better follow recommendations.

User input handling has been moved from post to pre (it was already separated from drawing in the previous update)

* Fog of war example: https://www.dropbox.com/s/jt7b2qmp5di33ny/marching-square-fog-of-war.zip?dl=1

### DIFF
--- a/Extensions/DraggableSliderControl.json
+++ b/Extensions/DraggableSliderControl.json
@@ -1,14 +1,14 @@
 {
-  "author": "Tristan Rhodes (https://victrisgames.itch.io/)",
-  "description": "The `Value` expression will return the position of the slider on the bar, from 0 to 1, where 0 is the left-side, and 1 is the right-side.\n\nYou can [watch this tutorial](https://youtu.be/iiTUwdAT_hs) showing how to use this extension - remember to attach it to a **Shape Painter** object.",
+  "author": "Tristan Rhodes (https://victrisgames.itch.io/), D8H",
+  "description": "This extension provides a behavior that draws a horizontal slider that can be dragged by the users. The value range and the tick spacing can be defined with properties.\nNote that this behavior can only be attached to a **Shape Painter** object.\n\nFurther details can be found in [this tutorial video](https://youtu.be/iiTUwdAT_hs).\n",
   "extensionNamespace": "",
   "fullName": "Draggable Slider Control",
   "helpPath": "",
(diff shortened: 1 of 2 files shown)


arthuro555:
@tristanbob Since you made this extension, would you like to review the changes?

tristanbob:
> @tristanbob Since you made this extension, would you like to review the changes?

Yes, I can review it tonight.  Thanks for these improvements, @D8H!

tristanbob:
@D8H:

I really like the changes you have made! (some of these are from previous updates)
- Not drawing the objects every frame GREATLY reduces performance impact 
- Providing a custom min/max value is a great idea (instead of doing tedious calculations for every slider)
- The tick spacing is also a great idea (instead of manually using the round() function, which can confuse new users)

**Issues:**

1: The tick spacing works most of the time, but your "0.01" tick can sometimes show a long number that is very slightly off (see video to see the problem). Do you have any ideas on what is happening, and how to fix it?

https://user-images.githubusercontent.com/8879811/134613193-75c0db60-ebae-4532-a665-9b9911e08149.mp4


2:  You need to move the "Prepare thumb settings" events above drawing the circle thumb (see screenshot for correct placement). Otherwise, thumb settings (opacity, color) won't be accurate until the slider is redrawn:

Screenshot 2021-09-23 211923

D8H:
> 1: The tick spacing works most of the time, but your "0.01" tick can sometimes show a long number that is very slightly off (see video to see the problem). Do you have any ideas on what is happening, and how to fix it?

It's because 0.01 is not a round number in binary. It would need a text formatter to avoid it.
Float

D8H:
I made the change asked by Tristan and updated the example.

tristanbob (approved):
LGTM.  Thanks @D8H!

Extensions/DraggableSliderControl.json
```diff
@@ -1,14 +1,14 @@
...
   "name": "DraggableSliderControl",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Line Hero Pack/Master/SVG/UI Essentials/UI Essentials_sliders_options.svg",
-  "shortDescription": "Use a shape-painter object to draw a horizontal slider that can be dragged by the users. ",
-  "version": "0.4.4",
+  "shortDescription": "A horizontal slider that can be dragged by the users. ",
+  "version": "0.5.5",
```
arthuro555, line 11: According to semver, bumping the minor version should reset the patch version to 0.

PR #230, opened 2021-09-23
Add more extension validation checks

arthuro555:
Adds more checks against the best practices to the extension validation and fix the extensions that did not comply where it did not imply a breaking change.

Checks added:
- Valid semantic versioning
- No expressions prefixed with `Get`
- Usage of PascalCase in internal names (except lifecycle functions)
- Check if names and descriptions are filled out

Additional changes:
- More types 🤓 
- Changed moved all validation code from generate-extensions-repository.js to an extra file
- Prettier output of errors


4ian:
Awesome! I think all of this make sense. Will be useful to ensure best practices are enforced. In the future, we could port more and more of these to the editor itself as some warnings or similar stuff, so that any problem is shown to the extension author directly.

PR #234, opened 2021-09-26
Fixed action Load the Leaderboard entries (YandexGamesSDK extension)

achubutkin:
Hi! There was a bug in the action Load the Leaderboard entries.


Bouh:
Thank you for the fix!

PR #235, opened 2021-09-26
Add github actions to make reviewing of extensions easier

arthuro555:
- Creates a PR for extension submission issues automatically
- Adds PR extension submissions to the Github Project automatically
- Post results of failed extensions automatic validation in PR discussion
- Deletes autoclose for invalid submissions as the form template prevents invalid/empty issues to be submitted in the first place
- Adds commands for making submissions easier to update and merge
  - `!rebase`: Merges master and regenerates extensions-registry.json
  - `!update <https://link.to/extension.zip>`: Updates the extension in the PR and regenerates extensions-registry.json

The actions are live on my fork (https://github.com/arthuro555/GDevelop-extensions), feel free to try them out there.


.github/workflows/auto-rebuild.yml
```diff
@@ -0,0 +1,38 @@
...
+          echo "CHANGED="$(git diff --name-only extensions-registry.json) >> $GITHUB_ENV
+
+      - name: Commit and push if changed
+        if: ${{ contains(env.CHANGED, 'extensions-registry') }}
+        run: |
+          git config --global user.name '${{ github.actor }}'
+          git config --global user.email '${{ github.actor }}@users.noreply.github.com'
+          git add extensions-registry.json
+          git commit -m "Regenerated extension registery"
+          git push
```
arthuro555, line 28: This step never triggers for a reason I cannot find. I've left it here as it cannot hurt and may make it easier to add this in if someone can figure out why it does not trigger.

4ian:
Wow! 

Everything looks awesome. I'll check that tomorrow after the conference we're at :) 

Definitely this will speed up our process, so we can concentrate on getting better extensions!

.github/workflows/auto-rebuild.yml
4ian: Ok, we'll probably remove this json soon as only useful for old GDevelop versions.

ClementPasteau: @4ian not sure what you mean, the `extensions-registry.json` is still being used by the latest version of GDevelop. (fetched and enriched)

We could modify our Extension service to fetch the extensions from the `/Extensions` folder rather than this json, and generate it for the app. Is this what you have in mind?

4ian: The latest versions are using https://resources.gdevelop-app.com/extensions-database/extensions-database.json - which is generated inside the `dist` folder when you run `npm run build` locally (and then deployed).
But not extensions-registry.json directly :) (unless I missed something else). This file is still in the git because some versions ago the registry was directly fetched by the editor from github. But now all calls should be going through the API, and from the very latest version it is even returned directly by the API (rather than a redirection).

ClementPasteau: ooooh right :+1: I misread the name of the file :)

PR #247, opened 2021-10-03
Add more authors in new format

arthuro555:
Converts to the new format extension authors that claimed a name (VictrisGames, VegeTato, TheGemDev and MrJeje/Jérémy Leriche).


4ian:
🚀

PR #256, opened 2021-10-09
[Turret] Add action "Change angle toward a position"

D8H:
The action allow to "teleport" the direction. It's useful to handle mouse or touch events.
This extension update is already included in the last version of BimBam example.


4ian:
Thanks!

PR #275, opened 2021-11-03
correction in sentence of the extension FlashTransitionPainter

westboy31:
Change PARAM4 by PARAM5 (which refers to direction) in the sentence of the action PaintEffect


4ian:
Thank you! 👍

PR #279, opened 2021-11-05
Add punctuation rule

arthuro555:
Adds a rule for descriptions to require a dot and some names (like extension full names) to disallow dots.
Also adds Auto-Fixing capabilities to the extension validator.


Extensions/YandexGamesSDK.json
```diff
@@ -1009,7 +1009,7 @@
         {
           "codeOnly": false,
           "defaultValue": "",
-          "description": "Data key name (e.g. \"Lives\")",
```
4ian, line 1012: Any reason for changing this? It seems `e.g.` is a proper form, and a bit more explicit than writing "eg" (I don't see it often, and it looks like "egg" with a missing g :)). https://www.aje.com/arc/editing-tip-using-eg-and-ie/

Extensions/TravelToRandomPositions.json
```diff
@@ -29,7 +29,7 @@
       "functionType": "Action",
       "name": "TravelToRandomPositions",
       "private": false,
-      "sentence": "Move _PARAM1_ to random positions, where each stop is at least _PARAM3_px and at most _PARAM4_px from the current position. Move towards angle _PARAM5_ with a biais of _PARAM6_.",
```
4ian, line 32: Would be worth adding either a "and" or a "-" or something.

Extensions/ShakeObject.json
```diff
@@ -3600,7 +3600,7 @@
             {
               "codeOnly": false,
               "defaultValue": "",
-              "description": "Duration of shake (in seconds) (Default: 0.5)",
```
4ian, line 3603: The fix is wrong here.

Extensions/ShakeObject.json
```diff
@@ -3123,12 +3123,12 @@
       "objectType": "Sprite",
       "eventsFunctions": [
         {
-          "description": "Shake an object, using one or more ways to shake (position, angle, scale)",
+          "description": "Shake an object, using one or more ways to shake (position, angle, scale).",
           "fullName": "Shake object (position, angle, scale)",
           "functionType": "Action",
           "name": "ShakeObject_PositionAngleScale",
           "private": false,
-          "sentence": "Shake object _PARAM0_ for _PARAM2_ seconds. Modify position amplitude _PARAM3_ on X axis and _PARAM4_ on Y axis, angle rotation amplitude _PARAM5_, and scale amplitude _PARAM6_.  Wait _PARAM7_ seconds between shakes.  Keep shaking until stopped: _PARAM8_",
```
4ian, line 3131: Should be replaced or not automatically fixed.

scripts/lib/rules/NoGet.js
```diff
@@ -5,13 +5,16 @@ const {
...
+  for (const func of publicEventsFunctions) {
+    const { name, functionType } = func;
     if (
       (functionType === 'Expression' || functionType === 'StringExpression') &&
       name.startsWith('Get') &&
       !legacyGetPrefixedExpressionsExtensions.has(extension.name)
     )
-      onError(`Expression '${name}' is using prohibited 'Get' prefix!`);
+      onError(`Expression '${name}' is using prohibited 'Get' prefix!`, () => {
+        func.name = func.name.replace('Get', '');
```
4ian, line 16: This could break the extension if the expression is used inside the extension? We should probably error and not fix anything.

4ian:
It's a good idea to add these checks and auto fix.
Careful though, I found several instances of places where dots have been wrongly removed: sentences (need manually fixes), or descriptions (0.08). This means your fixes/checks are too broad, and must be done on more specific checks (only check/remove/add dots if they are at the end, use regular expressions, etc...) - to avoid this having the opposite effect of what we want ;)

arthuro555:
Ok, I've relaxed the rules a lot - now only single dots at the end are disallowed, not all dots. I don't see any wrong dot removal anymore, please tell me if I missed an edge case :)
Sorry for taking so long 😅

4ian (approved):

PR #298, opened 2021-11-19
[Stay On Screen] Fix max X and max Y of the object position

Bouh:
Tested in fullscreen and windowed with the hitbox debugger, it's works fine now.


4ian:
This won't work for objects having their origin not at 0;0 though right?

Bouh:
> This won't work for objects having their origin not at 0;0 though right?

Indeed like on the existing live version.

4ian:
Ok, let's go ahead with this but we should rework this and the expressions we have to expose the AABB of an object I think.
I think we had other extensions wanting to compute the bounds/the center of an object and they could not, because CenterX/Y can also be changed as it's the center of rotation.

PR #307, opened 2021-12-01
[MarchingSquare] No longer modify the ShapePainterRuntimeObject prototype.

D8H:
The dimensions getters are only override for the owner of the behavior.
This code is part of the method that updates the hitboxes according to the scalar field.
```
      // Hack for the "separate" action that calculates a bounding disk with dimensions.
      shapePainter.width = this.dimX() * this._getCellWidth();
      shapePainter.height = this.dimY() * this._getCellHeight();
      shapePainter.getWidth = function () {
        return this.width;
      }
      shapePainter.getHeight = function () {
        return this.height;
      }
```
This is a review change that was missing from:
* https://github.com/GDevelopApp/GDevelop-extensions/pull/288


arthuro555:
👍 thanks!

PR #310, opened 2021-12-05
[ColorConversion] Weighted mean between 2 RGB colors.

D8H:
## Changes

* Add an expression that applies a weighted mean between 2 RGB colors.

```JavaScript
try {
    // split do allocations, but the code a lot simpler this way.
    const components1 = eventsFunctionContext.getArgument("Color1").split(";");
    const r1 = gdjs.evtTools.common.clamp(parseFloat(components1[0]), 0, 255);
    const g1 = gdjs.evtTools.common.clamp(parseFloat(components1[1]), 0, 255);
    const b1 = gdjs.evtTools.common.clamp(parseFloat(components1[2]), 0, 255);

    const components2 = eventsFunctionContext.getArgument("Color2").split(";");
    const r2 = gdjs.evtTools.common.clamp(parseFloat(components2[0]), 0, 255);
    const g2 = gdjs.evtTools.common.clamp(parseFloat(components2[1]), 0, 255);
    const b2 = gdjs.evtTools.common.clamp(parseFloat(components2[2]), 0, 255);

    const ratio = eventsFunctionContext.getArgument("Ratio");
    const r = gdjs.evtTools.common.clamp(Math.round((1 - ratio) * r1 + ratio * r2), 0, 255);
    const g = gdjs.evtTools.common.clamp(Math.round((1 - ratio) * g1 + ratio * g2), 0, 255);
    const b = gdjs.evtTools.common.clamp(Math.round((1 - ratio) * b1 + ratio * b2), 0, 255);

    eventsFunctionContext.returnValue = r + ";" + g + ";" + b;
} catch (e) {
    console.error("Unable to convert a color: ", e);
    eventsFunctionContext.returnValue = "0;0;0";
}
```

## Project example

This example uses it to make shades from green to beige.
* Project: https://www.dropbox.com/s/htn3nixly0pum77/marching-squares-terraforming.zip?dl=1
* Build: https://d8h.itch.io/marching-squares-terraforming


D8H:
@tristanbob, @Add00, you worked on this extension. Do you want to be part of the review?

Extensions/ColorConversion.json
```diff
@@ -58,6 +57,58 @@
...
+          "parameterObjects": "",
+          "useStrict": true,
+          "eventsSheetExpanded": true
+        }
```
arthuro555, line 76: Is it truly necessary to use JavaScript here? Unless there is a very big impact on convenience (usage of a complex premade JavaScript code to not rewrite it), performance, or feasibility (for example when requiring APIs not exposed to GDevelop events), JavaScript should be avoided, since it makes the extension harder to review, modify and understand for all non-JavaScript-using users, which constitute the majority of the community. 
It is also written in the best practices: 
image

4ian: Because this is a "tool" extension, might be worth keeping JS here.

D8H: The rest of the extension is in JS too, it was easier to copy-paste code.

4ian:
Merging as it looks good to me!

tristanbob:
> @tristanbob, @Add00, you worked on this extension. Do you want to be part of the review?

@D8H I did a quick search and saw that averaging two color values accurately requires squaring the numbers first, and returning the square root of the sum.

"Instead of summing up the components of the RGB color, sum their squares instead. Then when returning the average color, for each color component, find the mean of the sum (of squares) and return its square root."

https://sighack.com/post/averaging-rgb-colors-the-right-way

I'm not certain if this applies to your function (weighted mean), but I thought I would mention it to you.

D8H:
> @D8H I did a quick search and saw that averaging two color values accurately requires squaring the numbers first, and returning the square root of the sum.
> 
> "Instead of summing up the components of the RGB color, sum their squares instead. Then when returning the average color, for each color component, find the mean of the sum (of squares) and return its square root."
> 
> https://sighack.com/post/averaging-rgb-colors-the-right-way
> 
> I'm not certain if this applies to your function (weighted mean), but I thought I would mention it to you.

It's very interesting. I did a plot with Sage and the difference between quadratic and arithmetic means can be huge with 2 colors far from each other.
arithmetic-vs-quadratic

I will check what it does with actual colors, but according to what I read the quadratic is indeed a better way.

There is also a perceptually uniform space that might be interesting to check:
* https://en.wikipedia.org/wiki/CIELAB_color_space

D8H:
I did a comparison with a Gaussian shading and, with the quadratic mean, it looks indeed like a Gaussian whereas, with the arithmetic mean, it looks linear and it shouldn't.

* Quadratic

quadratic

* Arithmetic

arithmetic

@tristanbob Thanks a lot, I was planning to do diagrams for the marching squares documentation and it will help to make them a lot better.

PR #319, opened 2021-12-10
Update render to sprite extension

arthuro555:
## Version 0.0.2

 - Fixes bug that created a new RenderTexture on each call eventually crashing WebGL when used every frame
 - Fixes sprites not keeping their original height and width when rendering a scene or layer into them
 - Fixes culling not being updated before rendering

### DIFF
--- a/Extensions/RenderToSprite.json
+++ b/Extensions/RenderToSprite.json
@@ -10,6 +10,8 @@
   "shortDescription": "Renders an object, layer or scene and puts the resulting image into a sprite.",
   "version": "0.0.2",
   "tags": [
+    "experimental",
+    "non-public api",
     "visual",
     "sprite",
     "viewport",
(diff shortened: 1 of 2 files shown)


4ian:
I trust you for the fixes :) This being said this is starting to use non public API. As this is very dangerous, we should mark the extension as experimental and have a "non public apis" tag too I think?

arthuro555:
Maybe 🤔 It is kinda sad in my opinion though to not have the possibility to make this extension work in a "stable" way. On the other hand, I cannot think of any other extension that could make use of this API and don't think it would be a good idea to make it public just for that. I'll add the "experimental" and "non-public api" tag.

4ian:
Well, that's the interesting part actually - and why GDevelop is open-source ;) Finding how we could adapt the runtime scene to allow such rendering.
Of course if it's for a single extension we might not seen the exact best API, but it's still hinting in what we could do.

PR #320, opened 2021-12-11
[ExtendedMath] Fix the "Map" function.

D8H:
### Changes
* Fix a bug in the "Map" function `* (max2 - min1)`
* Move the multiplication first because the ranges will probably be wider than 1 and it's better to "divide" a rounding error rather than "multiplying" it.
* Remove keywords that were not relevant.

### Differences
* New
  ```JavaScript
  eventsFunctionContext.returnValue = (n - min1) * (max2 - min2) / (max1 - min1) + min2;
  ```

* Old
  ```JavaScript
  eventsFunctionContext.returnValue = (n - min1) / (max1 - min1) * (max2 - min1) + min2;
  ```

### DIFF
--- a/Extensions/ExtendedMath.json
+++ b/Extensions/ExtendedMath.json
@@ -8,14 +8,12 @@
   "name": "ExtendedMath",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/function-variant.svg",
   "shortDescription": "Additional math functions and constants expressions.",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "tags": [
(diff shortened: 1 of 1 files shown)


arthuro555:
🤔 Would that count as a breaking change?

D8H:
> 🤔 Would that count as a breaking change?

No the contract stays the same. It's just it was not fulfill before this fix.
I guess that the bug hasn't been found before now because it was used with min1 == min2 (probably 0).

4ian:
Considering this as a bug fix and merging now! Thanks :)

PR #322, opened 2021-12-12
[Noise] Add octaves handling to make procedural generation easier.

D8H:
## Changes

Tristan's changes:
* Add actions and expressions for frequency, octaves, persistence and lacunarity.
* Add the calculus with these "properties" for simplex2.

D8H's changes:
* Remove the amplitude "property" because it was useless (it was a new property).
* Fix the "properties" default values.
* Factorize the calculus to apply it to every noise functions.
* Optimize the calculus a bit.
* Remove unnecessary protections on the "properties" values.
* Call `seeding` only when the seed is changed and at the beginning.
* Add explanations in the description of the "properties" setter actions.
* Rewrite functions full name, description and phrase to fit to the GDevelop writing style.
* Change the parameter case to fit the good practices.

In the future, wraparound could be made easier but a 2-axis wraparound need 4d noise which means that another lib should be used. For instance, this one:
* https://github.com/jwagner/simplex-noise.js/blob/main/simplex-noise.ts

Note that with the default properties, the noise expressions give the same values as in the previous version. So, there is no breaking change.

## Example

* Build: https://d8h.itch.io/island-generator
* Project: https://www.dropbox.com/s/eybx7ytzctk8bq1/marching-squares-island-generator.zip?dl=1

## Code changes

```JavaScript
//default values
gdjs._extensionNoise = {
  seed: gdjs.randomInRange(1, 65536),
  frequency: 1,
  octaves: 1,
  persistence: 0.5,
  lacunarity: 2,
}
```
```JavaScript
  gdjs._extensionNoise.noise = function(noiseFunction, x, y, z) {
    const lacunarity = gdjs._extensionNoise.lacunarity;
    const octaves = gdjs._extensionNoise.octaves;
    const persistence = gdjs._extensionNoise.persistence;

    let frequency = gdjs._extensionNoise.frequency;

    let noiseSum = 0;
    let amplitudeSum = 0;
    let amplitude = 1;
    for (let i = 0; i < octaves; i++) { 
        noiseSum += noiseFunction(x * frequency, y * frequency, z * frequency) * amplitude;
        amplitudeSum += Math.abs(amplitude);
        amplitude *= persistence;
        frequency *= lacunarity;
    }
    return noiseSum / amplitudeSum;
  };
```


D8H:
@Add00 Do you want to be part of the review?

Bouh:
This version of the extension is currently used in the recently approved example.

PR #323, opened 2021-12-12
[MarchinSquare] Allow to set each field value independently.

D8H:
## Changes

* Add an action to set the field value of a grid point.
* Add expressions for the grid dimension (the number of rows and columns).
* Add an action to "clamp" the field with an upper and lower bound.
* Add the "Multiplication" and "Division" operations for the "merge" action.
* Fix: the field is cleared if its size is changed so that its size actually updates.
  This might not be the best way to handle size changes but it avoids bugs for now. I will rethink it when I do cropping, shifting, scaling and partial copy.

This allows to use the Noise extension to fill the scalar field with noise and generate a terrain.

## Example

* Build: https://d8h.itch.io/island-generator
* Project: https://www.dropbox.com/s/eybx7ytzctk8bq1/marching-squares-island-generator.zip?dl=1

## Code changes

* Clamp the field
```JavaSript
const behaviorName = eventsFunctionContext.getBehaviorName("Behavior");

const min = eventsFunctionContext.getArgument("Minimum");
const max = eventsFunctionContext.getArgument("Maximum");

const behavior = objects[0].getBehavior(behaviorName);

/** @type {Array<number[]>} */
const fieldValues = behavior.scalarFieldValues;

for (let rowValues of fieldValues) {
    for (let x = 0; x < rowValues.length; x++) {
        rowValues[x] = gdjs.evtTools.common.clamp(rowValues[x], min, max);
    }
}
```

* Set field value
```JavaSript
const behaviorName = eventsFunctionContext.getBehaviorName("Behavior");

const indexX = eventsFunctionContext.getArgument("IndexX");
const indexY = eventsFunctionContext.getArgument("IndexY");
const value = eventsFunctionContext.getArgument("Value");

const behavior = objects[0].getBehavior(behaviorName);

eventsFunctionContext.returnValue = behavior.setFieldValue(indexX, indexY, value);
```

```JavaSript
    /**
     * @param squareX {integer} x grid index
     * @param squareY {integer} y grid index
     * @param {float} the field value
     */
    prototype.setFieldValue = function (
      squareX,
      squareY,
      value
    ) {
      if (squareX < 0 || squareY < 0 || squareX >= behavior.dimX() || squareY >= behavior.dimY()) {
        return;
      }
      /** @type {Array<number[]>} */
      const fieldValues = this.scalarFieldValues;
      fieldValues[squareY][squareX] = value;
    }
```


Bouh:
This version of the extension is currently used in the recently approved example.
Simple merging.

PR #326, opened 2021-12-18
Added vertical slider

tristanbob:
## Changes

- Added new vertical slider behavior
- Updated horizontal slider to use color picker
- Updated horizontal slider to use a list of strings (Rectangle/Circle)
- Added link to test extension:
https://victrisgames.itch.io/draggable-slider-control-extension

## To Do:
- Fix logic on setting Inactive Track color
- Decide if the horizontal version should be renamed with the word "Horizontal"

## Video 

https://user-images.githubusercontent.com/8879811/146629931-5a50f539-8f8f-401a-bcf9-34f7e03d9fa0.mp4

### DIFF
--- a/Extensions/DraggableSliderControl.json
+++ b/Extensions/DraggableSliderControl.json
@@ -8,7 +8,7 @@
   "name": "DraggableSliderControl",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Line Hero Pack/Master/SVG/UI Essentials/UI Essentials_sliders_options.svg",
   "shortDescription": "A draggable slider that users can move to select a numerical value.",
-  "version": "0.7.10",
+  "version": "1.0.0",
   "tags": [
     "draggable",
     "slider",
--- a/Extensions/DraggableSliderControl.json
+++ b/Extensions/DraggableSliderControl.json
@@ -8,7 +8,7 @@
   "name": "DraggableSliderControl",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Line Hero Pack/Master/SVG/UI Essentials/UI Essentials_sliders_options.svg",
   "shortDescription": "A horizontal or vertical slider that can be dragged by the users. ",
-  "version": "0.7.0",
+  "version": "0.7.5",
   "tags": [
     "draggable",
     "slider",
@@ -72,7 +72,7 @@
                     {
                       "type": {
                         "inverted": false,
-                        "value": "DraggableSliderControl::DraggableSliderControl::PropertyIsBeingDragged"
+                        "value": "DraggableSliderControl::DraggableSliderControlVertical::PropertyIsBeingDragged"
                       },
                       "parameters": [
                         "Object",
@@ -197,8 +197,8 @@
                           "parameters": [
                             "Object",
                             "Behavior",
-                            "Object.Behavior::PropertyValueMin() + (Object.Behavior::PropertyValueMax() - Object.Behavior::PropertyValueMin()) * Object.Behavior::PropertyThumbOffset() / Object.Behavior::PropertyTrackWidth()",
-                            "Object.Behavior::PropertyValueMin() + (Object.Behavior::PropertyValueMax() - Object.Behavior::PropertyValueMin()) * Object.Behavior::PropertyThumbOffset() / Object.Behavior::PropertyTrackWidth()"
+                            "Object.Behavior::PropertyValueMin() + (Object.Behavior::PropertyValueMax() - Object.Behavior::PropertyValueMin()) * Object.Behavior::PropertyThumbOffset() / Object.Behavior::PropertyTrackLength()",
+                            "Object.Behavior::PropertyValueMin() + (Object.Behavior::PropertyValueMax() - Object.Behavior::PropertyValueMin()) * Object.Behavior::PropertyThumbOffset() / Object.Behavior::PropertyTrackLength()"
                           ],
                           "subInstructions": []
                         }
@@ -269,6 +269,17 @@
                         "Object"
                       ],
                       "subInstructions": []
+                    },
+                    {
+                      "type": {
+                        "inverted": true,
+                        "value": "DraggableSliderControl::DraggableSliderControlVertical::PropertyDisabled"
+                      },
+                      "parameters": [
+                        "Object",
+                        "Behavior"
+                      ],
+                      "subInstructions": []
                     }
                   ],
                   "actions": [],
@@ -285,7 +296,7 @@
                         "textG": 0,
                         "textR": 0
                       },
-                      "comment": "Detect mouse clicks near track, start dragging",
+                      "comment": "Detect mouse clicks near track, start dragging.  A scene variable is used so that only one slider can be dragged at once.",
                       "comment2": ""
                     },
                     {
@@ -312,7 +323,7 @@
                           "parameters": [
                             "",
                             ">=",
-                            "Object.X() - (Object.Behavior::PropertyThumbHeight() - Object.Behavior::PropertyTrackHeight()) / 2",
+                            "Object.X() - (Object.Behavior::PropertyThumbHeight() - Object.Behavior::PropertyTrackThickness()) / 2",
                             "Object.Layer()",
                             "0"
                           ],
@@ -326,7 +337,7 @@
                           "parameters": [
                             "",
                             "<=",
-                            "Object.X() + (Object.Behavior::PropertyThumbHeight() + Object.Behavior::PropertyTrackHeight()) / 2",
+                            "Object.X() + (Object.Behavior::PropertyThumbHeight() + Object.Behavior::PropertyTrackThickness()) / 2",
                             "Object.Layer()",
                             "0"
                           ],
@@ -340,7 +351,7 @@
                           "parameters": [
                             "",
                             ">=",
-                            "Object.Y() - Object.Behavior::PropertyTrackWidth() - Object.Behavior::PropertyThumbWidth() / 2",
+                            "Object.Y() - Object.Behavior::PropertyTrackLength() - Object.Behavior::PropertyThumbWidth() / 2",
                             "Object.Layer()",
                             "0"
                           ],
@@ -515,7 +526,7 @@
                           "parameters": [
                             "",
                             ">=",
-                            "Object.X() + Object.Behavior::PropertyThumbOffset() - max(Object.Behavior::PropertyThumbWidth(),Object.Behavior::PropertyThumbHeight()) / 2",
+                            "Object.X() + Object.Behavior::PropertyTrackThickness() / 2 - max(Object.Behavior::PropertyThumbWidth(), Object.Behavior::PropertyThumbHeight()) / 2",
                             "Object.Layer()",
                             ""
                           ],
@@ -529,7 +540,7 @@
                           "parameters": [
                             "",
                             "<=",
-                            "Object.X() + Object.Behavior::PropertyThumbOffset() + max(Object.Behavior::PropertyThumbWidth(),Object.Behavior::PropertyThumbHeight()) / 2",
+                            "Object.X() + Object.Behavior::PropertyTrackThickness() / 2 + max(Object.Behavior::PropertyThumbWidth(), Object.Behavior::PropertyThumbHeight()) / 2",
                             "Object.Layer()",
                             ""
                           ],
@@ -543,7 +554,7 @@
                           "parameters": [
                             "",
                             ">=",
-                            "Object.Y() - (Object.Behavior::PropertyThumbHeight() - Object.Behavior::PropertyTrackHeight()) / 2",
+                            "Object.Y() - Object.Behavior::PropertyThumbOffset() - max(Object.Behavior::PropertyThumbWidth(),Object.Behavior::PropertyThumbHeight()) / 2",
                             "Object.Layer()",
                             ""
                           ],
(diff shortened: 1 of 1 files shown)


tristanbob:
I'm not sure if I should leave this as a draft PR or not.  I am done working on it until I get feedback on the two TODO items.

4ian:
> Decide if the horizontal version should be renamed with the word "Horizontal"

This makes the upgrade a bit more difficult, but is I think better in the long term.

D8H:
What I saw so far
* The track dimensions will be more explicit with length/thickness instead of width/height.
* The value should be set to the minimal value at the beginning (I forgot to do it in a previous release, it defaults on 0).

The only difference between the 2 extensions is the preEvent and postEvent where the coordinates are rotated?

I wonder if we could make some improvement of the ShapePainter to handle any rotation angle. The Pixi graphics component can be transformed. The ShapePainter could use this to have a mode like a sprite where rotations and size changes transform the graphics. For the mouse handling, the inverse transformation can be used (either by the extension or with the help of the engine through an expression).

arthuro555:
I wonder if instead of making a new behavior, we shouldn't make the shape painter able to be rotated, that would be easier for users to use and understand and keep the extension more maintainable.

D8H:
> I wonder if instead of making a new behavior, we shouldn't make the shape painter able to be rotated, that would be easier for users to use and understand and keep the extension more maintainable.

Thinking the same thing 😉

D8H:
And it will be useful if we implement something to load a SVG into a ShapePainter one day (I know the SVG spec, it's very clean).

4ian:
In case we handle scaling, we'll also need to allow to define a default size (which would be useful for collisions, rather than having the default 32x32 box which does not make sense for a lot of objects). Though we can maybe add support for rotation without this.

tristanbob:
> > Decide if the horizontal version should be renamed with the word "Horizontal"
> 
> This makes the upgrade a bit more difficult, but one can still rename the old extension to keep it, then install this one, then upgrade progressively their objects to use the new version. So I think it is better in the long term.

To be clear, this is the same extension name but I'm asking about changing the name of the Behavior.  If we do this, and someone performs a manual upgrade of the extension, existing conditions/actions in the game will break because they reference the old behavior name.  (I know because my example game broke when I renamed the behavior)

tristanbob:
> What I saw so far
> 
> * The track dimensions will be more explicit with length/thickness instead of width/height.

Yes, I agree that with that wording exactly!

> * The value should be set to the minimal value at the beginning (I forgot to do it in a previous release, it defaults on 0).

Isn't 0 the correct default value?

> The only difference between the 2 extensions is the preEvent and postEvent where the coordinates are rotated?

There are two primary changes:
1) Change from X to Y
2) Account for the fact that Y axis is reversed from how the horizontal slider worked 

So those required changes to detecting mouse position and for drawing the slider.

> I wonder if we could make some improvement of the ShapePainter to handle any rotation angle. The Pixi graphics component can be transformed. The ShapePainter could use this to have a mode like a sprite where rotations and size changes transform the graphics. For the mouse handling, the inverse transformation can be used (either by the extension or with the help of the engine through an expression).

Good questions...

tristanbob:
After thinking about this a little more, I think I will combine the behaviors into one.  This will make it easier to maintain since most functions are the same (except the two lifecycle functions (pre=logic, post=drawing). 

I will work on merging these behaviors.

D8H:
We can probably wait for this PR to update this extension:
* https://github.com/4ian/GDevelop/pull/3402

D8H:
I have a version of the extension that uses the new shape painter rotation.
If you want I can submit it and you can report your changes over it then?

tristanbob:
> I have a version of the extension that uses the new shape painter rotation that I used for testing. If you want I can submit it and you can report your changes over it then?

Sure, let's try that.  Thanks!

tristanbob:
## Need help with:
- Stop the track from moving when the halo extends past the track
- Figure out why inactive track starts white until a hover or click is detected
- Verify that redraw is only happening when needed
- Provide the right condition to check if the inactive track color has been set or not

## Play it now
https://liluo.io/instant-builds/c6c06f73-3480-4483-a2cd-c711657106ed

## Video
https://user-images.githubusercontent.com/8879811/150627069-fab16ae4-7937-47ea-ba94-8d685b173d50.mp4

## Project files
[RotatedSlider.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/7917733/RotatedSlider.zip)

D8H:
I forgot that I left some events outside of the extension when I did the tests a long time ago.
This is the fixed version:
https://www.dropbox.com/s/tkkj2v3sxgidopf/RotatedSlider_REVIEW.zip?dl=1

I set the origin at the center of the thumb at minimum value and the hitbox where the thumb can go. You will notice strange things with AABB and hitbox when the object is not rotated but it the way GDevelop works: the AABB is around what is drawn at 0° and around the hitbox otherwise.

4ian:
> You will notice strange things with AABB and hitbox when the object is not rotated but it the way GDevelop works: the AABB is around what is drawn at 0° and around the hitbox otherwise.

Correct, the AABB can be slightly different because of an optimization at 0 degree (so in general the AABB should be considered as an approximation of the object but not exactly it)

tristanbob:
## Changes
- Fixed starting colors
- Fixed changing colors
- Adjusted halo to be above tracks but below thumb
- Adjusted logic to prevent malformed sliders
- Fixed some left-over wording of "TrackHeight" to "TrackThickness"

@D8H I'm ready for a full review, I think all known issues are addressed.

## Demo game
https://victrisgames.itch.io/draggable-slider-control-extension

## Project files
[SliderExtension_SourceFiles_2.0.0.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/7919223/SliderExtension_SourceFiles_2.0.0.zip)

## Video
https://user-images.githubusercontent.com/8879811/150654449-3da5e470-18f0-4bdf-b4dd-71d3980d2553.mp4

D8H:
@tristanbob I rebased the branch on the master. If you need to commit something, you will need to pull first.

tristanbob:
> @tristanbob I rebased the branch on the master. If you need to commit something, you will need to pull first.

Did you change anything, or just do the rebase?  I have some minor changes to add, and they conflicting.  I think I can force it to use my version, but I wanted to see if you changed anything.

D8H:
> Did you change anything, or just do the rebase? I have some minor changes to add, and they conflicting. I think I can force it to use my version, but I wanted to see if you changed anything.

I didn't change anything.

Bouh (approved):

D8H:
### Breaking changes
* Rename track width to track length (as track height was renamed into track thickness in the last PR which is also a breaking change)
  Users will loose their length and thickness property values, but this makes the code easier to understand.

### Changes
* Rename "Disable" to "Enable interactions"
* Mark "Enable interactions" as private as it's never used and there is no color properties for disabled sliders.
* Change the behavior description to start with a verb like the built-in extension.
* Sort functions
* Change wording of IsBeingDragged

D8H:
@tristanbob This is the example with the last committed version:
https://www.dropbox.com/s/pqgwno62l3bhd5d/SliderExtension_REVIEW2.zip?dl=1

tristanbob:
### Breaking change

Regarding the breaking change, we could leave the property and function names (unseen by users) with their original values (width/height) and simply change the labels and descriptions (length/thickness).  This would prevent the breaking change.

I don't see a huge value in making the names match, other than my OCD that wants them to match.  Since the names are similar in meaning and the descriptions/labels will show the better wording, I don't think we need to worry about future extension editors getting confused by this.

If you DO want to keep the breaking change, we should at least do what Arthuro recommends and increment the major version (1.x.x).

### Updates

Here is my most recent version (I apologize for breaking my local branch and not knowing how to recover it):

[SliderExtension_REVIEW3.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/7921399/SliderExtension_REVIEW3.zip)

- Added a check to make sure slider is Enabled before checking for hover or clicks  (NOT WORKING - Can you look at it?)
- Made "Enabled" properties and functions Public
- Changed halo size check to match recent changes to halo (in the last version, the slider would not stop redrawing because halo would never return to 0)

Bouh:
> If you DO want to keep the breaking change, we should at least do what Arthuro recommends and increment the major version (1.x.x).

Yes and adds a warning message in the description of the extension.

D8H:
Changes:
* Revert optional color properties as text as it's not possible to set an empty color (the UI default to black)
* Fix the "Enable" property
* Fix the animation asking a redraw every frame because of lerp never reaches its target and factorize it a bit.

https://www.dropbox.com/s/9o240reysrcy494/SliderExtension_REVIEW4.zip?dl=1

@tristanbob If you are ok with this version, we can add the warning and increment to 1.0.0

tristanbob:
> @tristanbob If you are ok with this version, we can add the warning and increment to 1.0.0

Davy - Thanks for your excellent help on this.  I found some regressions that I have added back.  After your next review, I am OK with a bump to 1.0.0.

## Version 0.7.7

[SliderExtension_REVIEW5.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/7930163/SliderExtension_REVIEW5.zip)

### Fix Regressions
- Added back the scene variable to prevent multiple sliders from being moved at once
- Added back the object/layer hidden check to prevent values from changing from unintended interaction with a hidden slider

### New logic
- Added logic when slider is disabled to stop dragging and set halo size to hide behind thumb
- Moved the "enabled" logic to top-level of PreEvents (to remove unnecessary duplication of the condition)

D8H:
### Changes

* changed the condition on the halo drawing to not draw it when it's smaller than the thumb.
* move the halo animation outside of the disable animation to let the shrink animation show (even if it will be very rare, for instance, when dragging and overlapping the check box; which is a bug of the checkbox)

> Added back the scene variable to prevent multiple sliders from being moved at once

Sorry, I didn't understand what it was for. I though it was a remain from when hidden properties didn't exist.
I rewrite it without a scene variable. It also fix a bug: a slider must be dragged only if the mouse was initially pressed inside the slider.

> Added back the object/layer hidden check to prevent values from changing from unintended interaction with a hidden slider

Actually, it's important that the slider works the same when it's hidden (apart from not rebuilding the shapes) because an extension user could want to use the slider extension hidden a put a Sprite for the track and the thumb to skin it.

https://www.dropbox.com/s/xh5h1rg0yjtb6a3/SliderExtension_REVIEW6.zip?dl=1

tristanbob:
> ### Changes of 0.7.8
> * changed the condition on the halo drawing to not draw it when it's smaller than the thumb.

Good idea, no need to waste CPU.  :)

> * move the halo animation outside of the disable animation to let the shrink animation show (even if it will be very rare, for instance, when dragging and overlapping the check box; which is a bug of the checkbox)

I'm not able to test this from here, but this only allows it to shrink, correct?  (We don't want the halo to grow while a slider is disabled)

> > Added back the scene variable to prevent multiple sliders from being moved at once
> 
> Sorry, I didn't understand what it was for. I though it was a remain from when hidden properties didn't exist. I rewrite it without a scene variable. It also fix a bug: a slider must be dragged only if the mouse was initially pressed inside the slider.

If we can make sure the initial click was inside the slider, then we don't need this scene variable.  My original problem was holding down the left mouse and dragging over multiple sliders; it was causing them all to activate until the mouse was released.  (I'll test this tonight)

> > Added back the object/layer hidden check to prevent values from changing from unintended interaction with a hidden slider
> 
> Actually, it's important that the slider works the same when it's hidden (apart from not rebuilding the shapes) because an extension user could want to use the slider extension hidden a put a Sprite for the track and the thumb to skin it.

Two thoughts on this:

1a) How will the user know the current location of the thumb?  Perhaps we need to publish some expressions to enable that?(ThumbCenterX, ThumbCenterY).  

1b) While on this topic, do you agree that the X, Y, and angle() values of the shape painter object can be used to find the starting point of the slider track?  How will they find the position of the end of the track?  

2) I am conflicted about this.  I'm worried that it won't be obvious to game creators that a hidden slider will still move if the user dragged over it.  However, I also like your idea of allowing sprites to be

D8H:
> I'm not able to test this from here, but this only allows it to shrink, correct? (We don't want the halo to grow while a slider is disabled)

Yes, that's what I was aiming for at least.

> 1a) How will the user know the current location of the thumb? Perhaps we need to publish some expressions to enable that?(ThumbCenterX, ThumbCenterY).

Yes, it will need more expressions, but I think we should keep it for another update.

> 1b) While on this topic, do you agree that the X, Y, and angle() values of the shape painter object can be used to find the starting point of the slider track? How will they find the position of the end of the track?

There are expression to convert coordinates back and forth. They can be used by the extension or the extension user.

>     2. I am conflicted about this.  I'm worried that it won't be obvious to game creators that a hidden slider will still move if the user dragged over it (they would need to disable the slider when they hide it).  However, I also like your idea of allowing sprites to be used on a hidden slider.  This could address Vegetato's idea for making sliders with images.

This is how behaviors usually work in GDevelop. Hiding a platformer character doesn't forbid it to move. It might be a bit less convenient but users can group them and disable them.

tristanbob:
## Version 0.7.9
- Fixed logic to decide when to draw halo (the halo was disappearing immediately and not animated)

## Kudos
@D8H the amount of finesse and polish that we have put into this extension is something to be proud of.  We have refined the logic, optimized the performance, and even added some juice.  This is a first-class extension, indeed!

I'm ready for 1.0.0 when you are.  :)

## Project files

[SliderExtension_REVIEW7.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/7938646/SliderExtension_REVIEW7.zip)

D8H:
I added the warning, you can squash and merge the branch if there isn't anything else to change.
https://www.dropbox.com/s/ajp8vmv3xdeft2x/SliderExtension_REVIEW8.zip?dl=1

D8H (approved):

4ian:
Incredible work everyone!! Love this 👍

PR #331, opened 2022-01-02
Updated Gamepad parameters to use a list of strings

tristanbob:
I predict that the gamepad extension is going to be one of the most used (and it will eventually be included by default), so I thought I would give it some love.

To do this, I changed almost all string parameters to use a list. 

The one exception was "C_Controller_type".

I think this function compares the given string parameter "controllerType" with the "gamepad.id" string provided by the gamepad device.  However, I don't understand how this is done using this line:

```
eventsFunctionContext.returnValue = gamepad ? gamepad.id.toUpperCase().indexOf(controllerType) !== -1 : false;
```

Can someone confirm that this function should not use a list of strings?

### DIFF
--- a/Extensions/Gamepads.json
+++ b/Extensions/Gamepads.json
@@ -1,6 +1,6 @@
 {
   "author": "Bouh",
-  "description": "Add support for gamepads (or other controllers) to your game, giving access to information such as button presses, axis positions, axis force, trigger pressure, deadzone for each gamepad, etc...\n\nUp to 4 gamepads can be connected: for each condition or expression, you'll have to enter the index of the gamepad to read. This is 1, 2,3 or 4.",
+  "description": "Add support for gamepads (or other controllers) to your game, giving access to information such as button presses, axis positions, axis force, trigger pressure, deadzone for each gamepad, etc...\n\nUp to 4 gamepads can be connected. For each condition or expression, you'll have to enter the index of the gamepad to read. This is 1, 2, 3 or 4.",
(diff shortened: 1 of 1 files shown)


4ian:
Thanks for taking care of this extension! It's indeed an extension that would almost benefit from being pushed forward in the interface or "featured" somehow because used a lot.

> eventsFunctionContext.returnValue = gamepad ? gamepad.id.toUpperCase().indexOf(controllerType) !== -1 : false;

This means:
- if there is a valid gamepad, then take its id, make it uppercase, and check if we find in the resulting string the "controller type" string passed as parameter.
  - if found, return true
  - if not found, return false
- if no valid gamepad, return false.

This means that if the gamepad id is "SUPER XBOX CONTROLLER", then indexOf("XBOX") will return 6 (because it's at position 6 in the string), so it's returning true. If I pass "PS5", it's not found, so indexOf returns -1.

I think we could add some default options but because the controller type can vary, it's not mandatory I would say!

4ian (approved):
I think it's good!

tristanbob:
> > eventsFunctionContext.returnValue = gamepad ? gamepad.id.toUpperCase().indexOf(controllerType) !== -1 : false;
> 
> This means that if the gamepad id is "SUPER XBOX CONTROLLER", then indexOf("XBOX") will return 6 (because it's at position 6 in the string), so it's returning true. If I pass "PS5", it's not found, so indexOf returns -1.

Thanks, this makes perfect sense now!  :)

PR #334, opened 2022-01-08
[PlatformerTrajectory] Add a behavior to evaluate jump trajectories.

D8H:
## Changes

Add a behavior (PlatformerEvaluator) that allows to
* Configure the height of a jump by automatically choosing the right jump speed.
* Know when to jump to reach a platform (can be useful for IA).
* Draw jump trajectories to check a level design.

I stated to write a documentation here:
https://github.com/4ian/GDevelop/tree/master/Extensions/PlatformBehavior#jump-trajectory

I used JavaScript because I think it might be useful to have some parts in GDevelop directly.

It needs GDevelop 125 to be more precise.

## Example

* Build: https://d8h.itch.io/platformer-jump-evaluator
* Project: https://www.dropbox.com/s/9mhv6vtxc0w3708/platformer-jump-evaluator.zip?dl=1

JumpSetup

### DIFF
--- a/Extensions/PlatformerTrajectory.json
+++ b/Extensions/PlatformerTrajectory.json
@@ -1,0 +1,1021 @@
+{
+  "author": "Entropy",
+  "description": "This extension allows to:\n* Configure the height of a jump by automatically choosing the right jump speed.\n* Know when to jump to reach a platform (can be useful for AI).\n* Draw jump trajectories to check a level design.",
+  "extensionNamespace": "",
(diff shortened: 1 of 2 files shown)


4ian:
I guess this needs a review that was never done, is it ready for one? :)

D8H:
Yes, it is ready. I never pushed it because it's a bit huge to review for the feature it gives.

4ian:
I think it’s good enough!

PR #336, opened 2022-01-13
Added condition "Is even"

tristanbob:
I think this condition makes sense in this extension, but if not please suggest another.  :)


arthuro555 (approved):
I think that makes sense too, don't forget to add yourself as an author ;)

tristanbob:
> I think that makes sense too, don't forget to add yourself as an author ;)

Thanks but I don't think I added enough on this change to be listed as an author.  

If I make significant improvements, I will add my name. :)

PR #338, opened 2022-01-15
Update ExtendedMath: Change ToFixed() to return a string instead of a number

tristanbob:
Change the ToFixed() function to return a string to match the javascript function that it emulates (and uses internally).
Previously, this function returned a number which meant that whole numbers would not display the right number of zeroes after the decimal point.

This update could break games that use this function AND manually update this extension.


D8H:
It seems good to me.

What do you thin about "Format" instead of "Convert"?

Extensions/ExtendedMath.json
```diff
@@ -1,19 +1,20 @@
...
   "name": "ExtendedMath",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/function-variant.svg",
   "shortDescription": "Additional math functions and constants as expressions and conditions.",
-  "version": "1.0.3",
+  "version": "1.0.5",
```
Bouh, line 11:
```suggestion
  "version": "1.1.0",
```

As there is a breaking change we would update to `1.1.0`

- First decimal is a major change in the conception + breaking change
- Second is a breaking change.
- Third is a simple fix that not affect the end user.

Extensions/ExtendedMath.json
```diff
@@ -1,19 +1,20 @@
...
   "tags": [
     "math",
     "constants",
     "conversion",
-    "coordinates"
+    "coordinates",
+    "convert"
```
Bouh, line 17: Can we add `modulo` for the condition `IsEven: Check if the number is even (divisible by 2)` ?

Extensions/ExtendedMath.json
```diff
@@ -1,19 +1,20 @@
 {
   "author": "add_, @Jurfix",
-  "description": "Additional math functions and constants expressions:\n - Mathematical constants\n - Greatest common factor/Lowest common multiple\n - Factorial\n - Coordinates conversion for polar and isometric systems\n - ToFixed is an expression for specify a limit of decimals places\n\nConditions:\n- Is even?",
+  "description": "Expressions:\n - Mathematical constants\n - Greatest common factor/Lowest common multiple\n - Factorial\n - Coordinates conversion for polar and isometric systems\n - ToFixed: Converts a number to a string with the specified number of decimal places\n\nConditions:\n- IsEven: Check if the number is even (divisible by 2)",
```
Bouh, line 3:
```suggestion
  "description": "Expressions:\n - Mathematical constants\n - Greatest common factor/Lowest common multiple\n - Factorial\n - Coordinates conversion for polar and isometric systems\n - ToFixed: Formats a number to a string with the specified number of decimal places\n\nConditions:\n- IsEven: Check if the number is even (divisible by 2)",
```

Extensions/ExtendedMath.json
```diff
@@ -763,9 +764,9 @@
       "objectGroups": []
     },
     {
-      "description": "Rounds a value to a specified number of decimal places.",
+      "description": "Converts a number to a string with the specified number of decimal places.",
```
Bouh, line 767:
```suggestion
      "description": "Formats a number to a string with the specified number of decimal places.",
```

Bouh (changes requested):

Extensions/ExtendedMath.json
arthuro555: That's not how semver works, see the spec: https://semver.org/
A breaking change always is a major bump. A minor bump is for backwards compatible feature addition.

arthuro555:
I'm unsure why this has to be a breaking change, not wanting to have a number exceed a certain amount of digits after the period is a perfectly valid use and what it must have been used for so far. Let's just make a separate function for formatting it as a string with the extra zeroes for those who want this other use case.

D8H:
> I'm unsure why this has to be a breaking change, not wanting to have a number exceed a certain amount of digits after the period is a perfectly valid use and what it must have been used for so far. Let's just make a separate function for formatting it as a string with the extra zeroes for those who want this other use case.

There is a reason why the JS method return a string:
* JS numbers have no explicit 0
* JS numbers are binary float not decimal float so it's not possible to round them in the decimal sense. `0.1` is not a rational binary number, printing it with the default format will give a lot of digits.

tristanbob:
> I'm unsure why this has to be a breaking change, not wanting to have a number exceed a certain amount of digits after the period is a perfectly valid use and what it must have been used for so far. Let's just make a separate function for formatting it as a string with the extra zeroes for those who want this other use case.

Doesn't the "trunc()" function already perform what you are describing?

arthuro555:
> There is a reason why the JS method return a string

No doubt that there's a reason why you would or should prefer a string there, though that's not the problem, the problem is that this PR is a breaking change. Therefore, I think that this should be made an extra function to not break people's games when updating. A breaking change must only be made of the tradeoffs are worth it, and I don't think that having some zeroes or specific numbers displayed, which is not even something we can be sure current users of this function may want or care about, is worth breaking all of their games if they decide to update, especially when GDevelop doesn't (yet) have a way to warn about breaking changes.

D8H:
> > There is a reason why the JS method return a string
> 
> No doubt that there's a reason why you would or should prefer a string there, though that's not the problem, the problem is that this PR is a breaking change. Therefore, I think that this should be made an extra function to not break people's games when updating. A breaking change must only be made of the tradeoffs are worth it, and I don't think that having some zeroes or specific numbers displayed, which is not even something we can be sure current users of this function may want or care about, is worth breaking all of their games if they decide to update, especially when GDevelop doesn't (yet) have a way to warn about breaking changes.

In another hand, this expression never did what it says it does and can not possibility do it.
Should we keep a bugged expression that seems to work at first glance but will sometimes do graphical bugs in games?

Extensions/ExtendedMath.json
tristanbob: I have just pushed a new version with 2.0.0 and a section with release notes about this change. 

Big picture, I think all extensions need a new "Release notes" field added, which can be shown before applying upgrades.

Extensions/ExtendedMath.json
tristanbob: I prefer the wording "convert".

Extensions/ExtendedMath.json
tristanbob: I'm never sure what should go in the tags.  Sometimes I feel like I add to much, like every keyword in the extension, and sometimes not enough.  I'm OK with adding modulo, although it is never exposed that way to the user.

tristanbob:
I added some release notes to the description and incremented a major version to honor SemVer.

```
Version 2.0.0
- "ToFixed()" now returns a string instead of a number.  This matches the javascript function and is helpful because whole numbers will now show the right number of zeroes.
- NOTE: This is a breaking change.  If you upgrade to this version, make sure to update every expression that uses "ToFixed()".
```

D8H:
I really think that "format" is a better choice than "convert".
* "convert" usually imply a operation that can be done both way. This is not the case here.
* "format" is widely used to describe a way to present a number with text.
  This not only a technical term, even Calc uses it. This "format" documentation has a "leading zeroes" section: https://help.libreoffice.org/latest/en-US/text/shared/01/05020300.html

tristanbob:
Since both of you agree, I will change the wording to "format". 

Just for fun:

Mozilla uses "format" 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

W3 uses "convert"
https://www.w3schools.com/jsref/jsref_tofixed.asp

arthuro555:
> this expression never did what it says it does and can not possibility do it.

I disagree, the expression says it rounds the number to up to a specific amount of digits. That's what it does, it never said it assured a fixed number of decimals. Sure, the name `ToFixed` might be confusing, and not correspond to JavaScript, but at no point does this extension or GDevelop assure a 1:1 mapping of the JavaScript API.

Either way, I think we should keep that function, bugged or not. More people will get issues by upgrading than because of graphical bugs due to one or two decimals that are 0 not being displayed.

The community keeps on complaining more and more about breaking changes. By allowing such changes because it doesn't fit your idea of what that function is supposed to do is not in my opinion for the best. Users use the function as it is and rely on it, and that change means suddenly breaking parts of their game, and while I see why you would want a different behavior for this expressions, this shouldn't be forced onto users if it may break their games.

Note also that keeping it doesn't mean we are recommending it in any way, we can make it private so that it doesn't show up anymore in autocomplete or the expression builder. It just means that we are not ruthlessly breaking players games.

D8H:
> Note also that keeping it doesn't mean we are recommending it in any way, we can make it private so that it doesn't show up anymore in autocomplete or the expression builder. It just means that we are not ruthlessly breaking players games.

Ok, what name should we choose? "FormatToFixed"?

We could add a warning in the description of the old expression like: "This expression is deprecated. Prefer using FormatToFixed to set values in text objects. This avoids to sometimes display a lot more digits than expected."

> I disagree, the expression says it rounds the number to up to a specific amount of digits. That's what it does, it never said it assured a fixed number of decimals. Sure, the name `ToFixed` might be confusing, and not correspond to JavaScript, but at no point does this extension or GDevelop assure a 1:1 mapping of the JavaScript API.

You can't decimally round a float. That is just not possible, many values doesn't exist.

D8H:
> > Ok, what name should we choose? "FormatToFixed"?
> 
> I reverted the breaking change and added the ToFixedString() function.

Ok

> > We could add a warning in the description of the old expression like: "This expression is deprecated. Prefer using FormatToFixed to set values in text objects. This avoids to sometimes display a lot more digits than expected."
> 
> I think you might be confused about the problem. Yes, the slider extension has the bug where the Value can have dozens of decimals for certain values (I wonder if we should fix that by inserting the ToFixed() javascript function into that extension...)
> 
> The "ToFixed" problem is that extra zeros at the end are not preserved, so ToFixed(5.000,3) becomes "5".

Indeed, JS numbers has enough precision to round display numbers with few digit correctly. So the fact that the returned number is not right doesn't show.
Yes, we can tell about the trailing zeroes. Something like: "This expression is deprecated. Prefer using ToFixedString to set values in text objects. This allows to keep the trailing zeroes."

> The ToFixed() function doesn't round, it truncates.

No, the difference between rounding and truncating is:
- round(1.127, 2) = 1.13
- truncate(1.127, 2) = 1.12
- round(1.124, 2) = 1.12
- truncate(1.124, 2) = 1.12

tristanbob:
I think I've made all the changes people wanted.  Does anyone want to change from yellow to green?

D8H:
> I think I've made all the changes people wanted. Does anyone want to change from yellow to green?

What about making "ToFixed" hidden and adding this: "This expression is deprecated. Prefer using ToFixedString to set values in text objects. This allows to keep the trailing zeroes." to its description?

D8H (changes requested):
> I think I've made all the changes people wanted. Does anyone want to change from yellow to green?

What about making "ToFixed" hidden and adding this: "This expression is deprecated. Prefer using ToFixedString to set values in text objects. This allows to keep the trailing zeroes." to its description?

tristanbob:
> What about making "ToFixed" hidden and adding this: "This expression is deprecated. Prefer using ToFixedString to set values in text objects. This allows to keep the trailing zeroes." to its description?

Great ideas!  I am submitting these changes soon...

Extensions/ExtendedMath.json
```diff
@@ -763,11 +788,12 @@
       "objectGroups": []
     },
     {
-      "description": "Rounds a value to a specified number of decimal places.",
+      "description": "Formats a number to use the specified number of decimal places. (Deprecated).",
```
D8H, line 791: There is a typo: ". ("?

tristanbob: Perhaps, I'm not sure what the right way is. I will remove that period.

arthuro555 (approved):

D8H (approved):

Extensions/ExtendedMath.json
D8H: There is already a modulo in the built-in expressions. This `IsEven` is probably useless, no need to advertise it.

D8H:
@Bouh I guess your change requests have been answered. Do you confirm?

Extensions/ExtendedMath.json
```diff
@@ -1,19 +1,22 @@
...
   "name": "ExtendedMath",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/function-variant.svg",
   "shortDescription": "Additional math functions and constants as expressions and conditions.",
-  "version": "1.0.3",
+  "version": "1.0.7",
```
D8H, line 11: We should go for `1.1.0` as there is a new function (no forward compatibility)

Bouh (approved):

Bouh:
> @Bouh I guess your change requests have been answered. Do you confirm?

I confirm :)
Merged, thank you everyone for this update! 👍

PR #343, opened 2022-01-20
[Slider] Handle object transformations.

D8H:
Changes
* Handle cursor events according to object rotation, scale and position
* Fix the initial value when it's not 0
* Use color property type
* Define properties groups
* Define functions groups
* Add getters and setters for valueMin, valueMax and tickSpacing properties
* Change the extension and behavior description
* Fix 2 typos

Test project: https://www.dropbox.com/s/bf8i3ototgwvjqn/ShapePainterTransformationTest.zip?dl=1

### DIFF
--- a/Extensions/DraggableSliderControl.json
+++ b/Extensions/DraggableSliderControl.json
@@ -1,18 +1,17 @@
 {
   "author": "Tristan Rhodes (https://victrisgames.itch.io/), D8H",
-  "description": "This extension provides a behavior that draws a horizontal slider that can be dragged by the users. The value range and the tick spacing can be defined with properties.\nNote that this behavior can only be attached to a **Shape Painter** object.\n\nFurther details can be found in [this tutorial video](https://youtu.be/iiTUwdAT_hs).\n",
+  "description": "This extension provides a behavior that draws a slider that can be dragged by the users. The value range and the tick spacing can be defined with properties.\nNote that this behavior can only be attached to a **Shape Painter** object.\n\nFurther details can be found in [this tutorial video](https://youtu.be/iiTUwdAT_hs).\n",
   "extensionNamespace": "",
-  "fullName": "Draggable Slider Control",
+  "fullName": "Draggable slider",
   "helpPath": "",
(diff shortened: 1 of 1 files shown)


tristanbob:
How do I submit my suggested changes of your version?

Do I "Add more commits by pushing to the TransformableSlider branch on D8H/GDevelop-extensions"?  Do I submit PR's on your forked branch there?

Or do I accept your version, and then base my branch on the merged version and submit a PR for my branch?

D8H:
> How do I submit my suggested changes of your version?
> 
> Do I "Add more commits by pushing to the TransformableSlider branch on D8H/GDevelop-extensions"? Do I submit PR's on your forked branch there?
> 
> Or do I accept your version, and then base my branch on the merged version and submit a PR for my branch?

It's better to merge mine when your are done with the review and rebase your PR over mine. Because, in theory, we shouldn't merge features we submitted. So, we can't both work on the same branch or we would have to find a 3rd person to review.

Bouh:
- The track thickness have a problem
image
- The draggable/focus area seem only on the track? If I click on the red point of the image below the slider isn't focus
image

D8H:
> * The track thickness have a problem
>     * The draggable/focus area seem only on the track? If I click on the red point of the image below the slider isn't focus
>     * The color of the track is pale/yellow-ish if not focus has happend atleast once

They're not regressions, it's how the extension from the repository works.

Bouh:
I know, I tested both versions to compare during my review.
I blame no one. I only made a review ☺

D8H:
> I know, I tested both versions to compare during my review. I blame no one. I only made a review ☺

I wasn't saying that it's not my fault. I was saying that it should not block a merge as this is not a new bug.

tristanbob:
> * The track thickness have a problem
>   image
> * The draggable/focus area seem only on the track? If I click on the red point of the image below the slider isn't focus
>   image
> * The color of the track is pale/yellow-ish if not focus has happend atleast once

I tried to follow Material Design for the slider UI:
https://mui.com/components/slider/

1) Halo should become visible when the mouse enters the radius of the halo.
2) The thumb should jump to the position of a click if the click is:
- Within the length of the track 
- AND 
- Within the width of the halo

I want to update the extension to follow this logic, but perhaps not inside Davy's PR.

Bouh:
Ok, lgtm 👍

PR #348, opened 2022-01-23
Various CI fixes

arthuro555:
- Make `!update` work when the file has been renamed
 - Don't generate a PR until the extension passes automatic tests
 - Add/Update issue output messages for increased clarity for the users of what's going on

Workflow deployed and tested on arthuro555/GDevelop-extensions, you can test it there.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,3 @@
 {
-    "typescript.tsdk": "node_modules/typescript/lib",
-    "yaml.schemas": {
-        "https://json.schemastore.org/github-workflow.json": "file:///c%3A/code/GDevelop-extensions/.github/workflows/build.yml"
-    }
+    "typescript.tsdk": "node_modules/typescript/lib"
 }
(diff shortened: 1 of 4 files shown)


.vscode/settings.json
```diff
@@ -1,3 +1,6 @@
 {
-    "typescript.tsdk": "node_modules/typescript/lib"
+    "typescript.tsdk": "node_modules/typescript/lib",
+    "yaml.schemas": {
```
4ian, line 3: Let's revert this maybe? You should be able to have this in your own settings?

4ian (approved):
👍

.vscode/settings.json
arthuro555: I've left this in here because I think everyone appreciates having autocomplete and validation, not just me :p

4ian: Sure but this is an absolute path on your computer ;)

arthuro555: oh 🤦 I'll remove it

PR #352, opened 2022-01-26
Update ScreenWrap.json

tristanbob:
## Version 0.2.0

- Added a new behavior for physics objects (Thanks Entropy!)
- Updated descriptions

## Project files
[ScreenWrapWithPhysics_Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/7938506/ScreenWrapWithPhysics_Example.zip)

## Video 

https://user-images.githubusercontent.com/8879811/151092622-8ac56153-3efc-44c1-9531-98c8cfd8165d.mp4


tristanbob:
Should we update this extension to have an option to use AABB?  If so, how should it be expose to the user?

- New string property called "TriggerSource", with options "Center of AABB bounding box (default)", "Center point of sprite", "Origin point of sprite" 

OR 

- Boolean property called "UseCenterPoint"

OR

- Some other method?

Bouh:
> Should we update this extension to have an option to use AABB? If so, how should it be expose to the user?

Based on the center is faster and makes more sense to me than the AABB.

tristanbob:
FYI - Sorry for the confusion.  This change is complete and ready for review.  My questions were about future improvements.

D8H:
* AngularVelocity should be hidden
* Why preEvents is not used in both?
* It needs a comment to explain why velocity need to be saved.
* The 2 behaviors are not small and there is a lot of duplicated code. This would probably be a nightmare to maintain.
  What about creating a new behavior to share all this code?
  * ManualScreenWrap will give:
    * 4 conditions to check if the object must be moved
	* 4 actions to move it on a screen border
	* it will handle all the configuration
  * ScreenWrap and ScreenWrap will require ManualScreenWrap and only contains a doStepPreEvent.

It's not an ideal solution and it's a breaking change but it should avoid headache.

D8H (changes requested):
* AngularVelocity should be hidden
* Why preEvents is not used in both?
* It needs a comment to explain why velocity need to be saved.
* The 2 behaviors are not small and there is a lot of duplicated code. This would probably be a nightmare to maintain.
  What about creating a new behavior to share all this code?
  * ManualScreenWrap will give:
    * 4 conditions to check if the object must be moved
	* 4 actions to move it on a screen border
	* it will handle all the configuration
  * ScreenWrap and ScreenWrap will require ManualScreenWrap and only contains a doStepPreEvent.

It's not an ideal solution and it's a breaking change but it should avoid headache.

tristanbob:
Should we:
- accept this PR knowing that works but has a lot of duplicated code
- people can start benefiting from using it today

Or:
- wait for someone to rewrite it to be more maintainable

I think we should accept this as-is, and we can always improve it later.

tristanbob:
I guess I could always resubmit it as a community extension?  Would that be the preferred option?

D8H:
Actually there is a lot of functions but not as much events as I though. I'm fine to keep it this way. Sorry for the trouble.

But this should still be addressed:
> * The AngularVelocity property should be hidden
> * It needs a comment to explain why velocity need to be saved.

Bouh:
@tristanbob There is only to add in 'doStepPreEvents' a comment for why the velocity needs to be saved.
If you know, you can tell me here, and I'll do the changes, then merge this PR.

D8H:
> I fixed the conflicts and moved this extension into the reviewed folder.
> 
> @tristanbob There is only to add in 'doStepPreEvents' (for physics behavior) a comment for why the velocity needs to be saved. If you know, you can tell me here, and I'll do the changes, then merge this PR.

It's because Physics2 reset object velocities when an object is moved from the outside of the extension.

Bouh:
Thanks for the detail.

I added the comment and tested the extension.

It can be merged.

D8H (approved):

PR #353, opened 2022-01-26
Update DragCameraWithPointer.json

tristanbob:
- Allow user to specify mouse button to use for dragging the camera
- Cleaned up logic to use fewer events
- Added prefix to variables to prevent collision with user variables (best practice)
- Updated descriptions


D8H:
Does this extension has scrolling inertia?

tristanbob:
> Does this extension has scrolling inertia?

No, it does not.  Why do you ask?  I'm trying to think of any games that use something like that.

D8H:
> > Does this extension has scrolling inertia?
> 
> No, it does not. Why do you ask? I'm trying to think of any games that use something like that.

No reason, just an idea.

D8H:
Is there any example to check it quickly?

VegeTato:
> Is there any example to check it quickly?

i had the same question, also please the extension file so i can download and test it out :3

tristanbob:
> Is there any example to check it quickly?

Sorry, I usually do include this, but this extension is very simple (and very useful).

To test:
1) Open any existing project that doesn't move the camera (or disable the camera events)
2) Install this extension (see below) 
3) Add the action "Drag camera with mouse"
4) Preview game and drag the mouse

[DragCameraWithPointer.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/7954876/DragCameraWithPointer.zip)

D8H:
* The camera actions are deprecated. They should be replaced by the new ones.
* Why not ">= 1" instead of "> 1"?
* Games that update the extensions will no longer be able to drag until the button parameter is set.
  What was the default button?
  What about a condition like this (to avoid a breaking change)?
  image

D8H (changes requested):
It works with a comparaison to an empty string but the action on my screenshot is more resilient to change in the engine because it can handle `null` and `undefinied` if it's necessary one day (its sentense is not great, but take a look at the description).

I think the sentense can be shrinked a little.
* from: "Drag camera on layer _PARAM2_ using _PARAM4_ mouse button and allow movement in _PARAM3_ directions"
* to "Drag camera on layer _PARAM2_ in _PARAM3_ directions using _PARAM4_ mouse button"

tristanbob:
@D8H great insights.   I have shortened the description using your wording (much clearer now) and used your condition to find out if the user has specified an input button or not.

D8H (approved):

Extensions/DragCameraWithPointer.json
```diff
@@ -1,14 +1,18 @@
...
   "name": "DragCameraWithPointer",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/drag-variant.svg",
-  "shortDescription": "An action to move the camera by dragging with the mouse (or scrolling with the touch), on vertical, horizontal or both directions.",
-  "version": "1.0.1",
+  "shortDescription": "Move a camera by dragging the mouse (or touchscreen).",
+  "version": "1.2.3",
```
D8H, line 11: The minor should be incremented only once.

tristanbob: Roger that.  Updated.

D8H: ok

4ian:
I went ahead and merged this, thanks everyone!

PR #360, opened 2022-02-06
[MarchingSquares] Collision optimization

D8H:
Also add:
* upside-down flood action
* group definitions
* fix the wiki link

# Example
## QIX
* Build: https://liluo.io/games/645ed2ca-2940-47e6-bd0a-df726cfa05c1
* Project: https://www.dropbox.com/s/121lrua9tws08l4/marching-squares-qix.zip?dl=1

## Platformer
* Build: https://liluo.io/games/91e4a59a-7743-4e2c-b1c3-dac7e157df30
* Project: https://www.dropbox.com/s/lbybhiii0uksrvy/marching-squares-editor.zip?dl=1

MarchingSquareLevel

# Optimization

The scenario is not favorable to the optimization:
* Small cells of 10 pixels
* A medium size map of 1600x1200
* a character with a sprite a bit bigger than its hitbox
* the character is against a wall

The character AABB is overlapping about 10 hitboxes (it could 3 or 4 in more common scenario). There is still an overall 25% CPU time save.

## Before

CollisionOptimizationCpuBefore

CollisionOptimizationBottomUpBefore

## After

CollisionOptimizationCpuAfter

CollisionOptimizationBottomUpAfter

# JavaScript main changes

This is to only check collision with the platform hitboxes next to the other object.

## Iterable over hitboxes of a given area

```JavaScript
  if (!gdjs.__marchingSquares) {
    // Create a namespace
    gdjs.__marchingSquares = {};
    /**
     * The hitboxes in a rectangular area.
     * @constructor
           * @param {any} behavior
           * @param {integer} xMin The fist column to include.
           * @param {integer} yMin The fist row to include.
           * @param {integer} xMax The last column to include.
           * @param {integer} yMax The last row to include.
     */
    gdjs.__marchingSquares.HitboxesIterable = function HitboxesIterable(
      behavior,
      xMin,
      yMin,
      xMax,
      yMax
    ) {
      this.behavior = behavior;
      this.xMin = xMin;
      this.yMin = yMin;
      this.xMax = xMax;
      this.yMax = yMax;
    };

    gdjs.__marchingSquares.HitboxesIterable.prototype.setBounds = function (
      xMin,
      yMin,
      xMax,
      yMax
    ) {
      this.xMin = xMin;
      this.yMin = yMin;
      this.xMax = xMax;
      this.yMax = yMax;
    }

    gdjs.__marchingSquares.HitboxesIterable.prototype[Symbol.iterator] = function () {
      // xMin and yMin next increment
      let x = this.xMax;
      let y = this.yMin - 1;
      let hitbox = null;
      let previousHitbox = null;
      /** @type {IteratorReturnResult<gdjs.Polygon>} */
      const result = {done: false,value: undefined};

      return {
        next: () => {
          do {
            x++;
            if (x > this.xMax) {
              y++;
              x = this.xMin;
            }
            if (y > this.yMax) {
              // done
              result.done = true;
              result.value = undefined;
              return result;
            }
            hitbox = this.behavior.getHitbox(x, y);
          }
          while (!hitbox || hitbox === previousHitbox);

          // happen with run-length encoding
          previousHitbox = hitbox;
          result.done = false;
          result.value = hitbox;
          return result;
        },
      };
    };
  }
```

## Overriding of object methods

```JavaScript

  /**
   * @type {Iterable<gdjs.Polygon>}
   */
  behavior.marchingSquaresHitBoxesIterable = new gdjs.__marchingSquares.HitboxesIterable(behavior, 0, 0, 0, 0);

  behavior.owner.updateHitBoxes = function () {
    this.hitBoxesDirty = false;
  }

  behavior.owner.getHitBoxesAround = function (
    left,
    top,
    right,
    bottom
  ) {
    const leftIndex = Math.max(0, Math.floor(behavior.convertToGridBasisX(left)));
    const topIndex = Math.max(0, Math.floor(behavior.convertToGridBasisY(top)));
    const rightIndex = Math.min(behavior.dimX() - 1, Math.ceil(behavior.convertToGridBasisX(right)));
    const bottomIndex = Math.min(behavior.dimY() - 1, Math.ceil(behavior.convertToGridBasisY(bottom)));
    /** @type {Iterable<gdjs.Polygon>} */
    const iterable = behavior.marchingSquaresHitBoxesIterable;
    iterable.setBounds(leftIndex, topIndex, rightIndex, bottomIndex);
    return iterable;
  }
```


4ian:
Lovely! I'm still a bit concerned with the usage of Iterable. I wonder if we could not get a plain old array to be returned?
What I don't like with Iterable is that I'm not sure how performant they are for a real time usage (how much garbage they create in memory for the GC? what is the overhead of next/calling their functions compared to accessing an index in an array, which is the simplest operation in the world).

D8H:
The big issue with arrays as return type is that they don't allow efficient parallelism. The class wouldn't be able to reuse the returned array in this context. I think we shouldn't close the door to parallelism for collision calculus. 

The returned instance can be reused to avoid too much allocation. We could even reuse the iterator which is easy to revert if we need parallelism one day.

4ian:
Sorry I'm a bit out of the loop here. Thanks for the optimization!
Still a few questions:

> The big issue with arrays as return type is that they don't allow efficient parallelism. The class wouldn't be able to reuse the returned array in this context. I think we shouldn't close the door to parallelism for collision calculus.

Agree with not closing the door to parallelism, but can you explain how the iterator would allow it?
I was thinking that an array would make it easier because you can launch multiple jobs on different parts of the array, but I guess you're thinking about another parallelism approach?

D8H:
> Sorry I'm a bit out of the loop here. Thanks for the optimization! Still a few questions:
> 
> > The big issue with arrays as return type is that they don't allow efficient parallelism. The class wouldn't be able to reuse the returned array in this context. I think we shouldn't close the door to parallelism for collision calculus.
> 
> Agree with not closing the door to parallelism, but can you explain how the iterator would allow it? I was thinking that an array would make it easier because you can launch multiple jobs on different parts of the array, but I guess you're thinking about another parallelism approach?

Let's say we want to check the collision between a grid collision mask object and N objects and each check can run in parallel.

A collision check will be something like this:
- find the subset of polygons from the grid collision mask around the other object
- loop them against the other object polygons

The subset of polygons given by the grid collision mask will be different for the N objects because it depends on their AABB. As, the polygons iteration might run in parallel, the collision checks need their own instance of iterator. This is not a problem with an iterator because its memory footprint is light, but with an array an allocation could make the check longer to process than without any parallelism.

D8H:
I added a build of the QIX-like example that uses collisions too.

4ian:
Finally merged this! :)

PR #362, opened 2022-02-06
Update Extended variables extension

Bouh:
- This fix function name to respect the PascalCase for function name.
(So a lot of breaking change in functions)
- Add functions to copy a variables to another one, type and value(s) are copied.

### DIFF
--- a/Extensions/ExtendedVariables.json
+++ b/Extensions/ExtendedVariables.json
@@ -1,6 +1,6 @@
 {
   "author": "@Bouh, D8H",
-  "category": "",
+  "category": "General",
   "description": "This extension provides: \n- Conditions to check if a global/scene/object variable exists.\n- Actions to copy object variables, global and scene variables.\n- Actions to delete an existing global/scene variable from memory.\n- Actions to create a global/scene variable using a text expression, rather than a variable name input.\n- Expressions to get the value or text from a global/scene variable using a text expression, rather than a variable name.\n\nText expressions allows to enter the name of the variable with an expression like `\"MyVar\"+ ToString(Variable(Health))`",
   "extensionNamespace": "",
   "fullName": "Extended variables support",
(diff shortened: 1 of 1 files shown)


Extensions/ExtendedVariables.json
```diff
@@ -308,7 +324,8 @@
       "description": "Modify the value of a scene variable.",
       "fullName": "Value of a scene variable",
       "functionType": "Action",
-      "name": "A_Modify_scene_variable",
+      "group": "Scene variables",
+      "name": "ModifySceneVariable",
       "private": false,
       "sentence": "Change the scene variable _PARAM1_ with value: _PARAM2_",
```
D8H, line 330: What about: `Change the text of global variable _PARAM1_: _PARAM2_` to use the same sentence as the built-in action.

Extensions/ExtendedVariables.json
```diff
@@ -28,7 +37,8 @@
       "description": "Check if a global variable exists.",
       "fullName": "Global variable exists",
       "functionType": "Condition",
-      "name": "C_global_variable_exist",
+      "group": "Global variables",
+      "name": "GlobalVariableExist",
```
D8H, line 41: Could we keep the old function make them private and call the new functions?

Bouh: I can add a warning message in the description to warn people.
Old functions will be shown as `Unknown`, so people will be forced to fix them.
I don't wanna duplicate all functions, the idea is to start again on a solid base.

D8H: Without having a message that gives the name of the old function in the error message (like you suggested the other day), users won't know what was the function they used and more importantly its parameter expressions.

Old functions can just be copy-pasted and will be one liners. The private icon and strange name will make them easy to identify for extension maintainers. I think it's not a huge effort and it can save a lot of time to extension users.

D8H:
* In "Copy..." functions, instead of:
  * "Erase existing values in destination variable"
  * "(exact copy: _PARAM3_)"
  * What about "Clear destination first" to refer to the action "Clear variable" that users might understand better.
* There is an empty event in "Copy..." functions.

D8H (approved):

PR #364, opened 2022-02-08
Remove console.error when no buttons are pressed on Gamepad

ClementPasteau:
Gamepads extension is generating error if no button is pressed when using the condition `any button pressed`.
I don't think it should, as it's just a normal behavior

image


4ian:
Thank you! More than the console.error, is the logic of the condition good (or is it fixed now)?

4ian (approved):

ClementPasteau:
Yep, probably a mistake on my end while testing 👍

PR #366, opened 2022-02-09
[Gamepad] Update - Add a function to check Xbox controllers

Bouh:
Check for "Xbox" and "Xinput" for Xbox Microsoft Controllers to fix a possible issue on Firefox that only detect "Xinput".

Function condition: **C_Controller_type** changed
And the function `isXbox()` used in the condition is in: **FirstSceneLoaded**

cc @Silver-Streak
cc @ClementPasteau


Extensions/Gamepads.json
```diff
@@ -9,7 +9,7 @@
   "name": "Gamepads",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/gamepad-variant-outline.svg",
   "shortDescription": "Add support for gamepads (or other controllers) to your game, giving access to information such as button presses, axis positions, trigger pressure, etc...",
-  "version": "0.0.10",
+  "version": "0.0.2",
```
4ian, line 12: Should be increasing, either 0.2.0 or 0.0.11 :)

Bouh, line 12:
```suggestion
  "version": "0.2.0",
```

Bouh:
[This could help?](https://gist.github.com/Enichan/de1618d2c481523bd0c1123800bfe747)

ClementPasteau:
That's a really interesting link, however it shows that it's almost impossible to know which gamepad is connected because of the browser 😓  (Opera and Safari are even worse lol)

Bouh:
Ok then as we can't trust Xinput, I comment on the line below, and I merge this PR.
image

4ian:
Ok! :)

PR #367, opened 2022-02-13
Add the SmoothCamera extension to follow an object.

D8H:
The camera follows an object according to:
- a frame rate independent speed to make the scrolling from smooth to strong
- a follow-free zone to avoid scrolling on small movements
- an offset to see further in one direction
- position forecasting and delay to simulate a cameraman response time
- an extra delay and catch-up speed to give an impression of speed (like a spin dash)
- a platformer dedicated behavior to stabilize the camera for jumps

## Demo
4 camera settings (press Return) and a catch up like the spin dash (the object need to be fixed before the dash)
- Stabilized jump exponential following
- Exponential following
- Linear following (Sonic like)
- Shoulder camera (not a great idea for platformers)

* Project: https://www.dropbox.com/s/08d29yjf5kf3817/AdvancedJumpAirAndWallJumpExample_D8H-1.zip?dl=1
* Build: https://liluo.io/instant-builds/6430a99f-51b0-4b0e-a77b-d2ea76ee02de

Low resolution examples:
* Exponential following: https://liluo.io/games/0a969e3e-21be-4c38-99d5-0c5a0f2ec5ad
* Linear following: https://liluo.io/games/22587fe8-6afd-4f0a-8140-969f480a3677

### DIFF
--- a/Extensions/SmoothCamera.json
+++ b/Extensions/SmoothCamera.json
@@ -5924,7 +5924,7 @@
       ]
     },
     {
-      "description": "Smoothly scroll to follow a character and stabilize the camera when jumping",
+      "description": "Smoothly scroll to follow a character and stabilize the camera when jumping.",
       "fullName": "Smooth platformer camera",
       "name": "SmoothPlatformerCamera",
       "objectType": "",
--- a/Extensions/SmoothCamera.json
+++ b/Extensions/SmoothCamera.json
@@ -731,7 +731,7 @@
           "objectGroups": []
         },
         {
-          "description": "Delay the camera and catch up the delay at a given speed. This action must be used when the object doesn't move yet.",
+          "description": "Delay the camera according to a maximum speed and catch up the delay.",
           "fullName": "Wait and catch up",
           "functionType": "Action",
           "group": "",
--- a/Extensions/SmoothCamera.json
+++ b/Extensions/SmoothCamera.json
@@ -6219,7 +6219,7 @@
           "name": "FloorFollowFreeAreaBottom"
         },
         {
-          "value": "0,95",
+          "value": "0.95",
           "type": "Number",
           "label": "Upward speed in the air (in ratio persecond)",
           "description": "",
@@ -6229,7 +6229,7 @@
           "name": "AirUpwardSpeed"
         },
         {
-          "value": "0,95",
+          "value": "0.95",
           "type": "Number",
           "label": "Downward speed in the air (in ratio persecond)",
           "description": "",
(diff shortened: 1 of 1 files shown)


4ian:
Very nice! I've not made an extensive review but happy to have this extension in. Can you elaborate on how the frame rate independent speed works?

Bouh:
I used the project example to review.

- In the engine until now units are in the parameters.
image
- I wonder what the "Layer name" input do:
I didn't know what should be the value, the layer of the object I guess? but what is the interest to change it.
I think the "Layer name" should be optionnal, the default value is an empty string by default and as the behavior is already on the object we could event use the layer name of the object.

Without trying the parameter, we don't know what its purpose is.
I think we can be more explicit about how/why to use a parameter by explaining it in the long description.

- By habit we return an error when the parameters are empty in the event sheet, even if the engine will handle it correctly for you by adding in the code generation an empty string. That's why I recommend to set double quotes as default value.
image

- Console log is used in function `WaitAndCatchUp`, it's nice to add logs, could you add a condition to check if the debug action is enabled? We don't have to log everytime I think.

- A disabled event in `DrawDebug`, we can clean it if not used.
- Unknown condition in `IsDelayed`: 

```
{"000kind":"GDEVELOP_EventsAndInstructions_CLIPBOARD_KIND-jsBdHbLy912y8Rc","content":{"eventsList":[],"eventsCount":0,"actionsList":[],"actionsCount":0,"conditionsList":[{"type":{"inverted":false,"value":"Egal"},"parameters":["Object.Behavior::CurrentDelay()",">","0"],"subInstructions":[]}],"conditionsCount":1}}```

Same in `UpdateForecastedPosition`

![image](https://user-images.githubusercontent.com/1670670/153852960-94db55a9-40a5-41c2-b771-068a3829f68c.png)

D8H:
> Very nice! I've not made an extensive review but happy to have this extension in. Can you elaborate on how the frame rate independent speed works?

I added this comment:
> At each frame, the camera must catchup the target by a given ratio (speed)
> cameraX(t) - targetX = (cameraX(t - 1) - targetX) * speed
> 
> The frame rate must not impact on the catch-up speed, we don't want a speed in ratio per frame but a speed ratio per second, like this:
> cameraX(t) - targetX = (cameraX(t - 1s) - targetX) * speed
> 
> Ok, but we still need to process each frame, we can use a exponent for this:
> cameraX(t) - targetX = (cameraX(t - timeDelta) - targetX) * speed^timeDelta
> cameraX(t) = targetX + (cameraX(t - timeDelta) - targetX) * exp(timeDelta * ln(speed))
> 
> pow is probably more efficient than precalculated log if the speed is changed continuously but this might be rare enough.

I think I've addressed Bouh comments. I guess, the extension is ready for another review.

What do you think of the forecasting feature?
It probably won't be used much and it's a big part of the code. I'm willing to remove it if it's too specific to be in an extension or if it can speed up the review.

4ian (approved):
Looks good to me to go ahead with this :)
Thanks @bouh for the review too!

PR #379, opened 2022-03-11
Added Draggable Physics

tristanbob:
Enables users to click and drag on objects to move them.  The object will retain velocity when the click is released, which can allow users to fling objects across the screen.

This extension will only work on objects that have the physics behavior.

This extension is similar to the standard "draggable" behavior, but that behavior will not work on objects using the physics behavior.

## Example project

~~[MouseJointExtension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8312164/MouseJointExtension.zip)~~

[MouseJointExtension-0.2.2.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8381381/MouseJointExtension-0.2.2.zip)

## Video

~~https://user-images.githubusercontent.com/8879811/157907604-4ee3bca8-9634-47e0-aa7b-40a05119d424.mp4~~

https://user-images.githubusercontent.com/8879811/160860278-c586b1b5-3c25-42c2-98aa-d1a42c6d104a.mp4

## Playable game

https://games.gdevelop-app.com/game-bc0967c8-9e93-4ae6-813d-1c741f7515f5/index.html


4ian:
Great idea! And great usage of "required behaviors" 👍👍

Can you explain what's the need for the "post events" logic? :)

tristanbob:
> Can you explain what's the need for the "post events" logic? :)

To be honest, I was just following @D8H example with how they split the Draggable Slider into:

Pre-events = User interaction
Post-events = Drawing slider

I don't know which life-cycle is best for this extension.  Do you have a recommendation?

D8H:
> > Can you explain what's the need for the "post events" logic? :)
> 
> To be honest, I was just following @D8H example with how they split the Draggable Slider into:
> 
> Pre-events = User interaction
> Post-events = Drawing slider
> 
> I don't know which life-cycle is best for this extension. Do you have a recommendation?

I guess that this extension doesn't do any rending related actions so everything can be in preEvent.

4ian:
As much as possible, favor the pre events lifecycle to give a chance to the user of the behavior to add events that would react to the result of the behavior before the scene is rendered on screen (here it's not that important, but can be upon the case of behaviors that are for example changing the position of an object: you want to give a chance to the scene events to adjust other object positions).

Apart from that, looks simple and useful to quickly make an object draggable, that's cool 👍

tristanbob:
Any other comments on this?  If not, does anyone want to approve it?

D8H:
The example seems to be from 2022-03-11. Can you update it?
(It's easier to just download a project rather than checkouting a branch and importing the extension in the project)

tristanbob:
> The example seems to be from 2022-03-11. Can you update it? (It's easier to just download a project rather than checkouting a branch and importing the extension in the project)

Sure!  I just updated the original link.

D8H:
In "Word of Goo", the dragged object is released if the mouse is too far away. Can you add this in you example? I guess, a "Is being dragged" condition (like in the DraggableObject extension) and an "Abort dragging" action will be needed.

Several objects can be dragged at the same time. I don't know if this is a bug of a feature.

image

Did you try it on a multi-touch device?

tristanbob:
> In "Word of Goo", the dragged object is released if the mouse is too far away. Can you add this in your example to verify that it is possible to do? I guess, a "Is being dragged" condition (like in the DraggableObject extension) and an "Abort dragging" action will be needed.

Great ideas!  Added, see below.  
**Note:**  The official draggable behavior should also have an action to cancel the drag.

> Several objects can be dragged at the same time. I don't know if this is a bug of a feature.

Not intended, I added a TriggerOnce to fix this.

> Did you try it on a multi-touch device?

I just added Multitouch version, but it is not working as expected.  Can you review what I tried?  (it is disabled in the extension)

## Version 0.2.0

- Added "Is being dragged" condition
- Added "Release dragged object" action
- Attempted multitouch functionality, but not working.  I am requesting help with this.

Example project above has been updated.

D8H:
I've never used touch events so I might misunderstand, but from what I can see it's not designed to be used within extensions
* It should be done only if "De/activate moving the mouse cursor with touches" is set to true but there is no conditions for it.
* According to the touch condition description, the touch event disappear when the condition is used by the 1st instance. Saving the touch identifier could work but the extension users still won't be able to handle touch events.

The multi-touch functions of the engine probably needs a rework.

The extension is still useful without multi-touch, it can be left aside for now.


> Note: The official draggable behavior should also have an action to cancel the drag.

Yes, I think someone asked it and the workaround is to deactivate and activate back the behavior if I remember correctly, but an "abort" action is clearer. This extension should probably handle behavior deactivation too.

tristanbob:
> The multi-touch functions of the engine probably needs a rework.

Yes, I agree. I stopped building a few multi-touch extensions due to this limitation.

> Yes, I think someone asked it and the workaround is to deactivate and activate back the behavior if I remember correctly, but an "abort" action is clearer. This extension should probably handle behavior deactivation too.

I added a "Behavior deactivated" event.

Extensions/DraggablePhysics.json
```diff
@@ -0,0 +1,554 @@
+{
+  "author": "",
+  "category": "",
+  "description": "Enables users to click and drag on objects to move them.  The object will retain velocity when the click is released, which can allow users to fling objects across the screen.\n\nNote:\n- The standard \"draggable\" behavior will not work on objects that have the physics behavior. This extension should be used instead.\n- This extension will only work on objects that have the physics behavior.",
+  "extensionNamespace": "",
+  "fullName": "Draggable (for physics objects)",
+  "helpPath": "",
+  "iconUrl": "",
```
Bouh, line 8: You didn't choose an icon for this new extension?
I see the Laser-pointer icon, feel free to choose one.

Bouh (approved):

Extensions/DraggablePhysics.json
tristanbob: Good catch! I picked one and pushed an update.  

Thanks for the review, Bouh!

D8H:
* `preEvent` could use "is being dragged" (a 2nd time) and "release drag"

* It's still possible to drag 2 objects at once when they are next to each other. A scene variable might be needed to fix this.

* I'm not sure if it's necessary to add "(physics)" at the end of sentences as the icon already allow to differentiate the 2 behaviors.

tristanbob:
> * `preEvent` could use "is being dragged" (a 2nd time) and "release dragged object"

Good call, added.

> * It's still possible to drag 2 objects at once when they are next to each other. A scene variable might be needed to fix this.

How does the built-in Draggable handle this?  I kinda think we should choose the Highest Z value, but that condition is not available in GDevelop yet.

> * I'm not sure if it's necessary to add "(physics)" at the end of sentences as the icon already allow to differentiate the 2 behaviors.

I do think it is important to keep the text to reduce confusion.

D8H:
> How does the built-in Draggable handle this?
> 
> I kinda think we should choose the Highest Z value, but that condition is not available in GDevelop yet. I tried using "pick a random object" to select a single object, but I forgot that behavior functions act similar to a ForEach() loop so this logic would have no effect.

It know every object that has the behavior which is not possible to do with events. But with solids there should not be superposition so z-order doesn't really mater. Not even distance mater as this only happen at the object bounds. The 1st evaluated object can be used, but this might be impossible to do when it handles multitouch. So, I guess it's fin not to fix it.

> > * I'm not sure if it's necessary to add "(physics)" at the end of sentences as the icon already allow to differentiate the 2 behaviors.
> 
> I do think it is important to keep the text to reduce confusion.

ok

Bouh:
If I read, we've done with this extension.
If we're ready for merging it's a go :)

D8H:
There is a small simplification of event using the behavior functions to do, I guess.

tristanbob:
> There is a small simplification of event using the behavior functions to do, I guess.

I pushed a commit with the changes we discussed above.

D8H:
> > There is a small simplification of event using the behavior functions to do, I guess.
> 
> I pushed a commit with the changes we discussed above.

I was thinking of something like this:

image

Also, for review iterations, only the revision should be incremented. It avoids to have version with higher value that what will be merged. It's a bit odd to release a 1st version with 0.4.0 like the ObjectSpawner. It should be either 1.0.0 or 0.1.0

tristanbob:
> I was thinking of something like this:

Good idea, added.

> Also, for review iterations, only the revision should be incremented. It avoids to have version with higher value that what will be merged. It's a bit odd to release a 1st version with 0.4.0 like the ObjectSpawner. It should be either 1.0.0 or 0.1.0

I think I understand; we don't need to increment revision numbers during a review, since Github is handling versioning.  I'll change it to 1.0.0 on my next commit.

tristanbob:
@D8H let me know if you think this is ready for approval.  

I have a quick tutorial video created that is scheduled to be published on Sunday morning.  :)

D8H:
The "release" action is still not used.

tristanbob:
> * The "release" action is still not used in preEvent.

I see where I missed that.  Fixed.

> * The condition in the "release" action doesn't seem useful, same for the one under "mouse released".

I was trying to optimize performance, but perhaps I will only leave the one inside "ReleaseDrag".  Changed.

> * "Physics behavior" name is "Property".

Changed to "PhysicsBehavior"

> * There is already a default value for properties, it should not be in the description. Make a change request to set back default value if you think it's important.

Text showing default values is removed.

> * I know that it's the Physics descriptions but telling to choose a positive number for the force, frequency and damping ratio is not necessary. I doubt people will be stuck trying only negative values. It would make more sense to state that the ratio is between 0 and 1.

Removed instructions but added "(Range: 0 to 1)" to Damping.

> * I prefer to leave hidden properties without description because it makes them easier to find.

Interesting view.  I am fine removing it in this case because the name is self-explanatory.  However, if the property is more complex I think it is useful to describe it, even if it is hidden.

Side note:  We need the ability to set a default selection when using "String from a list of options (text)".

D8H:
> > * The condition in the "release" action doesn't seem useful, same for the one under "mouse released".
> 
> I was trying to optimize performance, but perhaps I will only leave the one inside "ReleaseDrag". Changed.

The Physics action already checks that there is nothing to remove, so adding a double check won't make it faster. Removing the condition make the code faster to read, but it's not very important. It can stay this way.

D8H (approved):

PR #380, opened 2022-03-11
Added ObjectSpawner

tristanbob:
Spawn (create) objects over time

How to use:
- Add this behavior to an object
- Run the "Spawn objects" action on every frame

Tips:
- Use "Object was just spawned" condition to trigger visual and sound effects (if desired)

## Example project

[ObjectSpawnerExtension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8234243/ObjectSpawnerExtension.zip)

## Video

https://user-images.githubusercontent.com/8879811/157922877-69a8ac5b-1ec0-4700-be8b-cad70b0e5d8b.mp4

### DIFF
--- a/Extensions/ObjectSpawner.json
+++ b/Extensions/ObjectSpawner.json
@@ -1,0 +1,942 @@
+{
+  "author": "",
+  "description": "Spawn (create) objects over time\n\nHow to use:\n- Add this behavior to an object\n- Run the \"Spawn objects\" action on every frame\n\nTips:\n- Use \"Object was just spawned\" condition to trigger visual and sound effects (if desired)",
+  "extensionNamespace": "",
+  "fullName": "Object spawner",
+  "helpPath": "",
+  "iconUrl": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz48IURPQ1RZUEUgc3ZnIFBVQkxJQyAiLS8vVzNDLy9EVEQgU1ZHIDEuMS8vRU4iICJodHRwOi8vd3d3LnczLm9yZy9HcmFwaGljcy9TVkcvMS4xL0RURC9zdmcxMS5kdGQiPjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0ibWRpLXBsdXMtb25lIiB3aWR0aD0iMjQiIGhlaWdodD0iMjQiIHZpZXdCb3g9IjAgMCAyNCAyNCI+PHBhdGggZD0iTTEwLDhWMTJIMTRWMTRIMTBWMThIOFYxNEg0VjEySDhWOEgxME0xNC41LDYuMDhMMTksNVYxOEgxN1Y3LjRMMTQuNSw3LjlWNi4wOFoiIC8+PC9zdmc+",
+  "name": "ObjectSpawner",
+  "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/plus-one.svg",
+  "shortDescription": "Spawn (create) objects over time",
+  "version": "0.1.0",
+  "tags": [
+    "spawn",
+    "spawner",
+    "objects",
+    "create",
+    "timer"
+  ],
+  "authorIds": [
+    "gqDaZjCfevOOxBYkK6zlhtZnXCg1"
+  ],
+  "dependencies": [],
+  "eventsFunctions": [],
+  "eventsBasedBehaviors": [
+    {
+      "description": "Spawn (create) objects over time",
+      "fullName": "Spawn (create) objects over time",
+      "name": "ObjectSpawner",
+      "objectType": "",
+      "eventsFunctions": [
+        {
+          "description": "Periodically spawn (create) an object.  ",
+          "fullName": "Periodically spawn (create) an object",
+          "functionType": "Action",
+          "group": "",
+          "name": "SpawnObject",
+          "private": false,
+          "sentence": "Periodically create a _PARAM2_ located at _PARAM0_ while there are fewer than _PARAM3_ instances in the scene.",
+          "events": [
+            {
+              "disabled": false,
+              "folded": false,
+              "type": "BuiltinCommonInstructions::Comment",
+              "color": {
+                "b": 109,
+                "g": 230,
+                "r": 255,
+                "textB": 0,
+                "textG": 0,
+                "textR": 0
+              },
+              "comment": "Reset spawn detection",
+              "comment2": ""
+            },
+            {
+              "disabled": false,
+              "folded": false,
+              "type": "BuiltinCommonInstructions::Standard",
+              "conditions": [],
+              "actions": [
+                {
+                  "type": {
+                    "inverted": false,
+                    "value": "ObjectSpawner::ObjectSpawner::SetPropertyJustSpawnedAnObject"
+                  },
+                  "parameters": [
+                    "Object",
+                    "Behavior",
+                    "no"
+                  ],
+                  "subInstructions": []
+                }
+              ],
+              "events": []
+            },
+            {
+              "disabled": false,
+              "folded": false,
+              "type": "BuiltinCommonInstructions::Comment",
+              "color": {
+                "b": 109,
+                "g": 230,
+                "r": 255,
+                "textB": 0,
+                "textG": 0,
+                "textR": 0
+              },
+              "comment": "Spawn an object when conditions are met",
+              "comment2": ""
+            },
+            {
+              "disabled": false,
+              "folded": false,
+              "type": "BuiltinCommonInstructions::Standard",
+              "conditions": [
+                {
+                  "type": {
+                    "inverted": false,
+                    "value": "CompareObjectTimer"
+                  },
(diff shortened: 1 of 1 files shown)


D8H:
I have doubts about this extension
* The behavior needs an action to be called every frame which is against the good practice.
* Is it something that can be widely use as it is?
  I'm under the impression that users could need finer conditions and should probably do their own extension.
* Could users mistake it as a player respawn system?

tristanbob:
> * The behavior needs an action to be called every frame which is against the good practice.

How can this be made otherwise?  I'd love to make it a lifecycle function, but I'm not sure how to do that since we can't use objects as behavior properties.

> * Is it something that can be widely use as it is?

I'm hoping so.  It has a frequency of creation, an offset (can be used for random placement or if the objects should appear as if coming out of a visible object), and a max limit on spawned creatures.  These are all basic features that a beginner might want.  Do you have ideas on how to make it more useful?

>   I'm under the impression that users could need finer conditions and should probably do their own extension.

Conditions are fine, they could make the spawners active when the player is near the spawner, while a boss is still alive, when a button is pushed, etc.  This is normal and expected.

> * Could users mistake it as a player respawn system?

I don't think so, but we could add wording in the description to clarify this.

D8H:
> > * Is it something that can be widely use as it is?
> 
> I'm hoping so. It has a frequency of creation, an offset (can be used for random placement or if the objects should appear as if coming out of a visible object), and a max limit on spawned creatures. These are all basic features that a beginner might want. Do you have ideas on how to make it more useful?

Probably an instance limit per spawner but it would require local links not to collide the one defined by users.

> > * The behavior needs an action to be called every frame which is against the good practice.
> 
> How can this be made otherwise? I'd love to make it a lifecycle function, but I'm not sure how to do that since we can't use objects as behavior properties.

I guess it's a limitation of the extension system. It could be split into a condition "Can spawn" and an action "Spawn" that always spawn when called to avoid an action that doesn't seem to do anything most of the time. But this probably defeat the purpose of the behavior.

This extension has 2 features:
* Repeat periodically
* Create an object around another object

Taken apart, these features are too much simple too require an extension; put together, they need a hack.

tristanbob:
I'm going to push back on this.  I agree that this extension is simple, but there are many simple things (both extensions and native features) in GDevelop that are added to increase the productivity of users.

First, I think that the periodic creation of objects is a very common task that will be used by a lot of users.

Secondly, I think this task is harder than you think for many beginning users.  Here is a recent example from Discord chat:
https://discord.com/channels/258623956158906368/258659467887902730/956218236091904030

```
I'm trying to have enemies spawn from a particular point,
 but no matter what I do there's either one of two outcomes: 
either nothing happens or it spawns an infinite amount of sprites. 
I initially asked for help from someone I know, and they even made
 me an example project so that I can see/understand the events. 

After several failed attempts at adding the events myself, 
changing them up, reading forum posts about spawning,
 checking the debugger, and even copy/pasting what my
 acquaintance did in his example project......... 
I still can't get this to work properly!
```

Regarding the implementation, I think a per spawner object limit might be more useful than a global limit.  I will see if I can implement this with linking.

> Probably an instance limit per spawner but it would require local links not to collide the one defined by users.

Collision with user linking is something to consider, but I have used linking in other extensions before.  In this case, we can specify in the description that links are automatically built between spawner and spawned objects.

4ian:
I would be fine to try to have this extension in the extension library (if it's too confusing, we can always remove it in the future).
It's not a lot, but can make things simpler for people that just want to spawn something regularly.

D8H:
* The behavior name should not be the same as the description. What about "Object Spawner" like the internal name?
* What is in parenthesis in the function name can be removed. The names allow a rapid scan. Any needed precisions are already in the description.
* I think that the purpose of the extension description is to let people know "why" they should use the extension but not "how". Wiki, examples and function descriptions already fulfill this role. I personally usually give a list of abstract features with an example of application.

ObjectWasJustSpawned
* Double space in the description.

SpawnObjects
* There are 2 properties set in this action. I think it's easier to read if properties are reset right before they are evaluated.
* "Take into account all object linked to" can be avoided when "MaxQuantity" is 0.
* When the number of spawn object reached the maximum for a long time. Could some games need the timer to start instead of triggering a spawn?

tristanbob:
@D8H Thank you once again, for an excellent review!

> * The behavior name should not be the same as the description. What about "Object Spawner" like the internal name?

I agree, changed.

> * What is in parenthesis in the function name can be removed. The names allow a rapid scan. Any needed precisions are already in the description.

Good call.  Changed.

> * I think that the purpose of the extension description is to let people know "why" they should use the extension but not "how". Wiki, examples and function descriptions already fulfill this role. I personally usually give a list of abstract features with an example of application.

I'm conflicted on this.  I think it is very useful to have basic instructions in the extension, and more details in the wiki.  I left the descriptions as is.

> * The "frequency" property is actually a "period" property.

I actually knew you were going to point this out.   :)   I will change it.

> ObjectWasJustSpawned
> 
> * Double space in the description.

Removed extra space.

> SpawnObjects
> 
> * There are 2 properties set in this action. I think it's easier to read if properties are reset right before they are evaluated.

That sounds like a good idea.  Changed.

> * "Take into account all object linked to" can be avoided when "MaxQuantity" is 0.

Is this what you mean? 

Screenshot 2022-03-26 161100

> * When the number of spawn object reached the maximum for a long time. Could some games need the timer to start instead of triggering a spawn?

Ok, I think I know what you mean.  If there are 3 enemies, and I kill one of them, if the spawn period has been reached then it will instantly create a new enemy.  Let me think about this a little.  Should I give users the option, or simply restart the counter when the number of enemies is below maximum?  I'm worried it might confuse the user.

D8H:
> > * I think that the purpose of the extension description is to let people know "why" they should use the extension but not "how". Wiki, examples and function descriptions already fulfill this role. I personally usually give a list of abstract features with an example of application.
> 
> I'm conflicted on this. I think it is very useful to have basic instructions in the extension, and more details in the wiki. I left the descriptions as is.

If there is both "why" and "how", I'm fine with it.

> > * "Take into account all object linked to" can be avoided when "MaxQuantity" is 0.
> 
> Is this what you mean?
> 
> Screenshot 2022-03-26 161100

Yes and the last event becomes useless.

> > * When the number of spawn object reached the maximum for a long time. Could some games need the timer to start instead of triggering a spawn?
> 
> Ok, I think I know what you mean. If there are 3 enemies, and I kill one of them, if the spawn period has been reached then it will instantly create a new enemy. Let me think about this a little. Should I give users the option, or simply restart the counter when the number of enemies is below maximum? I'm worried it might confuse the user.

Yes, it's not a trivial choice. The use case I see is in a game design perspective to avoid players to feel their actions as useless: enemy respawn right after a kill. But there are certainly use cases for the other way too. It seems hard to explain it extension users.

tristanbob:
> If there is both "why" and "how", I'm fine with it.

Ah, I think I need to add a "why" users would want to use this extension.  

> Yes, it's not a trivial choice. The use case I see is in a game design perspective to avoid players to feel their actions as useless: enemy respawn right after a kill. But there are certainly use cases for the other way too. It seems hard to explain it extension users.

Check out my latest commit, I changed it to reset the period when leaving the MaxQuantityReached state.  As this probably makes the most sense to users, I think I'll leave it this way.  (Great insight, BTW)

tristanbob:
I also updated the project files to make it easier to test this.  Click on a cat to delete them.

D8H:
* The setters full names doesn't start with "Set" in the other actions of the engine.
* The setters description and sentence usually start with "Change the"
* What is in parenthesis in the function name can be removed. The names allow a rapid scan. Any needed precisions are already in the description.

SpawnObject
* The link condition can be factorized to avoid doing it twice.
* I think it's easier to read if MaxQuantity is set to "no" at the beginning because this way there is only the conditions on the action that set it to "yes" to look to understand the logic.

tristanbob:
> * The setters full names doesn't start with "Set" in the other actions of the engine.

I thought using "Set" in front of functions that set properties was correct. 
```
Use PascalCase for actions, conditions, expression, function and parameter names.

For example, name your action: “SetRangeOfAttack”, and the expression: “RangeOfAttack”.
```
https://wiki.gdevelop.io/gdevelop5/extensions/best-practices

What do you think it should be instead?

> * The setters description and sentence usually start with "Change the"

Changed "Set the" to " Change the"

> * What is in parenthesis in the function name can be removed. The names allow a rapid scan. Any needed precisions are already in the description.

I removed all parenthesis in function names.

> SpawnObject
> 
> * The link condition can be factorized to avoid doing it twice.

I'm not sure what you mean by this.  Can you explain further?

> * I think it's easier to read if MaxQuantity is set to "no" at the beginning because this way there is only the conditions on the action that set it to "yes" to look to understand the logic.

I don't know how to do that with the current logic of restarting the spawning period when leaving the MaxQuantityReached state.  Can you want to edit the extension to include your idea for this?

D8H:
Changes
* Factorized the link condition
* Add an expression "Remaining time before spawn" to allow visual effects to let players anticipate (added a tint in the example).
* Various description changes.

https://www.dropbox.com/s/lfqy70kqi4ihodg/ObjectSpawnerExtension_REVIEW.zip?dl=1

Feel free to change anything.

tristanbob:
@D8H I like your changes!  

Do you want to officially approve this?

4ian (approved):
Looking great :) Thanks everyone for your inputs on this :) 

Just to confirm: actions should indeed start with "Set" or a verb (like "SpawnObjects", "ResetSomething", "DoSomething"...). While conditions/expressions can simply be the name of what is tested.
In the future, an expression will be able to be transformed in a condition generated automatically from the expression.

PR #381, opened 2022-03-18
New extension: Record

Add00:
### Description
Utilizes JavaScript built in [Media Recording API ](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder) to provide a cross platform solution to recording videos within GDevelop games.

#### Basics
At a minimum to create a video use the `start`, `stop` and `save` actions. This will make a video generally good default options.

`start` initializes the recording with any options specified. 
`stop` ends the recording.
`save` if the user is on desktop the file is written directly to the filesystem like in the screenshot extension. If the user is on the browser the video will open in a new window where they can right click and download it.
`pause` halts recording until the `resume` action is used.
`resume` continues a paused recording.

#### Conditions
The `OnPause`, `OnStart` and other methods that start with `On` allow for developers to check the current state of the recording. Enabling them show different information to the viewer based on the state.

The last condition is the `IsFormatSupported` which checks if a video format is supported on the users device.

#### Option actions
These functions are in setter/getter pairs and include: `FrameRate`/`SetFrameRate`, `VideoBitRate`/`SetVideoBitRate`, and `Format`/`SetFormat`. The setters are actions and the getters are expressions in the engine. These actions allow for developers to change options on the recording to better fit their needs. 

If the format is set to an unsupported format for a user the extension will fall back to mp4 which has the widest support.

#### Disabled functions
Currently, there are 4 disabled functions in the extension. The error handling ones are disabled since browser support is not [universal](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/error_event#browser_compatibility). However, node seems to good support for it so developers who know what they are doing can enable it to use. 

As for the audio functions I could not get video's to properly record sound, so I disabled them. I figure that the extension still has a lot of use cases even if it can't record audio.

### Checklist
- [x] I've followed all of [the best practices](http://wiki.compilgames.net/doku.php/gdevelop5/extensions/best-practices).
- [x] I confirm that this extension can be integrated to this GitHub repository, distributed and MIT licensed. 

### Example
Working example project:
[record-example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8306045/record-example.zip)

Live web example:
[web example](https://liluo.io/instant-builds/702f3246-9a53-4615-a72f-472af06133e0)

I ended up writing wayyyy more then I thought I would. Sorry for making you read so much 😅. I've also only been able to test on Windows, Firefox, and Chromium. So if someone could test the extension on Mac and Safari that would be really helpful.


arthuro555:
[The automatic tests found issues with your extension](https://app.circleci.com/pipelines/github/GDevelopApp/GDevelop-extensions/601/workflows/7e5d7b1a-fc1c-4e3d-895f-4c8d38614962/jobs/1566/parallel-runs/0/steps/0-103), please fix those.

Add00:
I fixed the issue with the descriptions, but now it is saying that there is a different problem.
image

I've never had this type of error before, how would I go about fixing it?

arthuro555:
The problem is the javascript disallowed properties thing, but that rule is pretty much wrong and can be ignored

arthuro555:
> The error handling ones are disabled since browser support is not [universal](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/error_event#browser_compatibility)

After checking your link, it seems that the support has no problems. The only thing is that some browsers use a plain event instead of a media error as the spec says, but both have no difference (and because of that it's currently even [being removed from the spec](https://github.com/w3c/mediacapture-record/pull/212)). So error handling should work just fine everywhere.

arthuro555:
image
Try removing such leftover debug codes, even if they are commented out.

image
Extensions need to work on all platforms, try implementing it on mobile too. Maybe you could just reuse the web one, since there is not really a good way to save files on mobile right now. Could you also test if that API is supported on phone builds (not remote preview)? Cordova tends to be stripped of a lot of Web APIs by default.

image
No need for async, remove the console.log, and maybe you could see if you could add an option to download the file directly instead of opening it in a tab, by using an `<a>` element with the download attribute?

image
This doesn't needs to be a function and can be simplified to "are we using fs (electron) or a new tab (anything else)?":
image
We can go further by removing this and just checking if the FS module was loaded or not.


image

Remove this debugging code

image
This code is slow and gets executed often for nothing, instead, do: 
image


image

This can be const, since an array is not a primitive, and therefore a reference. Since you will not make that variable point to another array, and keep on working with the same array, it can be const.

image
Something should probably be done if a video is already recording (either stop the previous one or ignore the new start)

image

This should be useless, iirc node already takes care of making the path fit to the OS's path style. Especially if you make a unix path only to add a windows separator:
image

image
You shouldn't recreate the function every time the condition is callled, that will put stress on the garbage collector since it'll neeed to constantly delete the ild functions hich are the exact same! Instead, store the functions on the recorder object in the onFirstSceneLoaded.


I did a few of the changes I suggested quickly, here it is if it helps you:
[Record.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8310248/Record.zip)


Pretty cool extension so far, except for this few code quality details I think it is pretty well designed and will be a great addition.

Add00:
> Could you also test if that API is supported on phone builds (not remote preview)? Cordova tends to be stripped of a lot of Web APIs by default.

I spent a ton of time trying to get the extension to run on mobile with no luck. I have very little experience with Cordova, so that could be why. I did find, however that in the [documentation](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-media-capture/index.html#capturecapturevideo) that it has video capture capabilities. So it should be theoretically possible. Is it necessary for this to work on mobile? I know that there are a few platform specific extensions like the Cursor Movement Extension which only works on computers. 

> instead of opening it in a tab, by using an `<a>` element with the download attribute?

I tried to implement downloading through an `a` tag, but I ended up getting CORS errors. So I've kept the downloading the same.

> Something should probably be done if a video is already recording (either stop the previous one or ignore the new start)

I've set it so the start action again will override the previous media recording. I've also, fixed an issue where the save action could be called before the stop action. Additionally, stopped allowing for stop and save to be called before start. If users call actions out of order it should no longer create fatal errors.

> You shouldn't recreate the function every time the condition is called, that will put stress on the garbage collector since it'll need to constantly delete the functions which are the exact same! Instead, store the functions on the recorder object in the onFirstSceneLoaded.

I tried placing the functions in the `onFirstSceneLoaded` but that lead to issues since the `mediaRecorder` has to be created in the start function. So there was no `mediaRecorder` to attach functions to. To fix this I swapped the `onEvents` for `addEventListeners` in the start function. This should resolve the garbage collector issue since the functions are only created once and after the `mediaRecorder` is created. 

> Pretty cool extension so far, except for this few code quality details I think it is pretty well designed and will be a great addition.

Thanks so much for your feedback and code improvements, it really makes the extension better.

arthuro555:
!merge

arthuro555:
That... Didn't work. I'll do it manually then

arthuro555:
Hello @Add00 👋 !

Add00:
I've updated it with the feedback that arthuro555 suggested to get full review approval

arthuro555 (approved):
👍 Sorry for the delay

arthuro555:
Whoops actually nevermind the tests are not happy 😬

arthuro555 (changes requested):
I fixed the errors raised by the automatic tests, but that prompted me to retest everything once again, and I noticed that when pausing in your example, it acts as if I stopped the recording: footage after resuming and before stopping is not present.

Add00:
I've fixed the issue that stop the pause feature from updating

arthuro555 (approved):
Ok I've tested it, now everything should be good to go 👍

PR #384, opened 2022-03-26
New extension: EdgeScrollCamera

tristanbob:
Edge scroll camera allows the player to move around the scene simply by moving the cursor near the edge they want the camera to move towards.  This is common in RTS games, but it can be useful for many things.

How to use:
- Run the action "Enable (or disable) edge camera scrolling" at the beginning of scene
- Run the action "Configure edge camera scrolling" at the beginning of scene

Tips:
- Variable speed makes the camera scroll faster the closer the cursor gets to the edge of screen
- Use the action "Enable (or disable) edge camera scrolling" to turn the scrolling ON or OFF
- Use the action "Enforce camera boundaries" to restrict what the player can see
- Use the extension "Copy camera settings" if you want multiple layers to move

## Playable demo

~https://liluo.io/games/d29e90cd-058c-42ef-bc00-819361146082~

## Project files

~[EdgeScrollCamera_Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8367173/EdgeScrollCamera_Extension.zip)~

## Video

https://user-images.githubusercontent.com/8879811/160510078-aeb49e5f-834d-4637-a3ce-3a315275c4b3.mp4

### DIFF
--- a/Extensions/EdgeScrollCamera.json
+++ b/Extensions/EdgeScrollCamera.json
@@ -8,7 +8,7 @@
   "iconUrl": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz48IURPQ1RZUEUgc3ZnIFBVQkxJQyAiLS8vVzNDLy9EVEQgU1ZHIDEuMS8vRU4iICJodHRwOi8vd3d3LnczLm9yZy9HcmFwaGljcy9TVkcvMS4xL0RURC9zdmcxMS5kdGQiPjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0ibWRpLWNhbWVyYS1tZXRlcmluZy1tYXRyaXgiIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNNCw0SDIwQTIsMiAwIDAsMSAyMiw2VjE4QTIsMiAwIDAsMSAyMCwyMEg0QTIsMiAwIDAsMSAyLDE4VjZBMiwyIDAgMCwxIDQsNE00LDZWMThIMjBWNkg0TTUuNSw3LjVIMTFWOS4xN0MxMC4xNSw5LjQ3IDkuNDcsMTAuMTUgOS4xNywxMUg1LjVWNy41TTE4LjUsNy41VjExSDE0LjgzQzE0LjUzLDEwLjE1IDEzLjg1LDkuNDcgMTMsOS4xN1Y3LjVIMTguNU0xOC41LDE2LjVIMTNWMTQuODNDMTMuODUsMTQuNTMgMTQuNTMsMTMuODUgMTQuODMsMTNIMTguNVYxNi41TTUuNSwxNi41VjEzSDkuMTdDOS40NywxMy44NSAxMC4xNSwxNC41MyAxMSwxNC44M1YxNi41SDUuNU0xMiwxMC41QTEuNSwxLjUgMCAwLDEgMTMuNSwxMkExLjUsMS41IDAgMCwxIDEyLDEzLjVBMS41LDEuNSAwIDAsMSAxMC41LDEyQTEuNSwxLjUgMCAwLDEgMTIsMTAuNVoiIC8+PC9zdmc+",
   "name": "EdgeScrollCamera",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/camera-metering-matrix.svg",
-  "shortDescription": "Scroll camera when mouse is near edge of screen",
+  "shortDescription": "Scroll camera when mouse is near edge of screen.",
   "version": "0.1.0",
   "tags": [
     "camera",
@@ -22,7 +22,7 @@
   "dependencies": [],
   "eventsFunctions": [
     {
-      "description": "Scroll camera when mouse is near edge of screen",
+      "description": "Scroll camera when mouse is near edge of screen.",
       "fullName": "Scroll camera when mouse is near edge of screen",
       "functionType": "Action",
       "group": "",
@@ -47,6 +47,17 @@
               "type": "BuiltinCommonInstructions::Standard",
               "conditions": [],
               "actions": [
+                {
+                  "type": {
+                    "inverted": false,
+                    "value": "SetSceneVariableAsBoolean"
+                  },
+                  "parameters": [
+                    "__EdgeScrollCamera.Moving",
+                    "False"
+                  ],
+                  "subInstructions": []
+                },
                 {
                   "type": {
                     "inverted": false,
@@ -87,17 +98,6 @@
                   },
                   "parameters": [
                     "__EdgeScrollCamera.MovingRight",
-                    "False"
-                  ],
-                  "subInstructions": []
-                },
-                {
-                  "type": {
-                    "inverted": false,
-                    "value": "SetSceneVariableAsBoolean"
-                  },
-                  "parameters": [
-                    "__EdgeScrollCamera.Moving",
                     "False"
                   ],
                   "subInstructions": []
--- a/Extensions/EdgeScrollCamera.json
+++ b/Extensions/EdgeScrollCamera.json
@@ -1,7 +1,7 @@
 {
   "author": "",
   "category": "General",
-  "description": "Edge scroll camera allows the player to move around the scene simply by moving the cursor near the edge they want the camera to move towards.  This is common in RTS games, but it can be useful for many things.\n\nHow to use:\n- Run the action \"Enable (or disable) edge camera scrolling\" at the beginning of scene\n- Run the action \"Configure edge camera scrolling\" at the beginning of scene\n\nTips:\n- Variable speed makes the camera scroll faster the closer the cursor gets to the edge of screen\n- Use the action \"Enable (or disable) edge camera scrolling\" to turn the scrolling ON or OFF\n- Use the action \"Enforce camera boundaries\" to restrict what the player can see\n- Use the extension \"Copy camera settings\" if you want multiple layers to move\n\nExpressions:\n- EdgeScrollSpeedX() - Current scroll speed (horizontal)\n- EdgeScrollSpeedY() - Current scroll speed (vertical)\n",
+  "description": "Edge scroll camera allows the player to move around the scene simply by moving the cursor near the edge they want the camera to move towards.  This is common in RTS games, but it can be useful for many things.\n\nHow to use:\n- Run the action \"Enable (or disable) camera edge scrolling\" at the beginning of scene\n- Run the action \"Configure camera edge scrolling\" at the beginning of scene\n\nTips:\n- Progressive speed makes the camera scroll faster the closer the cursor gets to the edge of screen\n- Use the action \"Enable (or disable) edge camera scrolling\" to turn the scrolling ON or OFF\n- Use the action \"Enforce camera boundaries\" to restrict what the player can see\n- Use the extension \"Copy camera settings\" if you want multiple layers to move\n\nExpressions:\n- EdgeScrollSpeedX() - Current scroll speed (horizontal)\n- EdgeScrollSpeedY() - Current scroll speed (vertical)\n",
   "extensionNamespace": "",
   "fullName": "Edge scroll camera",
   "helpPath": "",
@@ -106,7 +106,7 @@
                   "colorG": 176,
                   "colorR": 74,
                   "creationTime": 0,
-                  "name": "Prevent camera movement until cursor has move away from starting position",
+                  "name": "Prevent camera movement until cursor has moved (cursor starts at 0,0 which triggers scrolling)",
                   "source": "",
                   "type": "BuiltinCommonInstructions::Group",
                   "events": [
@@ -669,7 +669,7 @@
                                             "value": "SceneVariableAsBoolean"
                                           },
                                           "parameters": [
-                                            "__EdgeScrollCamera.VariableSpeedEnabled",
+                                            "__EdgeScrollCamera.ProgressiveSpeedEnabled",
                                             "False"
                                           ]
                                         }
@@ -772,7 +772,7 @@
                                             "value": "SceneVariableAsBoolean"
                                           },
                                           "parameters": [
-                                            "__EdgeScrollCamera.VariableSpeedEnabled",
+                                            "__EdgeScrollCamera.ProgressiveSpeedEnabled",
                                             "True"
                                           ]
                                         }
@@ -1082,13 +1082,13 @@
       "objectGroups": []
     },
     {
-      "description": "Enable (or disable) edge scrolling camera.  Use \"Configure edge scrolling camera\" to adjust settings.",
-      "fullName": "Enable (or disable) edge scrolling camera",
+      "description": "Enable (or disable) camera edge scrolling .  Use \"Configure camera edge scrolling\" to adjust settings.",
+      "fullName": "Enable (or disable) camera edge scrolling",
       "functionType": "Action",
       "group": "",
       "name": "EnableEdgeScrollCamera",
       "private": false,
-      "sentence": "Enable edge scrolling camera? _PARAM1_",
+      "sentence": "Enable camera edge scrolling: _PARAM1_",
       "events": [
         {
           "type": "BuiltinCommonInstructions::Comment",
@@ -1160,7 +1160,7 @@
         {
           "codeOnly": false,
           "defaultValue": "yes",
-          "description": "Enable edge scrolling camera?",
+          "description": "Enable camera edge scrolling",
           "longDescription": "",
           "name": "EnableEdgeScrolling",
           "optional": true,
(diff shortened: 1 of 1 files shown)


D8H:
Even if the example is very simple, please attach one to the description.

tristanbob:
I agree, having an example makes it much easier to review.  An example project is now attached to the original post.

Extensions/EdgeScrollCamera.json
```diff
@@ -0,0 +1,1473 @@
...
+          "optional": false,
+          "supplementaryInformation": "",
+          "type": "expression"
+        }
+      ],
+      "objectGroups": []
+    },
+    {
+      "description": "Check if the camera is scrolling.",
+      "fullName": "Is camera scrolling?",
```
Bouh, line 1253:
```suggestion
      "fullName": "Is camera scrolling",
```

Extensions/EdgeScrollCamera.json
```diff
@@ -0,0 +1,1473 @@
...
+          ],
+          "events": []
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Check if the camera is scrolling up.",
+      "fullName": "Is camera scrolling up?",
```
Bouh, line 1297:
```suggestion
      "fullName": "Is camera scrolling up",
```

Extensions/EdgeScrollCamera.json
```diff
@@ -0,0 +1,1473 @@
...
+          ],
+          "events": []
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Check if the camera is scrolling down.",
+      "fullName": "Is camera scrolling down?",
```
Bouh, line 1341:
```suggestion
      "fullName": "Is camera scrolling down",
```

Extensions/EdgeScrollCamera.json
```diff
@@ -0,0 +1,1473 @@
...
+          ],
+          "events": []
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Check if the camera is scrolling left.",
+      "fullName": "Is camera scrolling left?",
```
Bouh, line 1385:
```suggestion
      "fullName": "Is camera scrolling left",
```

Extensions/EdgeScrollCamera.json
```diff
@@ -0,0 +1,1473 @@
...
+          ],
+          "events": []
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Check if the camera is scrolling right.",
+      "fullName": "Is camera scrolling right?",
```
Bouh, line 1429:
```suggestion
      "fullName": "Is camera scrolling right",
```

Extensions/EdgeScrollCamera.json
```diff
@@ -0,0 +1,1473 @@
...
+  "dependencies": [],
+  "eventsFunctions": [
+    {
+      "description": "Scroll camera when mouse is near edge of screen.",
+      "fullName": "Scroll camera when mouse is near edge of screen",
+      "functionType": "Action",
+      "group": "",
+      "name": "EdgeScrollCamera",
+      "private": false,
+      "sentence": "Scroll camera when mouse is near edge of screen (Layer: _PARAM9_, Trigger width: Top _PARAM1_ px, Bottom _PARAM2_ px, Left _PARAM3_ px, Right _PARAM4_ px, Move speed: Up _PARAM5_ px/second, Down _PARAM6_ px/second, Left _PARAM7_ px/second, Right _PARAM8_ px/second)",
```
Bouh, line 31: Lot of option, the sentence is so long, might be possible to hide some parameters?

Extensions/EdgeScrollCamera.json
```diff
@@ -0,0 +1,1473 @@
...
+                  "type": "BuiltinCommonInstructions::Comment",
+                  "color": {
+                    "b": 109,
+                    "g": 230,
+                    "r": 255,
+                    "textB": 0,
+                    "textG": 0,
+                    "textR": 0
+                  },
+                  "comment": "NOTE: Speeds are multiplied by \"0.7071\" to keep the linear speed constant",
```
Bouh, line 749: How did you find this magic value?

D8H: Why not use `Sqrt(2)`?
sqrt(2) = 0.70710678118654752440084436210485

D8H:
GDevelop doesn't handle mouse events outside of the rending area correctly. Rapid mouse movement won't be detected as the cursor position stays at the last cursor inside position.

In the engine, I think we should add conditions and fixes on cursor event handling:
- Add a new condition "Cursor is inside the screen"
- Fix: continue to get the position of the cursor outside of the screen ([mousemoveoutside](https://github.com/pixijs/pixijs/issues/1564))
- Fix: get button release events when the cursor is outside of the screen ([pointerupoutside](http://pixijs.download/alpha-filter/docs/PIXI.interaction.InteractionManager.html#event:pointerupoutside))

This should allow to get ride of the margin.

D8H:
What is the reason for allowing different speed and margin for 4 directions?

I'm not sure a normalized speed is better for a camera. For instance, what if the camera is moving diagonally to the bottom right and the scene bottom is already reached. Personally, I would prefer it to keep full speed to the right.

A modified version of the extension that uses the new mouse condition:
https://www.dropbox.com/s/8slmj26tw8kgylq/EdgeScrollCamera_Extension_REVIEW.zip?dl=1

Extensions/EdgeScrollCamera.json
tristanbob: Should we also change the placeholder text on the extension editor?  (because it uses a question mark in the suggested text)

Extensions/EdgeScrollCamera.json
tristanbob: Do you think people will want different borders and speeds per side?  In most games I have played, those values are the same on all 4 sides.

Extensions/EdgeScrollCamera.json
tristanbob: > How did you find this magic value?

It is cos(45) and sin(45) and apparently also "sqrt(2)".  I used a constant to save CPU, but maybe it is not a good idea?

tristanbob:
> GDevelop doesn't handle mouse events outside of the rending area correctly. Rapid mouse movement won't be detected as the cursor position stays at the last cursor inside position.
>
> In the engine, I think we should add conditions and fixes on cursor event handling:
> 
> * Add a new condition "Cursor is inside the screen"
> * Fix: continue to get the position of the cursor outside of the screen ([Fixed in PIXI 6.3.0](https://github.com/pixijs/pixijs/commit/db6deeb891cdd7987a6f63b4aa4930aaf5e66e71), a similar fix can probably be done in GDevelop)
> * Fix: get button release events when the cursor is outside of the screen ([pointerupoutside](http://pixijs.download/alpha-filter/docs/PIXI.interaction.InteractionManager.html#event:pointerupoutside))

These would be very useful!

tristanbob:
> What is the reason for allowing different speed and margin in 4 directions?

I wanted to provide flexibility for the user.  Do you think someone might use a different speed for each side?   

I do think we can consolidate the margin into a single value.

> I'm not sure a normalized speed is better for a camera. For instance, what if the camera is moving diagonally to the bottom right and the scene bottom is already reached. Personally, I would prefer it to keep full speed to the right.

I don't have a strong opinion on this.  What do other people think?  Should the camera move slightly faster in the corners?

> A modified version of the extension that uses the new mouse condition: https://www.dropbox.com/s/8slmj26tw8kgylq/EdgeScrollCamera_Extension_REVIEW.zip?dl=1

Yes! A great improvement.

Extensions/EdgeScrollCamera.json
Bouh: One value for speed is fine yeah.

Bouh:
- I agree with Davy, keep the same speed in a corner and on sides.

> > Davy: I'm not sure a normalized speed is better for a camera. For instance, what if the camera is moving diagonally to the bottom right and the scene bottom is already reached. Personally, I would prefer it to keep full speed to the right.
>
> Tristan: I don't have a strong opinion on this. What do other people think? Should the camera move slightly faster in the corners?

- I agree also with Davy.

- One speed for all directions is fine.

Keep in mind it's easier to add things later rather adds lot on options and remove them after.

D8H:
To fix the issue Bouh reported on the other PR, I added a condition to avoid scrolling at the beginning of the scene.
https://www.dropbox.com/s/8slmj26tw8kgylq/EdgeScrollCamera_Extension_REVIEW.zip?dl=1

tristanbob:
> * I agree with Davy, keep the same speed in a corner and on sides.

Which do you prefer:

A) Camera moves slightly faster when in corners (because it moves the horizontal and vertical speed) 

B) Camera always moves at the same speed in all directions, EXCEPT when hitting an enforced boundary.  In this case, it will move slightly slower when moving in a corner.

I still like the constant speed option B.

> * One speed for all directions is fine.
> Keep in mind it's easier to add things later rather adds lot on options and remove them after.

I changed to a single speed and single trigger width.

D8H:
> A modified version of the extension that uses the new mouse condition:
> https://www.dropbox.com/s/8slmj26tw8kgylq/EdgeScrollCamera_Extension_REVIEW.zip?dl=1

Did you used the new condition to check if the mouse is outside and removed the speed normalization?

tristanbob:
> Did you use the new condition to check if the mouse is outside and removed the speed normalization?

Yes.  Changes in this commit:

- Removed diagonal speed adjustment
- Does not move camera when cursor leaves screen
- Does not move camera until cursor has left starting position (0,0)

## Project file

[EdgeScrollCamera_Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8794029/EdgeScrollCamera_Extension.zip)

## Playable game

https://liluo.io/instant-builds/f80cb633-1053-4fa7-af25-5be8b0a72685

D8H:
- The camera can be moved directly in the detection events.
- "Scroll speed (pixels/second)" changed to "Scroll speed (in pixels per second)"?
- "Trigger width (pixels)" changed to "Screen margin" (no need for unit when it's just pixels I guess)
- Replace "mouse" by "cursor"
- In the sentence
  - "Move speed" should be "Scroll speed"
  - parameter name should be lower case
- What do you think about actions to set up and enable/disable and using the life-cycle event to run it. This way it follows a logic a bit like a behavior.

Ideas for a next versions:
- an acceleration over time
- or a progressive speed according to the distance between the cursor and the border.

tristanbob:
> * The camera can be moved directly in the detection events.

Now that I added variable speed, I have a shared event that runs after the detection.  If I consolidated the "move" events, I'd have to insert a copy of this event under each detection.  Short story; I want to leave it as is.

> * "Scroll speed (pixels/second)" changed to "Scroll speed (in pixels per second)"?

Done.

> * "Trigger width (pixels)" changed to "Screen margin" (no need for unit when it's just pixels I guess)

Done.

> * Replace "mouse" by "cursor"

Done.

> * In the sentence
>   
>   * "Move speed" should be "Scroll speed"

Done.

>   * parameter name should be lower case

Done.

> * What do you think about actions to set up and enable/disable and using the life-cycle event to run it. This way it follows a logic a bit like a behavior.

I liked this idea enough that I did it.  The downside is that this extension will only work for one layer at a time.  I made a note to use "Copy camera settings" if the user wants multiple layers to move at the same time.

> Ideas for a next versions:
> 
> * or a progressive speed according to the distance between the cursor and the border.

This idea I LOVED so I went ahead and added it.  Thanks for all your great ideas!

I also copied one of your "patterns" by adding a "Draw edge scrolling screen margin".  I hope you like it!

## Playable game

https://liluo.io/victrisgames/edge-scrolling-camera

## Project files

[EdgeScrollCamera_Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8839289/EdgeScrollCamera_Extension.zip)

4ian:
Very nice! I love these simple but useful extensions!

D8H:
Some suggestions
- A default configuration can be set in onSceneLoaded
- In the "configure"some events could be merged
- In the "enable" action
  - as it's disabled by default, it might be used to enable more often than to disable, the default value could be `Yes`
  - the 1st condition is not necessary
- Scrolling speed on X and Y should be independent (here the X has the precedence)
- A "current speed" variable might be easier to use than a "ratio". It could be given to users through expressions for instance.

tristanbob:
Playable game:
https://liluo.io/victrisgames/edge-scrolling-camera

Project files:
[EdgeScrollCamera_Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8884653/EdgeScrollCamera_Extension.zip)


> * A default configuration can be set in onSceneLoaded

Good idea, done.

> * In the "configure" some events could be merged

I usually split actions so I can add comments, but in this case, I am not doing that.  I have combined the actions into one event, done.

>   * as it's disabled by default, it might be used to enable more often than to disable, the default value could be `Yes`

Great idea!  Done.

**BTW, do you think I should also enable scrolling in OnSceneLoad so that as soon as someone installs this extension, it starts activated with default values?**

>   * the 1st condition is not necessary

I prefer to keep it so the logic is obvious to non-coders.  However, I removed the condition.

> * Scrolling speed on X and Y should be independent (here the X has the precedence)

Good catch!  I have split them to be independent.

> * A "current speed" variable might be easier to use than a "ratio". It could be given to users through expressions for instance.

I'm not sure how many people would use this value, but I went ahead and implemented it as two expressions (X and Y).

D8H:
> **BTW, do you think I should also enable scrolling in OnSceneLoad so that as soon as someone installs this extension, it starts activated with default values?**

An extension without any instruction in the events sheet of a scene won't run for this scene. Forcing the user to call the "configure" action allow to make sure it doesn't happen (and if it weren't the case, it might be strange to automatically start it for any scene).

Some suggestions
* To avoid repetition
  * from `EdgeScrollCamera::EdgeScrollSpeedX()`
  * to `EdgeScrollCamera::SpeedX()`
* According to best practive, there should never be interogative form.
  * from `Enable edge scrolling camera? _PARAM1_`
  * to `Enable camera edge scrolling: _PARAM1_`?
* The sentence could lead people think that calling the action on each layer would work. The description probably also needs to make it clear.
  * from `Configure edge scrolling camera on layer _PARAM3_, screen margin _PARAM1_, scroll speed _PARAM2_, use variable speed? _PARAM5_`
  * to `Configure camera edge scrolling (layer _PARAM3_, margin _PARAM1_, speed _PARAM2_, progressive _PARAM5_)`?
 * "horizontal" seems important enough not to be in parentesis.
  * from `Current edge scroll speed (horizontal)`
  * to `Current horizontal scrolling speed`
* New expressions description defined inside GDevelop are prefixed by "Return". I guess it could be a best practice.
  * from `Speed the camera is currently scrolling in horizontal direction (in pixels per second)`
  * to `Return the speed the camera is currently scrolling in horizontally direction (in pixels per second)`
* `__EdgeScrollCamera.CursorMoved` variable doesn't seem useful.
* `__EdgeScrollCamera.MovingUp` is the same as `__EdgeScrollCamera.CurrentScrollSpeedY < 0`, the code can be simplified.
* `__EdgeScrollCamera.SpeedPercentageX` variable doesn't seem useful.

arthuro555:
Hello @tristanbob 👋 !

tristanbob:
> * To avoid repetition
>   * from `EdgeScrollCamera::EdgeScrollSpeedX()`
>   * to `EdgeScrollCamera::SpeedX()`?

Changed.

> * According to best practice, there should never be interrogative form.
>   
>   * from `Enable edge scrolling camera? _PARAM1_`
>   * to `Enable camera edge scrolling: _PARAM1_`?

Changed.

> * The sentence could lead people to think that calling the action on each layer would work. The description probably also needs to make it clear.
>   * from `Configure edge scrolling camera on layer _PARAM3_, screen margin _PARAM1_, scroll speed _PARAM2_, use variable speed? _PARAM5_`
>   * to `Configure camera edge scrolling (layer _PARAM3_, margin _PARAM1_, speed _PARAM2_, progressive _PARAM5_)`?

Changed.

> * "horizontal" seems important enough not to be in parenthesis.
> * from `Current edge scroll speed (horizontal)`
> * to `Current horizontal scrolling speed`?

Changed.

>  * New expressions description defined inside GDevelop are prefixed by "Return". I guess it could be a best practice.
>   * from `Speed the camera is currently scrolling in horizontal direction (in pixels per second)`
>   * to `Return the speed the camera is currently scrolling in horizontally direction (in pixels per second)`?

Changed.

> * `__EdgeScrollCamera.CursorMoved` variable doesn't seem useful.

It is important. I updated the event group name: `Prevent camera movement until cursor has moved (cursor starts at 0,0 which triggers scrolling)`

> * `__EdgeScrollCamera.MovingUp` is the same as `__EdgeScrollCamera.CurrentScrollSpeedY < 0`, the code can be simplified.

This is a minor improvement but it requires a lot of work to implement.  I'm leaving it the same.

> * `__EdgeScrollCamera.SpeedPercentageX` variable doesn't seem useful.

I agree, I decided to use it to increase readability by breaking up a complex expression into two parts.  That said, I am leaving this the same.

I have pushed a new commit with these changes.

## Playable Game

https://liluo.io/victrisgames/edge-scrolling-camera

## Project files

[EdgeScrollCamera_Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/9183534/EdgeScrollCamera_Extension.zip)

D8H:
I simplified the events: 
https://www.dropbox.com/s/8y9wilycipapo0j/EdgeScrollCamera_Extension_REVIEW4.zip?dl=1

This also change:
- Use the new camera border expressions
- Make sure the margin is at least 1 pixel to be sure the scroll can always happen
- Set full speed when the mouse is less than 1 pixel from the border because the mouse can't always be exactly at the bottom or right border

I guess the debug action could be removed as margins is a concept that is easy to understand and its code is as big as the rest of the extension.

tristanbob:
> I simplified the events: https://www.dropbox.com/s/8y9wilycipapo0j/EdgeScrollCamera_Extension_REVIEW4.zip?dl=1

@D8H I love your changes, thank you.
 
> * Use the new camera border expressions

Good idea.

> * Set full speed when the mouse is less than 1 pixel from the border because the mouse can't always be exactly at the bottom or right border

I had noticed this problem so I'm glad you fixed it!

> I guess the debug action could be removed as margins is a concept that is easy to understand and its code is as big as the rest of the extension.

I really like the idea of including one-click debugging tools with extensions.  This saves a small amount of time, but it's still worth it in my opinion. 

I did update this function by using CameraBorder expressions and enforcing parameters (but only once) when the extension did not create the shape painter.

tristanbob:
@D8H can you review/approve this commit?

D8H:
What do you think about making the `ProgressiveSpeed` a string with choice instead of a boolean to allow to add other mode in following version. For instance, a 3rd mode could be a scroll speed that goes faster over time until it reaches the maximum speed.

tristanbob:
> What do you think about making the `ProgressiveSpeed` a string with choice instead of a boolean to allow to add other mode in following version? For instance, a 3rd mode could be a scroll speed that goes faster over time until it reaches the maximum speed.

I have changed from boolean to a string "scroll style" with "Progressive speed" (the default) and "Constant speed".

D8H (approved):

PR #386, opened 2022-03-31
New extension: CloneCamera - Copy the camera settings of a layer and apply them to another layer

tristanbob:
Useful when multiple layers need to use the same camera values.

How to use:
- Run the "Clone camera settings" action after all other camera actions have been performed

Tips:
- Do not use on layers that implement a parallax effect

## Playable game

https://liluo.io/games/1e988edc-a480-42c9-a6b4-b6454fb2454a

## Project file

[CloneCamera.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8393393/CloneCamera.zip)

## Video

https://user-images.githubusercontent.com/8879811/161161165-8342e1d3-cf29-46ee-b69e-71011d29aa7b.mp4

### DIFF
--- a/Extensions/CloneCamera.json
+++ b/Extensions/CloneCamera.json
@@ -1,0 +1,257 @@
+{
+  "author": "",
+  "category": "General",
+  "description": "Useful when multiple layers need to use the same camera values.\n\nHow to use:\n- Run the \"Clone camera settings\" action after all other camera actions have been performed\n\nTips:\n- Do not use on layers that implement a parallax effect",
+  "extensionNamespace": "",
+  "fullName": "Clone camera",
+  "helpPath": "",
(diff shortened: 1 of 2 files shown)


D8H:
* "Clone camera" might be confusing as it could be interpreted as "Fork Camera" multiple camera handling could comes back one day. Maybe "Copy camera settings" like it is in the sentence?
* Every parameter should be in the sentence (apart from the camera number)
* The icon is a bit generic, what about "layers-triple-outline" to convey that it aligns a lot of layer?

tristanbob:
> * "Clone camera" might be confusing as it could be interpreted as "Fork Camera". Multiple camera handling could comes back one day. Maybe "Copy camera settings" like it is in the sentence?

Sure, that will work.

> * Every parameter should be in the sentence (apart from the camera number)

Added all parameters to the sentence. 

> * The icon is a bit generic, what about "layers-triple-outline" to convey that it aligns a lot of layer?

Sure!

> * "General" is kind of the default category. I think it's better to leave it empty in case GDevelop rename it one day (it should probably not be in the list in the 1st place).

I was hoping for a Camera category, as I have made a handful that would fit under that name.  In the meantime, I will leave it blank if you think that is preferable.

D8H (approved):

PR #387, opened 2022-04-01
[ScreenShaking] Multi-layer, noise based shaking, scrolling-compatible.

D8H:
## Changes
* Embed the Noise extension for a more realistic shake (especially at low frequency)
* Configuration per layer
* Linear easing
* Fix conflicts with camera movements.

### With parallax
Try it:
* Project: https://www.dropbox.com/s/nai88cdnjg76ynm/ScreenShakeExamples.zip?dl=1
* Build: https://liluo.io/instant-builds/e0257129-0d90-46f3-b029-42c12f8b9854

### With impulse and scrolling
It also uses:
- https://github.com/GDevelopApp/GDevelop-extensions/pull/529

Press `I` to start an impulsion and `S` to shake.
I didn't find any conflict between the impulse and the scrolling or the screen shake.
 
- Project: https://www.dropbox.com/s/spvas50648blctc/platformer-with-tilemap-shake.zip?dl=1
- Build: https://games.gdevelop-app.com/game-881fc8e5-272e-460a-8d72-7d01baca255d/index.html

## JavaScript
It uses a JavaScript event but it's just a copy-paste of the Noise extension with a different name-space.

### DIFF
--- a/extensions/reviewed/CameraShake.json
+++ b/extensions/reviewed/CameraShake.json
@@ -1,10 +1,10 @@
 {
   "author": "westboy31, Tristan Rhodes (https://victrisgames.itch.io/)",
   "category": "",
-  "description": "Shake layer cameras with translation, rotation and zoom.\n- Short shaking can be used to give impact (explosion, hit)\n- Shaking can go indefinitly to set an ambiance (engine vibration, earthquake, pulsing)\n- Low frequency shaking allows to simulate slow moving objects (ship rocking back and forth)\n- Setting amplitudes per layer make it possible to respect the parallax and give more depth",
+  "description": "Shake layer cameras with translation, rotation and zoom.\n- Short shaking can be used to give impact (explosion, hit)\n- Shaking can go indefinitely to set an ambiance (engine vibration, earthquake, pulsing)\n- Low frequency shaking allows to simulate slow moving objects (ship rocking back and forth)\n- Setting amplitudes per layer make it possible to respect the parallax and give more depth\n\nRelease notes:\n- Version 3.0.0\n  - No adaptation of the game events is needed.\n  - It fixes an issue when used with scrolling, the amplitude will feel bigger in this case.\n  - The shaking relies on noise which could feel a bit different.",
   "extensionNamespace": "",
   "fullName": "Camera shake",
-  "helpPath": "https://victrisgames.itch.io/gdevelop-camera-shake-example",
+  "helpPath": "",
   "iconUrl": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz48IURPQ1RZUEUgc3ZnIFBVQkxJQyAiLS8vVzNDLy9EVEQgU1ZHIDEuMS8vRU4iICJodHRwOi8vd3d3LnczLm9yZy9HcmFwaGljcy9TVkcvMS4xL0RURC9zdmcxMS5kdGQiPjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0ibWRpLXZlY3Rvci1kaWZmZXJlbmNlLWFiIiB3aWR0aD0iMjQiIGhlaWdodD0iMjQiIHZpZXdCb3g9IjAgMCAyNCAyNCI+PHBhdGggZD0iTTMsMUMxLjg5LDEgMSwxLjg5IDEsM1Y1SDNWM0g1VjFIM003LDFWM0gxMFYxSDdNMTIsMVYzSDE0VjVIMTZWM0MxNiwxLjg5IDE1LjExLDEgMTQsMUgxMk0xLDdWMTBIM1Y3SDFNMTQsN0MxNCw3IDE0LDExLjY3IDE0LDE0QzExLjY3LDE0IDcsMTQgNywxNEM3LDE0IDcsMTggNywyMEM3LDIxLjExIDcuODksMjIgOSwyMkgyMEMyMS4xMSwyMiAyMiwyMS4xMSAyMiwyMFY5QzIyLDcuODkgMjEuMTEsNyAyMCw3QzE4LDcgMTQsNyAxNCw3TTE2LDlIMjBWMjBIOVYxNkgxNEMxNS4xMSwxNiAxNiwxNS4xMSAxNiwxNFY5TTEsMTJWMTRDMSwxNS4xMSAxLjg5LDE2IDMsMTZINVYxNEgzVjEySDFaIiAvPjwvc3ZnPg==",
   "name": "CameraShake",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/vector-difference-ab.svg",
@@ -117,7 +117,7 @@
             "textG": 0,
             "textR": 0
           },
-          "comment": "This variable is needed because the time will shift before next onScenePreEvents call.",
+          "comment": "Step time counters.",
           "comment2": ""
         },
         {
@@ -126,11 +126,12 @@
           "actions": [
             {
               "type": {
-                "value": "SetSceneVariableAsBoolean"
+                "value": "ModVarScene"
               },
               "parameters": [
-                "__CameraShake_NeedToRevertCameraChanges",
-                "False"
+                "__CameraShake.Time",
+                "+",
+                "TimeDelta()"
               ]
             }
           ]
@@ -151,19 +152,10 @@
           "actions": [
             {
               "type": {
-                "value": "SetSceneVariableAsBoolean"
-              },
-              "parameters": [
-                "__CameraShake_NeedToRevertCameraChanges",
-                "True"
-              ]
-            },
-            {
-              "type": {
                 "value": "ModVarScene"
               },
               "parameters": [
-                "__CameraShake_EaseFactor",
+                "__CameraShake.EaseFactor",
                 "=",
                 "1"
               ]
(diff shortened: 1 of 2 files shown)


tristanbob:
This looks interesting and it might be able to replace all the use-cases of my version.  

Both versions seem to include:
- Variable duration
- Variable frequency (I call this "time between shakes" because I feel it is more use friendly, despite 4ian's suggestion to use frequency)
- Variable amplitude
- Affect position (X and/or Y), rotation, and zoom
- Include helper condition ("Is shaking")
 
 Your version has some very nice benefits:
 - Works on multiple layers
 - Noise looks more natural than random
 
 Your version has some minor negatives:
 - More steps to use. Can this be improved?
 - Can it perform a "shake forever" functionality?
 - Forced to include a full copy of the noise extension. We need extension dependencies, similar to how behavior dependencies work.
 - I think you might need a unique timer per layer.  It looks like a single timer is used.
 
@D8H I would be probably be convinced if we can implement everything in my screenshake example (source code included):
https://victrisgames.itch.io/gdevelop-camera-shake-example

D8H:
> More steps to use. Can this be improved?

It depends what the user do. With the default configuration only 1 action have to be called and it only have 3 parameters.

> Can it perform a "shake forever" functionality?

It can. There is a start and a stop action.

> Forced to include a full copy of the noise extension. We need extension dependencies, similar to how behavior dependencies work.

It's not really an issue. The user doesn't see it and it's relatively easy to copy/paste it with a text editor if an update is needed.

> I think you might need a unique timer per layer.  It looks like a single timer is used.

To shake layers with different delays? What use case do you see for it?

> Missing a "single shake" effect that allows a one-time deterministic movement (for a knockback (or similar) effect)

It's not a shake. It should be another extension.

> @D8H I would be probably be convinced if we can implement everything in my screenshake example (source code included): https://victrisgames.itch.io/gdevelop-camera-shake-example

There it is: https://www.dropbox.com/s/vkm5lmh20ipbayp/CameraShakeExample_Tristan.zip?dl=1
I fixed a few bugs in the compatibility action.

arthuro555:
Hello @D8H 👋 !

HelperWesley:
I love that this can cover multiple layers, that's something that's been missed in the original camera shake extension. 👍

I will say though...
The philosophy for extensions that I would like to see implemented is "fewer more powerful extensions". The way that the current object and camera shake extensions work allows for a huge variety of options:

-Single knock back/hit/bounce effects.
-Bouncing screen elements up and down.
-Screen shake.
-Stretch and squash the screen.
and the list goes on.

The only thing it was missing, in my opinion, was the the multi layer functionality. 🤔

If we lose the ability to do directional single shakes, the entire extension will be lesser for it. Adding complexity to the process, by requiring multiple extensions or actions to do similar things, while not adding a lot of benefit to the user. 🤔

Also, having to set your settings at the beginning of a scene instead of within the action itself is really restrictive, can't easily set different shakes for different actions, and adds a lot more clutter to the event sheet. (This part might just be because it's in development, but a finished product should look similar to the current shake action's layout)

I think the noise and the easing are really nice, the example project shows what would be really good for an "earthquake" type shake. 🤔

I would just rather not see this replace the current extension if it's going to exclude some of the existing features, like being able to use it for single bounces/shakes, and simplicity of the original.

D8H:
> I love that this can cover multiple layers, that's something that's been missed in the original camera shake extension. 👍
> 
> I will say though... The philosophy for extensions that I would like to see implemented is "fewer more powerful extensions". The way that the current object and camera shake extensions work allows for a huge variety of options:

It's nice when an extension can be used to do more than what is was originally designed for but only if it doesn't make it more complicated than 2 dedicated extensions.

> -Single knock back/hit/bounce effects. -Bouncing screen elements up and down. -Stretch and squash the screen.

These one should probably use camera tweens on game without scrolling or a dedicated extension because it's the kind of thing where easing functions is very important and the screen-shake extension can't give these features.

> Also, having to set your settings at the beginning of a scene instead of within the action itself is really restrictive, can't easily set different shakes for different actions, and adds a lot more clutter to the event sheet. (This part might just be because it's in development, but a finished product should look similar to the current shake action's layout)

Extensions have properties. It's not yet possible to define one with event-based extensions but it will come and any extension that puts everything in one action will have to do a breaking change to use properties.

It allows cleaner events because if you only change the amplitude of the shake, you don't have to put back the whole configuration.

> I would just rather not see this replace the current extension if it's going to exclude some of the existing features, like being able to use it for single bounces/shakes, and simplicity of the original.

We will have a lot of time to discuss it because this extension needs a very bad hack to work so even if the other extension bugs with scrolling games, it won't replace it soon.

Bouh (approved):
- There is a `test` scene variable useless in `Shake camera` and in `Shake camera (deprecated)`
- `OnSceneLoaded` have the same action as in `ShakeCamera` but with different values.


Ps: The duration in `StartShaking` make the camera shakes by default for 1234567890sc, ~39 years, lol It's just to say that it's not infinite ^^

D8H:
> * There is a `test` scene variable useless in `Shake camera` and in `Shake camera (deprecated)`
> 
>     * `OnSceneLoaded` have the same actions as in `ShakeCamera` but with different values.

The condition on this event is "never". It was a workaround for a bug that didn't generate the code of functions that was only called from life-cycle function, but it's no longer needed. I removed them.

HelperWesley:
I like the shake effect this produces.

I don't like how it's presented in the events.😅

There is a lot of clutter. A lot of these things, like the default settings and start/stop shaking, could be combined in to single events to reduce clutter and simplify the work flow.

Just being able to start and stop the shake during a specific time requires 3 actions, where the existing extension has the duration build in.
image

If someone wanted two entirely different forms of camera shake for some reason, they would need all of this to accomplished 2 different kinds of shakes within the same project. Where previously they would need 2 actions.
image

These are a lot of actions that each do only one thing.

D8H:
> I like the shake effect this produces.
> 
> I don't like how it's presented in the events.😅
> 
> There is a lot of clutter. A lot of these things, like the default settings and start/stop shaking, could be combined in to single events to reduce clutter and simplify the work flow.
> 
> Just being able to start and stop the shake during a specific time requires 3 actions, where the existing extension has the duration built in. image

There is already an action for it.
image


> If someone wanted two entirely different forms of camera shake for some reason, they would need all of this to accomplished 2 different kinds of shakes within the same project. Where previously they would need 2 actions. image
> 
> These are a lot of actions that each do only one thing.

There are default values, it's very likely users need rotation and zoom. With this version, you don't need to set them, when the previous version the shake action had 9 parameters (most of them useless 90% of the time).
It also offer more features, for instance you could put a sinusoid in the amplitude to add some effects. Every properties can be eased independently during a shake.

HelperWesley:
image
So there are even more actions that do a single thing.

Could this be a toggle on the other "start shaking" action? Instead of further cluttering the action list. 🤔

image

Being "useless" is a bit subjective, and default values being built in is great, but that still leaves us with a big list of actions that do single things. Increasing clutter in the event sheet and action list.

D8H:
>> image
> 
> So there are even more actions that do a single thing.
> 
> Could this be a toggle on the other "start shaking" action? Instead of further cluttering the action list. 🤔

In this case, when an hypothetic  "permanent" toggle is enabled, the "duration" and the "easing duration to stop" wouldn't make sense.

> image
> 
> Being "useless" is a bit subjective, and default values being built in is great, but that still leaves us with a big list of actions that do single things. Increasing clutter in the event sheet and action list.

Both way works, but I have the impression that configurations are usually done with independent actions in the engine. For instance, the jump of a platformer character can be configured with 3 actions for: the jump speed, the gravity, the maximum and falling speed.
- When you only want to change the jump speed, It avoids having to set the other 2 parameters.
- It also avoids to break the compatibility when properties are added or removed.

If I had the time I would make a pass on the reviewed extensions to comply this, because all these actions with a lot of parameters are not practical to use.

HelperWesley:
Those aren't really comparable though I think. 🤔

I set my default controls for my platformer based on how I want it to feel, and then tweak individual variables to achieve some temporary goal. A run, a high jump, some kind of restriction from a thing the player is standing it, etc. But I'll generally want all other variables to stay as they were, and I'll want to go back to that preset after the temporary effect has ended. It's relatively constant.

With camera shakes I could want dramatically different shakes depending on the situation I'm in. If you're making a platformer game and the character is riding in a shaking elevator you could would want something with entirely different settings compared to an earthquake or general rumble of some kind. They probably aren't going back to a default setting. Shakes could be dramatically different depending on the circumstances.

D8H:
> Having a single action that contains all of your default settings, and then having another single action with toggles that could be used to trigger the starting/stopping of the shaking, would minimize the number of actions cluttering the action list and event sheet.
> 
> The fact that I didn't originally know(After looking) that there was a third start/stop action, shows that the clutter can hide useful tools that a user would want to use if it were presented in a more minimalist and organized way.

I think the issue here is that my examples are not useful to learn the extension, only to test it.

The 3 actions are pretty clear in my opinion.


> Having a single action that contains all of your default settings, and then having another single action with toggles that could be used to trigger the starting/stopping of the shaking, would minimize the number of actions cluttering the action list and event sheet.
> 
> The fact that I didn't originally know(After looking) that there was a third start/stop action, shows that the clutter can hide useful tools that a user would want to use if it were presented in a more minimalist and organized way.

I put a green cross on actions that will satisfy 90% of the use cases.

If the layer selection was in the same action that other properties, users would need to copy-paste 4 values for each layers.
Merging the 4 properties would force users to set something for rotation and zoom to save only one action.

So, I think that it wouldn't make it easier to understand and would remove flexibility (like dynamically changing the properties value while the screen is shaking).

image

HelperWesley:
I think you just highlighted my point. 🤔

4 out of 12 of those actions are what you think people are going to use, and the rest are cluttering the action list.

Would combining them, and having blank(Or toggled) parameters pick the previously set default settings, be possible? That way you wouldn't need to copy the variables each time you want to change the shake settings. 🤷‍♂️

D8H:
> I think you just highlighted my point. 🤔
> 
> 4 out of 12 of those actions are what you think people are going to use, and the rest are cluttering the action list.
> 
> Would combining them, and having blank(Or toggled) parameters pick the previously set default settings, be possible? That way you wouldn't need to copy the variables each time you want to change the shake settings. 🤷‍♂️

I guess we don't understand each other at all 😅 

These 2 event are enough to make a screen shake:

image

HelperWesley:
I do not like this much clutter in the event sheet or action list, I personally find it overwhelming and distracting, but the extension functionally works and does what it sets out to do.

I'm not a UX designer so I'll stop arguing that point. I think you can merge this extension, but we'll need documentation showing the two events you show in the above comment making it very clear that those are the actions you need and the rest are only needed if you want a more specific kind of screen shake.

PR #390, opened 2022-04-05
Fix typo (trickness -> thickness)

Bouh:
Fix https://github.com/GDevelopApp/GDevelop-extensions/issues/388


tristanbob (approved):
This matches what I was expecting:

- 2 uses inside descriptions
- 6 uses inside inline javascript

Looks good to me.

D8H (approved):

PR #395, opened 2022-04-15
Update CameraZoom so negative speeds have exact opposite effect as positive speed

tristanbob:
The current CameraZoom extension requires users to follow this process:

To zoom IN:  Select a speed greater than 1.
To zoom OUT:  Select a speed between 0 and 1.  To match the "Zoom In" speed, use "1 / ZoomInSpeed"

This PR allows users to simply use negative numbers to perform the opposite action of the positive number.  This is how most value changes in GDevelop work (position, forces, rotation, etc).  Now users can use this process:

To zoom IN:  Select a speed greater than 1.  (No change from previous version)
To zoom OUT:  Select a speed less than -1.  To match the "Zoom In" speed, use "-ZoomInSpeed"

Special cases:

0 < Speed < 1:  Makes the camera zoom OUT.  (This behaves as it did before the change, so no impact on existing users.)
-1 < Speed < 0: Makes the camera zoom IN.

I haven't included detailed info on the valid inputs yet, because I wanted to get feedback from the team first.

### DIFF
--- a/Extensions/CameraZoom.json
+++ b/Extensions/CameraZoom.json
@@ -27,127 +27,360 @@
   "eventsFunctions": [
     {
       "description": "Change the camera zoom at a given speed (in factor per second).",
-      "fullName": "Zoom camera with speed",
+      "fullName": "Zoom camera with speed (symmetrical)",
       "functionType": "Action",
       "group": "",
       "name": "ZoomWithSpeed",
       "private": false,
-      "sentence": "Zoom camera with speed: _PARAM1_ (layer: _PARAM2_, camera: _PARAM3_)",
+      "sentence": "Zoom camera (symmetrical) with speed: _PARAM1_ (layer: _PARAM2_, camera: _PARAM3_)",
       "events": [
         {
+          "colorB": 228,
+          "colorG": 176,
+          "colorR": 74,
+          "creationTime": 0,
           "disabled": false,
           "folded": false,
-          "type": "BuiltinCommonInstructions::Comment",
-          "color": {
-            "b": 109,
-            "g": 230,
-            "r": 255,
-            "textB": 0,
-            "textG": 0,
-            "textR": 0
-          },
-          "comment": "Convert negative numbers to the inverse positive value to makes the speed symmetric.  A positive and negative value perform the exact opposite effect. ",
-          "comment2": ""
-        },
-        {
-          "disabled": false,
-          "folded": false,
-          "type": "BuiltinCommonInstructions::Standard",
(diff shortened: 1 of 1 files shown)


D8H:
> This PR allows users to simply use negative numbers to perform the opposite action of the positive number. This is how most value changes in GDevelop work (position, forces, rotation, etc).

It works like the "Change camera zoom" action does. Negative values has no effect because it's likely a mistake. A negative zoom factor would in theory do a 180° rotation and go from tiny around 0 to huge around -infinity, but I don't see any use for it.

A negative factor speed doesn't really make sens because it would be like a 180° rotation that happens so fast that it would be like a superposition of 2 images if it were physically possible. 

I think the extension should stays the way it is, because giving sens to something that has none for convenience will likely defeat its purpose. For instance, users could make a setting go from -2 to 2 with 0.1 steps and don't understand what happens between -1 and 1. It's easier to explain or understand that negative values don't work.

tristanbob:
I made a game to compare how the two versions act with various zoom speeds (controlled by a slider).

https://liluo.io/instant-builds/707ec7e3-1b82-46f5-aef3-86a408e29319

I was able to make [-1 to 1] move in the right directions, but the speed changes are not smooth.  I'm wondering if we should not perform changes to speed in this range (and inform users about this).

In summary, I still think that symmetrical speed (where positive and negative values perform the exact opposite effect)  provides a much better user experience. 

## Project files

[ZoomSpeedSymmetrical.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8500934/ZoomSpeedSymmetrical.zip)

## Video

https://user-images.githubusercontent.com/8879811/163698898-1903f585-6884-45f6-b925-0a68f31c0584.mp4

D8H:
> I made a game to compare how the two versions act with various zoom speeds (controlled by a slider).
> 
> https://liluo.io/instant-builds/707ec7e3-1b82-46f5-aef3-86a408e29319
> 
> I was able to make [-1 to 1] move in the right directions, but the speed changes are not smooth. I'm wondering if we should not perform changes to speed in this range (and inform users about this).
> 
> In summary, I still think that symmetrical speed (where positive and negative values perform the exact opposite effect) provides a much better user experience.

With this action 0.5 and 2 act the same way. It will be confusing.

tristanbob:
I added one last commit to save what I had done.  Based on your feedback, will close this PR.

However, to continue my goal of improving the user experience, I'm going to make a new PR that adds an action that users can use directly "Change camera zoom with mouse wheel".

PR #397, opened 2022-04-18
Updated Checkbox extension (based on template from DraggableSlider)

tristanbob:
## Summary

Started with the DraggableSlider extension and then manually ported the Checkbox logic.

Improvements:
- Checkbox only toggles when a click started inside the checkbox
- Halo is now animated
- Better performance because drawing now only happens after a change
- Added more functions to control the look and feel of the checkbox
- Checkbox can be rotated (but... why would someone do this?)

To do:  
- Review descriptions and names for consistency and clarity (I would love help with this)
- Determine the proper version number.  I proposed 1.0.0 because this matches the DraggableSlider.  However, there are no breaking changes, based on my testing.

## Playable game

https://liluo.io/games/0e1bed4e-3a5c-4f38-bd78-977de21aad67

## Project files

[CheckboxExtension_SourceFiles.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8503406/CheckboxExtension_SourceFiles.zip)

## Video

https://user-images.githubusercontent.com/8879811/163753890-575e7a08-0a46-4ab0-9e10-048f7c1b5628.mp4


Bouh:
As you base this new extension on DraggableSlider I have to tell you [I fixed two behavior functions:](https://github.com/GDevelopApp/GDevelop-extensions/pull/399)
- SetActiveTrackColor
- SetInactiveTrackColor

The error was a mix of parameters, the sign `=` is replaced by a string, which shouldn't be possible.
image

tristanbob:
> As you base this new extension on DraggableSlider I have to tell you [I fixed two behavior functions:](https://github.com/GDevelopApp/GDevelop-extensions/pull/399)

Thanks for asking that, but no, I removed those functions as they are not needed for the checkbox.

Bouh:
- [ ] While using the Hot Reload from behavior properties, I change properties, but I've to hover the checkbox again to redraw it with new styles.
- [ ] By default the engine return False for a condition, so as you made for `IsEnabled` you can just use an event in `IsChecked` to return true when needed.

tristanbob:
> * [ ]  While using the Hot Reload from behavior properties, I change properties, but I've to hover the checkbox again to redraw it with new styles.

Any ideas on how to detect when a hot reload has occurred? (Which is used only during development, so it won't affect exported games.)  If there was a condition for this, I could trigger the "NeedsRedraw" flag.

> * [ ]  By default the engine return False for a condition, so as you made for `IsEnabled` you can just use an event in `IsChecked` to return true when needed.

Thanks, I can change that.

4ian:
> Any ideas on how to detect when a hot reload has occurred? (Which is used only during development, so it won't affect exported games.) If there was a condition for this, I could trigger the "NeedsRedraw" flag.

Good question, there is nothing for now but we could have a "Just hot reloaded" condition (just added: https://trello.com/c/uXHkrXyv/520-add-a-condition-to-check-if-a-hot-reload-just-happened)

tristanbob:
Any other comments on this?  I'd like to merge when ready.

D8H:
I didn't make any change to the extension, I just modified the example to be able to check when redraw happens (the example was changing the properties every frames so it was always redrawing).
https://www.dropbox.com/s/xykur2onwounnwi/CheckboxExtension_REVIEW.zip?dl=1

I read the code quickly, it seems good. I have some suggestion about wording:

I think that the pixel unit is a given, it could be removed everywhere.

Action full names should not start with Change unless it's important for understanding, for instance:
* from `Change border thickness of checkmark`
* to `Border thickness`?

SetChecked
* from `Add checkmark to _PARAM0_? _PARAM2_`
* to `Check _PARAM0_: _PARAM2_ `?

ToogleChecked
* from `Change _PARAM0_ to opposite state (checked/unchecked)`
* to `Toogle checkmark on _PARAM0_`?

SetHaloOpacityPressed and SetHaloOpacityHover
* in the description: "when" instead of parenthesis?

About "primary" and "secondary" colors, it's not easy to remember what they does without reading the description. What about something like "box" and "background" colors?

> Determine the proper version number. I proposed 1.0.0 because this matches the DraggableSlider. However, there are no breaking changes, based on my testing.

Keeping the same version is useful only when the components only works with eachother at the same version. Here, the 2 extensions are independent. Changing the major will scare users for nothing.

tristanbob:
> I didn't make any change to the extension, I just modified the example to be able to check when redraw happens (the example was changing the properties every frames so it was always redrawing). https://www.dropbox.com/s/xykur2onwounnwi/CheckboxExtension_REVIEW.zip?dl=1

Thank you for testing this aspect of the extension!

> I think that the pixel unit is a given, it could be removed everywhere.

I prefer to leave it for now. 

> Action full names should not start with Change unless it's important for understanding, for instance:
> 
> * from `Change border thickness of checkmark`
> * to `Border thickness`?

It looks like many actions throughout the engine use the longer text format.  I prefer to leave it for now.

> SetChecked
> 
> * from `Add checkmark to _PARAM0_? _PARAM2_`
> * to `Check _PARAM0_: _PARAM2_ `?

"Check" could be confused with another meaning, like "test the checkbox".  I prefer to leave it.

> ToogleChecked
> 
> * from `Change _PARAM0_ to opposite state (checked/unchecked)`
> * to `Toogle checkmark on _PARAM0_`?

Good idea, I have made this change.

> SetHaloOpacityPressed and SetHaloOpacityHover
> 
> * in the description: "when" instead of parenthesis?

Good idea, I have made this change.

> About "primary" and "secondary" colors, it's not easy to remember what they does without reading the description. What about something like "box" and "background" colors (as this last color is transparent in Material-UI)?

I agree this can be confusing, but those names were already used and I prefer to leave it for now. 

> > Determine the proper version number. I proposed 1.0.0 because this matches the DraggableSlider. However, there are no breaking changes, based on my testing.
> 
> Keeping the same version is useful only when the components only works with eachother at the same version. Here, the 2 extensions are independent. Changing the major will scare users for nothing.

Thanks, I agree with you.  Swapping this to a minor version change. (now version 0.1.0)

Lastly, I don't want to increase the scope of this PR any further.  This is a good change that "modernizes" the checkbox and provides more consistency with other UI elements.  Let's get this merged soon so people can start using it.  :)

D8H:
> > Action full names should not start with Change unless it's important for understanding, for instance:
> > 
> > * from `Change border thickness of checkmark`
> > * to `Border thickness`?
> 
> It looks like many actions throughout the engine use the longer text format. I prefer to leave it for now.

There is indeed a lot of action that start with a verb and a few that start with "change".

image

But, the built-in extension property action doesn't start with "Change". I guess it would look silly to have "change" repeated 10 times here.

image

`Change border thickness of checkmark` is too long, at least "of checkmark" should be dropped.

D8H:
> @D8H How about this compromise? I removed the "change ... to" wording (which has little value) and left the "checkbox" and "checkmark" words because they provide additional useful context.

I don't understand. I have the impression that there are 2 trickness properties one for the border and another one for the checkmark. If it is the case "Change border thickness of checkmark", sounds very confusing to me. Is it the border or the checkmark?

tristanbob:
"Border thickness" is around the square checkbox.

"Line Width" is the checkmark itself. 

The actions use the correct names above.

D8H:
Ok, it's clearer with"checkbox".
I think that "of checkbox" in the full name is redundant with the icon and the full name is usually very short (it helps a lot when scanning the list).

I asked to remove "Change" from the full name but not from the sentences. It should even be "Change the" in sentences.

D8H:
I merged the main into this branch to move the extension file in the review folder. After a fetch you should be able to continue working on this extension.

tristanbob:
> I asked to remove "Change" from the full name but not from the sentences. It should even be "Change the" in sentences and descriptions.

I have added the "Change to" on action sentences.
I also added "Return the" on expressions sentences.

## Playable game

https://liluo.io/victrisgames/extension--checkbox

D8H:
> > SetChecked
> > 
> > * from `Add checkmark to _PARAM0_? _PARAM2_`
> > * to `Check _PARAM0_: _PARAM2_ `?
> 
> "Check" could be confused with another meaning, like "test the checkbox". I prefer to leave it.

Ok, but the "?" should be a ":"

tristanbob:
@D8H how about now?  :)

D8H (approved):

PR #416, opened 2022-04-23
Force the pathfinder to NOT use diagonals

tristanbob:


tristanbob:
This closed #504 (sorry for the late update)

PR #422, opened 2022-04-23
Added ZoomWithMouseWheel and ZoomWithKeypress actions

tristanbob:
This will make it super simple to add user-controlled zoom to games with a single action.

@D8H this should work but I'm getting these errors.  Is there some small mistake?

## Project file
[ZoomWithMouseWheel.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8547684/ZoomWithMouseWheel.zip)

image

### DIFF
--- a/Extensions/CameraZoom.json
+++ b/Extensions/CameraZoom.json
@@ -1,14 +1,14 @@
 {
   "author": "",
   "category": "",
-  "description": "## Description\n\nThis extension allows to zoom a camera on a layer at a given speed. The zoom speed is configurable and allows a constant as well as a variable zoom speed. An anchor point can be defined, for instance, to keep what is under the cursor at the same place on screen.\n\n### Actions\n\n- Change camera zoom with mouse wheel\n- Zoom at a given speed\n- Change the zoom with an anchor point\n- Zoom at a given speed and with an anchor point\n",
+  "description": "Change camera zoom using actions or let users change zoom by scrolling the mouse wheel.\n\nZoom speed is configurable and can use constant or variable speed.  An anchor point can be defined to keep what is under the cursor at the same place on screen.\n\n### Actions\n\n- Change camera zoom with mouse wheel\n- Zoom at a given speed\n- Change the zoom with an anchor point\n- Zoom at a given speed and with an anchor point\n",
   "extensionNamespace": "",
   "fullName": "Camera Zoom",
   "helpPath": "",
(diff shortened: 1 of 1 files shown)


D8H:
If you enable the 2 events, but put a inverted Always, it will work.
There is issues with code generator. It doesn't detect function usage from life-cycle function but I thought it worked for function to function dependencies.

tristanbob:
> If you enable the 2 events, but put a inverted Always, it will work. There are issues with code generator. It doesn't detect function usage from life-cycle function but I thought it worked for function to function dependencies.

I followed your instructions and it worked.  

How should we try to fix this in the engine?  Do you want me to file an issue?

tristanbob:
> > If you enable the 2 events, but put a inverted Always, it will work. There are issues with code generator. It doesn't detect function usage from life-cycle function but I thought it worked for function to function dependencies.
> 
> I followed your instructions and it worked.
> 
> How should we fix this in the engine? Do you want me to file an issue?

@D8H is this something that should be tracked as an issue?

D8H:
> > > If you enable the 2 events, but put a inverted Always, it will work. There are issues with code generator. It doesn't detect function usage from life-cycle function but I thought it worked for function to function dependencies.
> > 
> > 
> > I followed your instructions and it worked.
> > How should we fix this in the engine? Do you want me to file an issue?
> 
> @D8H is this something that should be tracked as an issue?

Yes, it's related to this issue:
* https://github.com/4ian/GDevelop/issues/2495

But in this case, it wasn't a life-cycle function which is strange.

tristanbob:
I just tested this and it works!  @D8H thank you for your work on this: https://github.com/4ian/GDevelop/pull/3908)

I updated some of the descriptions and names in the original extension to make them easier to understand.

## Project files

[ZoomWithMouseWheel-06-05-22.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8841577/ZoomWithMouseWheel-06-05-22.zip)

tristanbob:
> Giving something easy to use to zoom with the mouse wheel is a good idea. I wonder we should rather do an example than an extension, at least as the 1st step.

I strongly feel that this action `Change zoom with mouse wheel` will be very useful for users and widely used.  Part of the reason I made this action is because the zoom factor is not intuitive: Users must use inverse values to move in opposite directions.  For instance, "200" and "1/200".  This alone is going to trip up users, and so the action will save them time and frustration.

> The zoom with speed should only be used over time. For a instant zoom changes, the "change zoom" action must be used. It's a bit like "add force" and "set position" actions. Because of this, the example doesn't do the same at 20 fps and 60 fps.

I'm confused what you wrote here.  I'll paste the names of the functions that I proposed.  Did I misunderstand how one of the functions operated?

```
Zoom camera over time
- Change the camera zoom at a given speed (in factor per second).
- Change camera zoom at speed: _PARAM1_ (layer: _PARAM2_, camera: _PARAM3_)

Change camera zoom and use an anchor point
- Change camera zoom (instantly) and keep an anchor point fixed on screen (instead of the center).
- Set camera zoom to: _PARAM1_ with anchor at: _PARAM4_; _PARAM5_ (layer: _PARAM2_, camera: _PARAM3_)

Zoom camera over time and use an anchor point
- Change the camera zoom at a given speed (in factor per second) and keep an anchor point fixed on screen (instead of the center).
- Zoom camera at speed: _PARAM1_ and anchor at: _PARAM4_; _PARAM5_ (layer: _PARAM2_, camera: _PARAM3_)
```

> It could be interesting to compare instant and progressive zoom and explains which action to use. There might be some research to do about reacting to the wheel speed.

The speed a mouse wheel turns might affect the zoom rate, but zooming is usually a non-precise action.  If a game needs precise zoom with mouse wheel, the dev can offer a draggable slider that modifies the rate of change in camera zoom.

D8H:
> The zoom with speed should only be used over time. For a instant zoom changes, the "change zoom" action must be used. It's a bit like "add force" and "set position" actions. Because of this, the example doesn't do the same at 20 fps and 60 fps.
> 
> I'm confused what you wrote here. I'll paste the names of the functions that I proposed. Did I misunderstand how one of the functions operated?

I was referring to the built-in action:

image

In this action, the parameter is a factor.
- If the value `2` is set and it runs once, it will zoom by a factor of 2.
- It's the equivalent of "Set position" in a position/speed analogy.

The "Zoom camera over time" parameter is a factor per second.
- If the value `2` is set and it runs during 1 second, it will zoom by a factor of 2.
- If the value `2` is set and it runs once, it will zoom by a factor of 2^TimeDelta.
- It's the equivalent of "Add force" in a position/speed analogy (as a force is actually a velocity in GDevelop).


> > Giving something easy to use to zoom with the mouse wheel is a good idea. I wonder we should rather do an example than an extension, at least as the 1st step.
> 
> I strongly feel that this action `Change zoom with mouse wheel` will be very useful for users and widely used. Part of the reason I made this action is because the zoom factor is not intuitive: Users must use inverse values to move in opposite directions. For instance, "200" and "1/200". This alone is going to trip up users, and so the action will save them time and frustration.

I agree zooms are not intuitive and that's why I think we should explain how to use it rather than hide the knowledge. The built-in zoom action can be useful in many ways, so I think it's important that game creators understand it.

tristanbob:
@D8H thanks for your feedback.  I think the value of my additions to this extension depends on your perspective.  Your skills enable you to make magic using advanced methods that our power users crave and love (including me).   

However, I often see people frustrated with managing camera controls so I want to remove that barrier so people can focus on making their dream game.  

@4ian can you provide some guidance on this PR? 

Playable game: 
https://liluo.io/victrisgames/camera-zoom

Project files:
[ZoomWithMouseWheel-06-10-22.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8883193/ZoomWithMouseWheel-06-10-22.zip)

D8H:
In the action for keys, why did you used a power formula?
The "Change camera zoom at speed" action already do the calculus to have a constant speed. It would be something like this:

image

D8H:
Hello Tristan, do you still want to upgrade this extension?

arthuro555:
This PR is looking dead, I'll close it, feel free to reopen if/when someone's getting back to it

PR #423, opened 2022-04-23
Added helper function for 1D noise (using 2D noise with a static value for one dimension)

tristanbob:
Project file:

[Testing-1D-Noise.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8548411/Testing-1D-Noise.zip)

### DIFF
--- a/Extensions/Noise.json
+++ b/Extensions/Noise.json
@@ -1,7 +1,7 @@
 {
   "author": "Add00",
   "category": "",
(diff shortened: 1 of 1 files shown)


Astr0Cookie:
1d noise 😱😳, is it possible to make Randomly generated worlds with this?

Bouh (approved):

PR #431, opened 2022-04-30
Merge #402 with 0.0.1 of Time formatting extension

Bouh:
Merge https://github.com/GDevelopApp/GDevelop-extensions/pull/402 with 0.0.1 of Time formatting extension

@D8H as you suggested in https://github.com/GDevelopApp/GDevelop-extensions/pull/402#issuecomment-1107455596

### DIFF
--- a/Extensions/TimeFormatter.json
+++ b/Extensions/TimeFormatter.json
@@ -1,38 +1,1330 @@
 {
   "author": "@Bouh",
-  "description": "Expressions to transform time in seconds to format like HH:MM:SS. Ideal to display timers on screen.",
+  "category": "",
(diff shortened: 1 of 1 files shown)


Bouh:
@D8H as you suggested in https://github.com/GDevelopApp/GDevelop-extensions/pull/402#issuecomment-1107455596

D8H:
It would probably better to wrap `DateTimeFormat`. It is much more flexible (localization and verbosity options).
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

The extension should probably allow to format any date and not just current one.

arthuro555:
This PR is looking dead, I'll close it, feel free to reopen if someone's getting back to it

PR #437, opened 2022-05-12
Update gamepad extension 

Bouh:
- Add "Any" to the parameters of the direction for the condition Stick pushed in a direction.
- Change wording to PascalCase for inputs name  (only cosmetic).


Bouh:
I need you @Silver-Streak & @tristanbob to double check, (I checked on my side)
Tell me if you get a regression with this new version 0.3.0 in an existing project. 
See the changelog in my first post.

tristanbob:
I tested this on an old project and a new one, and it worked great.  

image

tristanbob (approved):
LGTM!

Silver-Streak:
I use the controller type detection in my main game to determine whether it displays Xbox or PS4 button prompts. I'll test with this and report back if anything behaves oddly.

PR #446, opened 2022-06-04
Add resource loading tools extension

arthuro555:
Example file: [load-from-url.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8836589/load-from-url.zip)
 
Adds a few actions that have something to do with loading resources.

Points of importance for the review:
 - A private JS variable is being used in the action for audio resources, which is not acceptable. My implementation of the audio one is generally meh in my opinion, I think we probably would remove it unless you think it is essential to keep. Other actions are good.
 - The name is kind of bad.
 - The extension is maybe not focused enough.
 - It uses parameter types not officially supported by events-based extensions. They work just fine without hacks in JavaScript though, only real downside is that if you select another type for those fields while editing the extension you will be stuck unable to go back unless you modify the JSON directly or redownload the extension. I think that this is fine but you may think otherwise.

### DIFF
--- a/Extensions/ResourceLoadingTools.json
+++ b/Extensions/ResourceLoadingTools.json
@@ -1,0 +1,209 @@
+{
+  "author": "",
+  "category": "General",
+  "description": "Adds a set of actions to work with the loading of resources: Downloading a texture at runtime, loading a JSON into a scene variable, and more!",
+  "extensionNamespace": "",
+  "fullName": "Resources loading tools",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


Bouh:
Is the video can be supported?
I made a tiny tweak to be able to [show to entropy](https://forum.gdevelop.io/t/video-object-questions/40137/2) how to do, by changing the object type in **LoadURLIntoSprite** function, it's maybe optimal but it works.

4ian:
Thanks for getting this moving :) 

> They work just fine without hacks in JavaScript though, only real downside is that if you select another type for those fields while editing the extension you will be stuck unable to go back unless you modify the JSON directly or redownload the extension. I think that this is fine but you may think otherwise.

I think that's fine!

> A private JS variable is being used in the action for audio resources, which is not acceptable. My implementation of the audio one is generally meh in my opinion, I think we probably would remove it unless you think it is essential to keep

I would remove this yes, as this can be brittle.

The extension seems not to provide though:
- error handling (what happens if something can't be loaded?)
- conditions to check/know when something has finished loading or errored.

D8H:
The extension is not in the right folder.

arthuro555:
I separated it into two extensions: #626 and #627

PR #455, opened 2022-06-11
[CursorMovement] Fix conditions were always returning true when called a 2nd time.

D8H:
## Change
* The cursor position is saved in `onScenePostEvents` instead of the condition itself.

## Known issues
* The condition won't work if the default layer is moving.
  This might need some JavaScript to get the cursor position in the window referential so I prefer to do it in another PR to merge this quickly.

## Demo
* https://www.dropbox.com/s/wgwkzboyinbu3vi/StillCursorTest.zip?dl=1

Thank you Ash for reporting the issue.

### DIFF
--- a/Extensions/CursorMovement.json
+++ b/Extensions/CursorMovement.json
@@ -1,14 +1,19 @@
 {
   "author": "D8H",
+  "category": "",
   "description": "Provides two conditions:\n* The cursor is moving\n* The cursor has stayed still for a given duration",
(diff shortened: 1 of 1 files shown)


tristanbob:
I reviewed the extension and tested your game with a couple of extra events.  It appears to work correctly!

https://user-images.githubusercontent.com/8879811/173203524-3efc7f12-c61b-4b8a-9cc4-37f37ac18b5c.mp4

PR #457, opened 2022-06-11
Made FireBullet extension awesome!

tristanbob:
- Added the ability to shoot multiple bullets at once
- Cooldown: Time between shots (seconds)
- Bullet Quantity: Number of bullets shot at once
- Angle Variance: Each bullet angle will be randomized in this range (degrees)
- Firing Arc: Range of angles that bullets will shoot.  Bullets are evenly spread within this range.
- Changed "fire at position" to simply use "fire bullet" with a calculated angle

## Project files
[Multiple Shot Fire Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8884844/Multiple.Shot.Fire.Extension.zip)

## Video
https://user-images.githubusercontent.com/8879811/173207256-9b4d0949-6b1b-41c6-9c13-5680d7fec16f.mp4

### DIFF
--- a/Extensions/FireBullet.json
+++ b/Extensions/FireBullet.json
@@ -42,41 +42,82 @@
           "group": "",
           "name": "FireTowardPosition",
           "private": false,
-          "sentence": "Fire _PARAM4_ from _PARAM0_, at position _PARAM2_;_PARAM3_, toward position _PARAM5_;_PARAM6_ with speed _PARAM7_ px/s",
+          "sentence": "Fire _PARAM4_ from _PARAM0_, at position _PARAM2_;_PARAM3_, toward position _PARAM5_;_PARAM6_ with speed _PARAM7_ px/s. Rotate bullet: _PARAM8_.",
           "events": [
             {
-              "type": "BuiltinCommonInstructions::Comment",
-              "color": {
-                "b": 109,
-                "g": 230,
-                "r": 255,
-                "textB": 0,
-                "textG": 0,
-                "textR": 0
-              },
-              "comment": "Calculate angle to fire bullets",
-              "comment2": ""
-            },
-            {
-              "type": "BuiltinCommonInstructions::Standard",
-              "conditions": [],
-              "actions": [
+              "colorB": 228,
+              "colorG": 176,
+              "colorR": 74,
+              "creationTime": 0,
+              "name": "Calculating the trajectory angle and use \"fire\" action",
+              "source": "",
+              "type": "BuiltinCommonInstructions::Group",
+              "events": [
                 {
-                  "type": {
-                    "value": "FireBullet::FireBullet::Fire"
-                  },
-                  "parameters": [
-                    "Object",
-                    "Behavior",
-                    "GetArgumentAsNumber(\"XPosition\")",
-                    "GetArgumentAsNumber(\"YPosition\")",
-                    "Bullet",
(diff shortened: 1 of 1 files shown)


Bouh (approved):
👍

D8H:
Some suggestions
* What happens when FiringArc = 45° and BulletQuantity = 1?
  Does it go in the right direction?
  Does the division by 0 cause any issue?
* Rotation could be made optional to avoid blurring round projectiles
* Some games always fire when a key is pressed again. I don't like it because it leads to painful fingers but people may want it.
* Can games show the cool-down recharge?
* I think that an action that does nothing in some conditions is not intuitive. I wonder if the fire rate should be left aside. It's really easy to do with a timer.

tristanbob:
## Updated project

[Multiple Shot Fire Extension 06-14-22.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8902878/Multiple.Shot.Fire.Extension.06-14-22.zip)

## Playable game
https://liluo.io/games/5ce85053-a1d7-43e5-96da-33738912109b


> * What happens when FiringArc = 45° and BulletQuantity = 1?
>   Does it go in the right direction?

Good catch!  This causes the bullets to travel off at a 22.5 degree angle.  
I just added a special case for bullets = 1.  Let me know if there is a better way to handle this.

>   Does the division by 0 cause any issue?

Not that I could see, but it should never happen anymore.

> * Rotation could be made optional to avoid blurring round projectiles

There is only a one-time setting of the angle to match the trajectory of the bullet.

> * Some games always fire when a key is pressed again. I don't like it because it leads to painful fingers but people may want it.

This can be done manually, by setting cooldown property to "0" on button release and setting cooldown to another value when condition "HasJustFired" is true.

image

> * Can games show the cool-down recharge?

Good idea.  I added a condition "CooldownTimeLeft".

> * I think that an action that does nothing in some conditions is not intuitive. I wonder if the fire rate should be left aside. It's really easy to do with a timer.

The cooldown timer was already in the extension and it is one of the primary features of the extension.  It should stay.

tristanbob:
> > * Rotation could be made optional to avoid blurring round projectiles

I just realized what you meant by this.  In a pixel game, a circular bullet could look ugly if rotated.  
I will make the bullet rotation a boolean parameter, enabled by default.

Bouh:
Still to me 👍

D8H:
* What should happen when the firing arc is 360 and it fires 4 bullets? 2 bullets overlap and only 3 seems to be fired?

* RotateBullet should be a property. It won't change from a shoot to another.

> Time remaining before a bullet may be fired

Remaining cooldown time

> Seconds remaining before the cooldown will permit a bullet to be fired.

Return the time to wait before another bullet can be fired (in seconds).

tristanbob:
> * What should happen when the firing arc is 360 and it fires 4 bullets? 2 bullets overlap and only 3 seems to be fired?

Yes, that is what would happen.  What do you think should happen?

> * RotateBullet should be a property. It won't change from one shoot to another.

I agree. I will try to submit this change this tonight before many people use the new version.

> > Time remaining before a bullet may be fired
> Remaining cooldown time
> > Seconds remaining before the cooldown will permit a bullet to be fired.
> Return the time to wait before another bullet can be fired (in seconds).

I'll have to look at the context of these before deciding to change them.

PR #466, opened 2022-06-16
Extension update: Mouse helper

planktonfun:
* Added Smoothing animation option for the mouse cursor extension
 
 ### Game Preview
 https://liluo.io/games/c1ef6005-d51f-4c35-8f7f-6b306593c4b6
 
 ### Project Files
 [mouse-helper-update.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/8901384/mouse-helper-update.zip)

### DIFF
--- a/Extensions/MouseHelper.json
+++ b/Extensions/MouseHelper.json
@@ -1,5 +1,6 @@
 {
   "author": "Lim98SE",
+  "category": "",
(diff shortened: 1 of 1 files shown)


Bouh:
Might be a bug with the example, when I enable the smooth it works, when I disable it the smooth is still there.
@tristanbob have you faced a similar problem with the checkbox in the past?

PR #469, opened 2022-06-19
FireBullet: Change rotate bullet from a parameter to a property

tristanbob:
Update some text also.


tristanbob:
I went ahead and pushed this so people could start using the version with "rotate bullet" as a property.

D8H:
Ok, I'm sorry I didn't have time to check, but I'm sure it's good.

PR #483, opened 2022-07-02
Make the Auto-PR system more robust

arthuro555:
Made a few changes to make the auto-pr system more robust:

- Many edge cases are now handled when submitting extensions, each with a custom message sent back by the bot to ensure the submitter knows exactly what's going on
- Updated to the latest versions of the actions
- Only check the extension that has been submitted, not every other, for faster runs
- Extract the zip programmatically with a lot of restrictions to prevent using weird zip path shenanigans to put files at places they are not supposed to
- Unit tests for all of this 😎 

You can check out the bot's new error messages here: https://github.com/arthuro555/GDevelop-extensions/issues/20
(The actions are pushed on my fork, you can generally test out the changes there)

Possible follow-up work:
- Make the auto-pr better at handling updates:
  - Changing the tag if the extension already exists
  - Extracting to the right tier folder if the extension already exists
- Reuse the new scripts to improve the PR update command
- Adding an auto pr type to move an extension from community to reviewed tier


scripts/check-single-extension.js
```diff
@@ -0,0 +1,59 @@
...
+    ),
+  ]);
+
+  if (!community && !reviewed) return { code: 'not-found' };
+  if (community && reviewed) return { code: 'duplicated' };
+
+  //@ts-ignore We know this cannot be null thanks to the checks done just before
+  /** @type {string} */ const file = (
+    community ? community : reviewed
+  ).toString();
```
4ian, line 35: What about:
```suggestion
  /** @type {string} */ const file = community ? community : reviewed;
```

.github/workflows/auto-pr.yml
```diff
@@ -10,71 +10,130 @@ jobs:
...
 
-      - name: Notify build errors
-        if: ${{ !contains(env.BUILD_LOGS, 'successfully updated') }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: |
-          gh issue comment ${{ github.event.issue.number }} --body "Hi @${{ github.actor }}!
 
-          Thanks for your submission! Unfortunately, our automated test system found a few issues with your extension:
+            const { error, extensionName } = await extractExtension();
```
4ian, line 31: Pass the path to the zip file, otherwise from this point of view it looks like it's magically guessed (and could break without knowing for someone doing a change in this file).

scripts/extract-extension.js
```diff
@@ -0,0 +1,52 @@
...
+const pipeline = require('util').promisify(require('stream').pipeline);
+
+/**
+ * Extracts exactly one extension into the community extensions folder from a zip file.
+ * @param {string} [zipPath] The path to the zip file to extract.
+ * @param {string} [extensionsFolder] The folder with the extensions.
+ * @returns {Promise<{error: "too-many-files" | "no-json-found"| "invalid-file-name" | "zip-error", details?: any} | {error?: undefined,extensionName: string}>} the name of the extracted extension if successful, else a generic error code.
+ */
+exports.extractExtension = async function (
+  zipPath = '/tmp/ext.zip',
```
4ian, line 14: Don't make such an important parameter implicit:
- not setting it would most likely be an error from the caller (and confusing for the reader "who the heck do they know where the extension to extract is??")
- it's dividing the places where `/tmp/ext.zip` is set to two files, so there is one more chance one is updated without the other by a future ("why this file was not passed by the caller, argh I forgot to update it!!").

scripts/extract-extension.js
```diff
@@ -0,0 +1,52 @@
...
+
+  // Find JSON files in the zip top level folder
+  const jsonFiles = zip.file(/.*\.json$/);
+
+  // Ensure there is exactly 1 file
+  if (jsonFiles.length === 0) return { error: 'no-json-found' };
+  if (jsonFiles.length > 1) return { error: 'too-many-files' };
+  const [file] = jsonFiles;
+
+  if (file.name.slice(-5) !== '.json') return { error: 'invalid-file-name' };
```
4ian, line 32: Nit:
```suggestion
  if (path.extname(file.name) !== '.json') return { error: 'invalid-file-name' };
```

scripts/extract-extension.js
```diff
@@ -0,0 +1,52 @@
...
+  const extensionName = file.name.slice(0, -5);
```
4ian, line 33: Ditto

scripts/extract-extension.js
```diff
@@ -0,0 +1,52 @@
...
+
+  // Ensure no special characters are in the extension name to prevent relative path
+  // name schenanigans with dots that could put the extension in the reviewed folder
```
4ian, line 36: Ubernit: I think best practices encourage to put a period, comments are sentences:  https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

```suggestion
  // name schenanigans with dots that could put the extension in the reviewed folder.
```

scripts/lib/ExtensionNameValidator.js
```diff
@@ -0,0 +1,31 @@
+/** @param {string} extensionName */
+exports.isValidExtensionName = (extensionName) => {
+  if (extensionName.length === 0) return false;
+
+  {
```
4ian, line 5: No need for this block?

.github/workflows/auto-pr.yml
```diff
@@ -10,71 +10,130 @@ jobs:
...
+                repo: context.repo.repo,
+                body: `👋 Thanks for your submission! We are sorry, but it seems we already have a reviewed extension with that name. Extensions cannot be in both the community and reviewed extensions list at once. If you are trying to update that extension, please ask a member of the extension team for help. If the name clash is a coincidence, please update your original submission post with a new zip file containing your extensions with a file name that is not already taken 🙏`
+              })
+
+            if(code === "invalid-json")
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: `👋 Thanks for your submission! We are sorry, but your extension file inside the zip file seems to be damaged, as our system wasn't able to understand it's content. Try to re-export the extension JSON file, or to create the zip with another software, and to update your original submission post with the new zip file. If that still doesn't work, ask a member of the extension team for help 😉`
```
4ian, line 101:
```suggestion
                body: `👋 Thanks for your submission! We are sorry, but your extension file inside the zip file seems to be damaged, as our system wasn't able to understand its content. Try to re-export the extension JSON file, or to create the zip with another software, and to update your original submission post with the new zip file. If that still doesn't work, ask a member of the extension team for help 😉`
```

.github/workflows/auto-pr.yml
```diff
@@ -10,71 +10,130 @@ jobs:
...
+                repo: context.repo.repo,
+                body: `👋 Thanks for your submission! We are sorry, but it seems your extension file miraculously disappeared just after being extracted 🤔. This should never happen, please ask @GDevelopApp/extensions-team for assistance.`
+              })
+
+            if(code === "duplicated" )
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: `👋 Thanks for your submission! We are sorry, but it seems we already have a reviewed extension with that name. Extensions cannot be in both the community and reviewed extensions list at once. If you are trying to update that extension, please ask a member of the extension team for help. If the name clash is a coincidence, please update your original submission post with a new zip file containing your extensions with a file name that is not already taken 🙏`
```
4ian, line 93:
```suggestion
                body: `👋 Thanks for your submission! We are sorry, but it seems we already have a reviewed extension with that name. Extensions cannot be in both the community and reviewed extensions list at once. If you are trying to update that extension, please ask a member of the extension team for help. If the name clash is a coincidence, please update your original submission post with a new zip file containing your extension with a file name that is not already taken 🙏`
```

.github/workflows/auto-pr.yml
```diff
@@ -10,71 +10,130 @@ jobs:
...
-          Please fix them, and edit the link in your original submission to a zip file with the fixed extension.
+            const { code, errors } = await verifyExtension(extensionName);
 
-          Thanks again for your hard work!"
+            if(code === "invalid-file-name")
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: `👋 Thanks for your submission! We are sorry, but the filename of the extension has unrecognized characters. Since filenames with weird character can cause all kinds of problems and security flaws, our system won't allow file names with other characters than normal latin upper- and lowercase characters or numbers. Additionally, the first character must be an uppercase character. Please update your original submission post with a new zip file containing your extensions with another file name following those guidelines 🙏`
```
4ian, line 77:
```suggestion
                body: `👋 Thanks for your submission! We are sorry, but the filename of the extension has unrecognized characters. Since filenames with weird character can cause all kinds of problems and security flaws, our system won't allow file names with other characters than normal latin upper- and lowercase characters or numbers. Additionally, the first character must be an uppercase character. Please update your original submission post with a new zip file containing your extension with another file name following those guidelines 🙏`
```

.github/workflows/auto-pr.yml
```diff
@@ -10,71 +10,130 @@ jobs:
...
+              })
+
+            if(error === "invalid-file-name")
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: `👋 Thanks for your submission! We are sorry, but the filename of the extension has unrecognized characters. Since filenames with weird character can cause all kinds of problems and security flaws, our system won't allow file names with other characters than normal latin upper- and lowercase characters or numbers. Additionally, the first character must be an uppercase character. Please update your original submission post with a new zip file containing your extensions with another file name following those guidelines 🙏`
```
4ian, line 62:
```suggestion
                body: `👋 Thanks for your submission! We are sorry, but the filename of the extension has unrecognized characters. Since filenames with weird character can cause all kinds of problems and security flaws, our system won't allow file names with other characters than normal latin upper- and lowercase characters or numbers. Additionally, the first character must be an uppercase character. Please update your original submission post with a new zip file containing your extension with another file name following those guidelines 🙏`
```

This message is duplicated, is this normal?

.github/workflows/auto-pr.yml
```diff
@@ -10,71 +10,130 @@ jobs:
...
 
-          Thanks for your submission! Unfortunately, our automated test system found a few issues with your extension:
+            const { error, extensionName } = await extractExtension();
+
+            if(error === "zip-error")
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: `👋 Thanks for your submission! Sadly, we were unable to properly extract your extension from it's zip file. Try to create the zip with another software, and to update your original submission post with the new file. If that still doesn't work, ask a member of the extension team for help 😉`
```
4ian, line 38:
```suggestion
                body: `👋 Thanks for your submission! Sadly, we were unable to properly extract your extension from its zip file. Try to create the zip with another software, and to update your original submission post with the new file. If that still doesn't work, ask a member of the extension team for help 😉`
```

4ian (approved):
Very nice! 
`actions/github-script` seems nice. Let's try not to put more code in it (the comments are fine, but nothing else, or it will become too complex).
I've added a few comments and nitpickings. 

Happy that we're making that more robust! Thanks for the tests in particular :)

scripts/check-single-extension.js
arthuro555: It is a `Buffer`, not a `string`

.github/workflows/auto-pr.yml
arthuro555: Damn I keep on making that mistake, I got to do something to learn writing "it's/its" correctly 😓

.github/workflows/auto-pr.yml
arthuro555: Yes, because both are basically the same error. That check is technically duplicated, but the first one is made with the file in JSZip and the other on the extracted file, so I prefer to keep both just in case there is a path name representation inconsistency between the system FS and the JSZip FS that would cause this to be somehow different

scripts/check-single-extension.js
4ian: Fine then!

scripts/lib/ExtensionNameValidator.js
arthuro555: It is not needed but I like to make blocks like that to get rid of the temporary variable as soon as I stop needing it.

4ian:
Feel free to merge once ready!

PR #494, opened 2022-07-14
[PinchGesture] Add an extension to move the camera or objects with pinching gestures.

D8H:
## Description

This extension allows to:
* Move, rotate and/or zoom the camera according to pinch gestures
* Move, rotate and optionally scale objects according to pinch gestures
* Handle more specific needs with various expressions

## Demo
* Project: https://www.dropbox.com/s/rcuboz3maw9eqb7/PinchExample.zip?dl=1
* Build: https://liluo.io/instant-builds/09b9ce7a-2a2a-48dd-9af6-3de4614f5720

## JavaScript
There is some lines of JavaScript.

```JavaScrip
const x = eventsFunctionContext.getArgument("PositionX");
const y = eventsFunctionContext.getArgument("PositionY");
const layer = runtimeScene.getLayer(eventsFunctionContext.getArgument("Layer"));

eventsFunctionContext.returnValue = layer.convertInverseCoords(x, y)[0];
```

### DIFF
--- a/extensions/reviewed/PinchGesture.json
+++ b/extensions/reviewed/PinchGesture.json
@@ -1,7 +1,7 @@
 {
   "author": "",
   "category": "",
-  "description": "This extension allows to:\n* Move, rotate and/or zoom the camera according to pinch gestures\n* Move, rotate and optionally scale objects according to pinch gestures\n* Handle more specific needs with various expressions",
+  "description": "This extension allows to:\n* Move, rotate and/or zoom the camera according to pinch gestures\n* Move, rotate and optionally scale objects according to pinch gestures\n* Handle more specific needs with various expressions\n\nLimitations:\n* Only one pinching gesture is detected at a time",
   "extensionNamespace": "",
   "fullName": "Pinching gesture",
   "helpPath": "",
@@ -2244,6 +2244,19 @@
               "actions": [],
               "events": [
                 {
+                  "type": "BuiltinCommonInstructions::Comment",
+                  "color": {
+                    "b": 109,
+                    "g": 230,
+                    "r": 255,
+                    "textB": 0,
+                    "textG": 0,
+                    "textR": 0
+                  },
+                  "comment": "Only the objects with the highest Z-Order will be actually pinched (see doStepPostEvents).",
+                  "comment2": ""
+                },
+                {
                   "type": "BuiltinCommonInstructions::Standard",
                   "conditions": [
                     {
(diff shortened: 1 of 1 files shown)


tristanbob:
Thanks for working on this! It will be very useful in MANY games, including Absorbus.  :) 

- The "AbortPinching" action has the exact same event from condition "IsPinched"

- Object pinch currently checks if the starting center of the pinch is inside the object.  I think it would be better if both starting points of the pinch started inside the object.

D8H:
> The "AbortPinching" action has the exact same event from condition "IsPinched"

Good catch, I made the change.

> Object pinch currently checks if the starting center of the pinch is inside the object.  I think it would be better if both starting points of the pinch started inside the object.

The issue is that when a finger touches an object physically, its point is not necessarily inside the object. I guess it could be frustrating especially for tiny objects.
As the pinch needs 2 touches, the scenes events can probably disable the behaviors when a 1st button of the UI is pressed to avoid unrequited pinches.

Jurfix:
Very nice and useful extensions! Thank you for it! I myself sat and thought for a week how to do it, and all the time it turned out crooked, but everything is at the highest level with you. The only thing I should pay attention to is scene 3 with dragging objects, they sometimes behave strangely, but everything is fine, finally my Alarm Map will have an adequate zoom.

tristanbob:
I'll try to implement this on Absorbus tomorrow (Saturday).   I'll either approve or give feedback then.

Jurfix:
I noticed a small feature of this extension, it starts working without any declarations in events, i.e. in order for the zoom and camera rotation to work on all scenes, it’s enough just to add the extension to the project, it’s a little inconvenient and confused me, because I needed zoom only on the first scene, and I couldn’t think that it would work on all scenes at the same time, I had to turn it off behavior on each scene manually..

Jurfix:
I found an error in your extension, this is the "index" variable, I personally use this variable very often in my projects, in almost all cycles, please change the name of the variable to something more unpredictable, I personally did not notice this at first, which led to the repetition of my cycle adding text to an infinite number of times.
image

D8H:
> I found an error in your extension, this is the "index" variable, I personally use this variable very often in my projects, in almost all cycles, please change the name of the variable to something more unpredictable, I personally did not notice this at first, which led to the repetition of my cycle adding text to an infinite number of times.

Yes, good catch, I copied pasted this code from an example and forgot to rename it.

D8H:
> I noticed a small feature of this extension, it starts working without any declarations in events, i.e. in order for the zoom and camera rotation to work on all scenes, it’s enough just to add the extension to the project, it’s a little inconvenient and confused me, because I needed zoom only on the first scene, and I couldn’t think that it would work on all scenes at the same time, I had to turn it off behavior on each scene manually..

I changed it to disabled by default.

> The only thing I should pay attention to is scene 3 with dragging objects, they sometimes behave strangely

Yes, I'd added the draggable behavior to copy the wording and forgot to removed it. I updated the example.

Bouh (approved):

PR #497, opened 2022-07-17
Update DraggablePhysics.json to force Kinematic to Dynamic

tristanbob:
- Change Kinematic mode to Dynamic mode to prevent crash
- Set "left" mouse button as default

## Playable game

https://liluo.io/games/02910630-65f7-4dc4-aa26-3f262bc888ec
This object is assigned kinematic, but the extension forces it to be dynamic.

## Project files
[DraggableNonDynamic_BugFix.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/9128505/DraggableNonDynamic_BugFix.zip)

### DIFF
--- a/extensions/reviewed/DraggablePhysics.json
+++ b/extensions/reviewed/DraggablePhysics.json
@@ -1,7 +1,7 @@
 {
   "author": "",
   "category": "",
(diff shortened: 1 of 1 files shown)


D8H:
What about the static body type? Won't the same issue happen?

tristanbob:
> What about the static body type? Won't the same issue happen? 

I checked and objects that are "static" does not cause the error messages on the browser console.

> Could the check only be done in `onCreate`? It will still avoid bad configurations and I doubt beginners will change the body type with events.

I really like this idea, but for some reason, that method makes the object's position always start at 0,0.

https://user-images.githubusercontent.com/8879811/179432450-8c9c5fa2-b136-429e-b453-fd1fe25604be.mp4

D8H (approved):

PR #529, opened 2022-08-06
Add the CameraImpulse extension.

D8H:
* It can be used to simulate earthquakes or impacts.

## Example
It also uses:
- https://github.com/GDevelopApp/GDevelop-extensions/pull/387

Press `I` to start an impulsion and `S` to shake.
I didn't find any conflict between the impulse and the scrolling or the screen shake.
 
- Project: https://www.dropbox.com/s/spvas50648blctc/platformer-with-tilemap-shake.zip?dl=1
- Build: https://games.gdevelop-app.com/game-881fc8e5-272e-460a-8d72-7d01baca255d/index.html

### DIFF
--- a/extensions/reviewed/CameraImpulse.json
+++ b/extensions/reviewed/CameraImpulse.json
@@ -1,0 +1,795 @@
+{
+  "author": "",
+  "category": "",
+  "description": "Move the camera following an impulse trajectory.\n\nIt can be used to simulate earthquakes or impacts.",
+  "extensionNamespace": "",
+  "fullName": "Camera impulse",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


tristanbob:
Now that we have Camera tweens, do you think this extension is still needed?  It seems the logic could be duplicated with 3 events:

1) Tween camera to new position/rotation/zoom.
2) Wait for duration of tween + any additional duration desired
3) Tween back to original camera values

Am I missing something?  Will my logic above be incompatible with screenshake at the same time?

D8H:
> Now that we have Camera tweens, do you think this extension is still needed? It seems the logic could be duplicated with 3 events:
> 
>     1. Tween camera to new position/rotation/zoom.
> 
>     2. Wait for duration of tween + any additional duration desired
> 
>     3. Tween back to original camera values
> 
> 
> Am I missing something? Will my logic above be incompatible with screenshake at the same time?

There are 2 differences:
- The tween action will override any change done on the camera position for instance a scrolling. The camera tween action changes the position of the camera whereas this extension does a displacement of the camera.
- This extension reverts any change before scene events. This allows to do a scrolling without fighting against the impulse.

Bouh (approved):

PR #542, opened 2022-08-23
Health with shield (update)

tristanbob:
Health extension

## Playable game:

https://liluo.io/victrisgames/extension-health

## Project files

https://github.com/GDevelopApp/GDevelop-examples/pull/374

## Summary:

Manage health (life) and shield points of an object. Health can be assigned to players, enemies, NPCs, or even inanimate objects such as breaking doors and mining rocks.

## How to use:

- Set desired values in "Health" and "Shield" behavior properties
- To apply damage use “Damage an object” or “Damage an object protected by a shield”
- To heal health, use “Heal object” action 
- To activate the shield, use “Activate Shield” action
- Health and shield regeneration will happen automatically based on properties
 
## Tips:

- While damage cooldown is active, no damage will be applied to health or shield
- Try using the “Blink” extension while damage cooldown is active
- Use the “Is just damaged” condition to trigger audio and visual effects

## Health properties:

- Health (life points) Used for starting health
- Maximum health (0 for no maximum)
- Damage cooldown (seconds) Use "0" for no cooldown
- Rate of health regeneration (points per second) Will not exceed max health, even with overhealing enabled
- Delay (seconds) before health regeneration starts after being hit
- Allow heal to increase health above max health (will not exceed max health)

## Shield properties:

- Max shield points (Shield will start with this many points when activated)
- Duration of shield (seconds) Use "0" to disable shield timer
- Rate of shield regeneration (points per second)
- Shield will block all excess damage when shield is destroyed
- Delay (seconds) before shield regeneration starts after being hit
- Regenerate shield when it has zero points


Bouh (approved):

D8H (changes requested):
Very nice improvements for this extension.

Some suggestions:
- There is no description for "RequireHealth" maybe something like "Health behavior". The name should probably be "HealthBehavior" instead of "RequireHealth" as it's a property and not an action.
- ShieldDuration, "Use 0 for making it permanent" instead of "Use 0 to disable shield timer"
- MaxShieldPoints, "Shield will start with this many points when activated" should there be another property for the initial shield value? (like there is a Health and MaxHealth properties)
- Add function groups for the configuration to clearly see what the "true" actions are.
- Avoid to start each action by "Change" it's very hard to scan a list if the 1st letter is always the same.
- Try to be as much concise as possible for function full name. For instance, "Change delay before health regeneration starts" can just be "Health regeneration delay".
- The health generation action is in the shield group.

tristanbob:
> * There is no description for "RequireHealth" maybe something like "Health behavior". The name should probably be "HealthBehavior" instead of "RequireHealth" as it's a property and not an action.

Good idea, done.

> * ShieldDuration, "Use 0 for making it permanent" instead of "Use 0 to disable shield timer"

Done.  (I actually called it "permanent" before and changed it.  But I think your idea is good.)

> * MaxShieldPoints, "Shield will start with this many points when activated" should there be another property for the initial shield value? (like there is a Health and MaxHealth properties)

Great question!

**Health**

Starting value: Health points
Dynamic value: Health points
How to reset to starting value? Impossible, but hopefully the user set MaxHealth the same as Health.

**Shield**

Starting value: Max shield points
Dynamic value: Shield points
How to reset to starting value? Max shield points

I wish Health worked like I designed Shield, but I think changing it might be considered a breaking change.  
What do you suggest? I don't think we need another property.

> * Add function groups for the configuration to clearly see what the "true" actions are.

Done.

> * Avoid to start each action by "Change" it's very hard to scan a list if the 1st letter is always the same.

Done.

> * Try to be as much concise as possible for function full name. For instance, "Change delay before health regeneration starts" can just be "Health regeneration delay".

I like that; thanks! I changed all similar ones.

> * The health generation action is in the shield group.

Fixed.

D8H:
> > * MaxShieldPoints, "Shield will start with this many points when activated" should there be another property for the initial shield value? (like there is a Health and MaxHealth properties)
> 
> Great question!
> 
> **Health**
> 
> Starting value: Health points Dynamic value: Health points How to reset to starting value? Impossible, but hopefully the user set MaxHealth the same as Health.
> 
> **Shield**
> 
> Starting value: Max shield points Dynamic value: Shield points How to reset to starting value? Max shield points
> 
> I wish Health worked like I designed Shield, but I think changing it might be considered a breaking change. What do you suggest?

It's possible to see `Health` as initial health and add an hidden property `CurrentHealth`.

> I don't think we need another property.

It's almost a convention to start full health, but I think it's good that the extension allows to start with any value. I think it should be the same for the shield especially because extension users may want their game to start with a shield at 0 if collectables regenerate it.

tristanbob:
> It's almost a convention to start full health, but I think it's good that the extension allows to start with any value. I think it should be the same for the shield especially because extension users may want their game to start with a shield at 0 if collectables regenerate it.

After thinking about this for a few days, I have updated both health and shield so that there are 3 properties for each:

- Health (this is used for starting health, but I'm keeping the name for backward compatibility)
- CurrentHealth (hidden)
- MaxHealth

- StartingShieldPoints
- CurrentShieldPoints (hidden)
- MaxShieldPoints

Let me know what you think.

D8H:
- I wonder if there is too much function group.
  - I usually do a category for actions that change a property value because users define them with the properties editor so they will rarely use them. It allows to quickly find what the "real" actions are.
  - The groups could be:
    - Health
	- Health configuration
	- Shield
	- Shield configuration
  - I'm not sure if "configuration" is the best term for it though.
- What is the purpose of these functions? Should they be hidden?
  - SetHitAtLeastOnce
  - SetJustDamaged
- "Duration of damage cooldown" to "Damage cooldown"?
- "Regenerate shield when it has zero points" to "Regenerate when broken"?
- "Delay before shield regeneration starts" to "Regeneration latency" (same for health)?
- "Shield blocks excess damage when it breaks" to "Block exceeding damaged"?

tristanbob:
> * I wonder if there is too much function groups.

Yes, I think you are probably right.  Here are the current conditions and actions:

image
image

>   * I usually do a group for actions that change a property value because users define them with the properties editor so they will rarely be used. It allows to quickly find what the "real" actions are.
>   * The groups could be:
>     
>     * Health
>     * Health configuration
>     * Shield
>     * Shield configuration
>   * I'm not sure if "configuration" is the best term for it though.

I have changed it to match your example.  We could use "Shield settings" but since you have used "configuration" already, lets keep it consistent.

> * What is the purpose of these functions? Should they be hidden?
>   
>   * SetHitAtLeastOnce
>   * SetJustDamaged

Yes, these are now hidden.  Thanks!

> * Some names are a bit long
>   
>   * "Duration of damage cooldown" to "Damage cooldown"?
>   * "Regenerate shield when it has zero points" to "Regenerate when broken"?
>   * "Delay before shield regeneration starts" to "Regeneration latency" (same for health)?
>   * "Shield blocks excess damage when it breaks" to "Block exceeding damaged"?

I tried to improve these. 

Thanks for the reviews!

D8H:
Why did you choose to keep these?
  - "Regenerate shield when it has zero points"
  - "Shield blocks excess damage when it breaks"

- There is a typo in "zero points".
- "Apply damage to an object" to "Damage the health"?
- "Apply damage to a shielded object" to "Damage (with shield)"?

tristanbob:
To be honest, I was rushing to get this updated before taking the kids to school.  :)

> * Why did you choose to keep these?
>   
>   * "Regenerate shield when it has zero points"

Ok, I changed it.

>   * "Shield blocks excess damage when it breaks"

OK, I made it a bit shorter:

"Block excess damage when shield is broken"

> * There is a typo in "zero points".

Where? I guess I removed this.

> * "Apply damage to an object" to "Damage the health"?
> * "Apply damage to a shielded object" to "Damage (with shield)"?

I thought this was better.

extensions/reviewed/Health.json
```diff
@@ -19,96 +19,3851 @@
...
+          "objectGroups": []
+        },
+        {
+          "description": "Change shield points. This is not considered damage.",
+          "fullName": "Shield points",
+          "functionType": "Action",
+          "group": "Shield configuration",
+          "name": "SetShieldPoints",
+          "private": false,
+          "sentence": "Change the shield on _PARAM0_ to _PARAM2_ points",
```
D8H, line 3354: > This is not considered damage.

The sentence doesn't seem clear.

extensions/reviewed/Health.json
```diff
@@ -19,96 +19,3851 @@
...
+          "objectGroups": []
+        },
+        {
+          "description": "Check if the object has been hit at least once.",
+          "fullName": "Object has been hit at least once",
+          "functionType": "Condition",
+          "group": "Health",
+          "name": "HitAtLeastOnce",
+          "private": false,
+          "sentence": "_PARAM0_ has been hit at least once",
```
D8H, line 2343: Should this be private?
What is it used for?

D8H:
It's just an idea: instead of expressions to get the result of the last extension, users could use an expression to evaluate a hit before applying it. It could be useful for AI or do a damage preview in the UI.

Most expressions full name are too long.

extensions/reviewed/Health.json
tristanbob: Thanks, I have updated it to:

`Change current shield points. Will not trigger damage cooldown.`
`Change current shield points on _PARAM0_ to _PARAM2_ points`

extensions/reviewed/Health.json
tristanbob: Yes, it should be private, I have changed it.

It is used so that when the game starts, the damage cooldown is not considered active.

PR #545, opened 2022-08-27
[New community extensions] Geolocation, References and ObjectPickingTools

arthuro555:
Adds a few community extension I made some time ago and forgot to submit.

Geolocation - Allows to locate a player using the Web Geolocation API
References - Allows to make a scene variable a reference to some other variable (necessary for some algorithms and data structures, albeit rather advanced hence why it is an extension rather than a core feature)
ObjectPickingTools - An extension requested by Silver-Streak sometime ago to help with object picking by allowing to pick objects of an object list one-by-one, unpick them, pick the one with highest/lowest z-Order...

Also adds a small fix to the build code for the auto-fixing feature.

### DIFF
--- a/extensions/community/Geolocation.json
+++ b/extensions/community/Geolocation.json
@@ -1,0 +1,138 @@
+{
+  "author": "Arthur Pacaud (arthuro555)",
+  "category": "Device",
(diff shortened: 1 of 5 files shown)


D8H (approved):
I haven"t looked into the extensions, but It looks good to me.

PR #554, opened 2022-09-01
Sine movement; Added actions to set the horizontal and vertical progress

tristanbob:
A user requested this feature and I think it is useful also.

## Playable game

https://liluo.io/instant-builds/22eed7d5-6c41-4f15-9705-b1b50aab2947

## Project

[SineMovement-0.0.6.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/9467006/SineMovement-0.0.6.zip)

## Video

https://user-images.githubusercontent.com/8879811/187835369-8d94f1a6-9d31-4747-b559-51d6abe87cb6.mp4

### DIFF
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -970,7 +970,7 @@
           "objectGroups": []
         },
         {
-          "description": "Set vertical distance.",
+          "description": "Change vertical amplitude.",
           "fullName": "Vertical distance",
           "functionType": "Action",
           "group": "Sine movement configuration",
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -1239,7 +1239,7 @@
         {
           "value": "100",
           "type": "Number",
-          "label": "Horizontal distance: amplitude of the movement on X axis (0 to deactivate)",
+          "label": "Horizontal amplitude",
           "description": "",
           "group": "Sine movement",
           "extraInformation": [],
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -474,7 +474,7 @@
         },
         {
           "description": "Horizontal distance.",
-          "fullName": "Horizontal distance",
+          "fullName": "Horizontal amplitude",
           "functionType": "Expression",
           "group": "Sine movement configuration",
           "name": "HorizontalDistance",
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -976,7 +976,7 @@
           "group": "Sine movement configuration",
           "name": "SetVerticalDistance",
           "private": false,
-          "sentence": "Set vertical distance of _PARAM0_ to _PARAM2_",
+          "sentence": "Change the vertical amplitude of _PARAM0_ to _PARAM2_",
           "events": [
             {
               "type": "BuiltinCommonInstructions::Standard",
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -473,7 +473,7 @@
           "objectGroups": []
         },
         {
-          "description": "Horizontal distance.",
+          "description": "Return the horizontal amplitude.",
           "fullName": "Horizontal amplitude",
           "functionType": "Expression",
           "group": "Sine movement configuration",
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -909,7 +909,7 @@
           "objectGroups": []
         },
         {
-          "description": "Set horizontal distance.",
+          "description": "Change the horizontal amplitude.",
           "fullName": "Horizontal distance",
           "functionType": "Action",
           "group": "Sine movement configuration",
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -521,7 +521,7 @@
           "objectGroups": []
         },
         {
-          "description": "Vertical distance.",
+          "description": "Return the vertical amplitude.",
           "fullName": "Vertical amplitude",
           "functionType": "Expression",
           "group": "Sine movement configuration",
--- a/extensions/reviewed/SineMovement.json
+++ b/extensions/reviewed/SineMovement.json
@@ -1,7 +1,7 @@
 {
   "author": "@4ian, Tristan Rhodes (https://victrisgames.itch.io/)",
   "category": "",
-  "description": "Allow an object to move smoothly on the X and/or Y axis following a sine wave, or an ellipsis.\n\n\nExample uses:\n- Floating objects, such as powerups or coins\n- Moveable platforms\n- Enemy movement patterns\n\nProperties:\n- Horizontal distance\n- Vertical distance\n- Horizontal speed\n- Vertical speed\n- Center of movement, X position\n- Center of movement, Y position\n\nTips:\n- For circular or elliptical movement, the horizontal and vertical speed need to be the same\n- For horizontal movement only, set vertical distance to 0\n- For vertical movement only, set horizontal distance to 0\n- For figure-8 movement, set horizontal speed to 1/2 of the vertical speed",
+  "description": "Allow an object to move smoothly on the X and/or Y axis following a sine wave, or an ellipsis.\n\nExample uses:\n- Floating objects, such as powerups or coins\n- Moveable platforms\n- Enemy movement patterns\n\nProperties:\n- Horizontal distance\n- Vertical distance\n- Horizontal speed\n- Vertical speed\n- Center of movement, X position\n- Center of movement, Y position\n\nTips:\n- For circular or elliptical movement, the horizontal and vertical speed need to be the same\n- For horizontal movement only, set vertical distance to 0\n- For vertical movement only, set horizontal distance to 0\n- For figure-8 movement, set horizontal speed to 1/2 of the vertical speed",
   "extensionNamespace": "",
   "fullName": "Sine (or ellipsis) Movement",
   "helpPath": "https://victrisgames.itch.io/extension-sinemovement-and-deptheffect",
@@ -32,7 +32,7 @@
   "eventsFunctions": [],
   "eventsBasedBehaviors": [
     {
-      "description": "Allow an object to move smoothly on the X and/or Y axis following a sine wave.\n\nExample uses:\n- Floating pickups\n- Moveable platforms\n- Enemy movement\n\nProperties:\n- Center of movement, X position\n- Center of movement, Y position\n- Horizontal distance\n- Vertical distance\n- Horizontal speed\n- Horizontal distance\n\nTips:\n- For circular or elliptical movement, the horizontal and vertical speed need to be the same\n- For horizontal movement, set vertical distance to 0\n- For vertical movement, set horizontal distance to 0\n- For figure-8 movement, set horizontal speed to 1/2 of the vertical speed.",
+      "description": "Allow an object to move smoothly on the X and/or Y axis following a sine wave.",
       "fullName": "Sine Movement",
       "name": "SineMovement",
       "objectType": "",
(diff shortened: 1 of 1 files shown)


D8H:
What's the use of CenterPointX and CenterPointY properties?

Some suggestions:
- "Reset sine progress counters" to "Reset at center"?
- A "Sine movement configuration" functions group? to have the 3 progress actions in the default group. 
- Remove the "Set" prefix in actions full name.
- "amplitude instead of "distance".

tristanbob:
> What's the use of CenterPointX and CenterPointY properties?

It saves the starting position (or it can be set manually).  I tried moving this logic to "OnCreated", but that function doesn't know the X() and Y() position of the object because they always return "0".

> Some suggestions:
> 
> * "Reset sine progress counters" to "Reset at center"?

I changed it to "Reset to starting position"

> * A "Sine movement configuration" functions group? to have the 3 progress actions in the default group.

Ok, I think I changed this to what you ask.

> * Remove the "Set" prefix in actions full name.

Done.

> * "amplitude instead of "distance".

Where did you want this changed?  I see the properties has both "distance" and "amplitude" in the description.

tristanbob:
@D8H any other changes for this review?  (Thanks for all your help!)

extensions/reviewed/SineMovement.json
```diff
@@ -1185,6 +1241,7 @@
           "type": "Number",
           "label": "Horizontal distance: amplitude of the movement on X axis (0 to deactivate)",
```
D8H, line 1242:
```suggestion
          "label": "Horizontal amplitude",
```

extensions/reviewed/SineMovement.json
```diff
@@ -1194,6 +1251,7 @@
           "type": "Number",
           "label": "Vertical distance: amplitude of the movement on Y axis (0 to deactivate)",
```
D8H, line 1252:
```suggestion
          "label": "Vertical amplitude",
```

extensions/reviewed/SineMovement.json
```diff
@@ -899,33 +971,29 @@
         },
         {
           "description": "Set vertical distance.",
-          "fullName": "Set vertical distance",
+          "fullName": "Vertical distance",
```
D8H, line 974:
```suggestion
          "fullName": "Vertical amplitude",
```

extensions/reviewed/SineMovement.json
```diff
@@ -899,33 +971,29 @@
...
           "functionType": "Action",
+          "group": "Sine movement configuration",
           "name": "SetVerticalDistance",
           "private": false,
           "sentence": "Set vertical distance of _PARAM0_ to _PARAM2_",
```
D8H, line 979:
```suggestion
          "sentence": "Change the vertical amplitude of _PARAM0_ to _PARAM2_",
```

extensions/reviewed/SineMovement.json
```diff
@@ -899,33 +971,29 @@
         },
         {
           "description": "Set vertical distance.",
```
D8H, line 973:
```suggestion
          "description": "Change vertical amplitude.",
```

extensions/reviewed/SineMovement.json
```diff
@@ -834,33 +910,29 @@
         },
         {
           "description": "Set horizontal distance.",
-          "fullName": "Set horizontal distance",
+          "fullName": "Horizontal distance",
           "functionType": "Action",
+          "group": "Sine movement configuration",
           "name": "SetHorizontalDistance",
           "private": false,
           "sentence": "Set horizontal distance of _PARAM0_ to _PARAM2_",
```
D8H, line 918:
```suggestion
          "sentence": "Change the horizontal amplitude of _PARAM0_ to _PARAM2_",
```

extensions/reviewed/SineMovement.json
```diff
@@ -834,33 +910,29 @@
         },
         {
           "description": "Set horizontal distance.",
-          "fullName": "Set horizontal distance",
+          "fullName": "Horizontal distance",
```
D8H, line 913:
```suggestion
          "fullName": "Horizontal amplitude",
```

extensions/reviewed/SineMovement.json
```diff
@@ -834,33 +910,29 @@
         },
         {
           "description": "Set horizontal distance.",
```
D8H, line 912:
```suggestion
          "description": "Change the horizontal amplitude.",
```

extensions/reviewed/SineMovement.json
```diff
@@ -550,28 +524,24 @@
           "description": "Vertical distance.",
           "fullName": "Vertical distance",
```
D8H, line 525:
```suggestion
          "fullName": "Vertical amplitude",
```

extensions/reviewed/SineMovement.json
```diff
@@ -550,28 +524,24 @@
           "description": "Vertical distance.",
```
D8H, line 524:
```suggestion
          "description": "Return the vertical amplitude.",
```

extensions/reviewed/SineMovement.json
```diff
@@ -498,28 +476,24 @@
           "description": "Horizontal distance.",
           "fullName": "Horizontal distance",
```
D8H, line 477:
```suggestion
          "fullName": "Horizontal amplitude",
```

extensions/reviewed/SineMovement.json
```diff
@@ -498,28 +476,24 @@
           "description": "Horizontal distance.",
```
D8H, line 476:
```suggestion
          "description": "Return the horizontal amplitude.",
```

D8H:
> > What's the use of CenterPointX and CenterPointY properties?
> 
> It saves the starting position (or it can be set manually). I tried moving this logic to "OnCreated", but that function doesn't know the X() and Y() position of the object because they always return "0".

The behavior should probably watch position changes done by the outside and update the center accordingly.
If the property is only hidden (but keeping the action and expression), I think it's still fine. I can't picture a user setting the property when they can just place the object.
SineProgressX and SineProgressX could be made visible as initial values. Should they be between 0 and 1 or be called angle?

tristanbob:
>

tristanbob:
> The behavior should probably watch position changes done by the outside and update the center accordingly. 

This would be a nice feature, but it is not simple and I don't want to add that scope to this PR. 

> If the property is only hidden (but keeping the action and expression), I think it's still fine. I can't picture a user setting the property when they can just place the object. 

A user requested it, and I think my example that tracked mouse position could make some fun visual effects.

> SineProgressX and SineProgressY could be made visible as initial values. Should they be between 0 and 1 or be called angle?

If I was rewriting this from scratch, I would hide degrees from the user completely.  They just need to know:
- How far will it move in each direction (pixels)
- How long it takes to complete a full cycle (seconds)

Since we are not creating this from scratch, I'm not sure how to improve it. I wasn't planning on a full refactor when adding these additional functions that a user requested.

D8H:
> > If the property is only hidden (but keeping the action and expression), I think it's still fine. I can't picture a user setting the property when they can just place the object.
> 
> A user requested it, and I think my example that tracked mouse position could make some fun visual effects.

It don't understand. It doesn't need the property CenterPointX and CenterPointY to be visible in the configuration, does it?

> > SineProgressX and SineProgressY could be made visible as initial values. Should they be between 0 and 1 or be called angle?
> 
> If I was rewriting this from scratch, I would hide degrees from the user completely. They just need to know:
> 
>     * How far will it move in each direction (pixels)
> 
>     * How long it takes to complete a full cycle (seconds)
> 
> 
> Since we are not creating this from scratch, I'm not sure how to improve it. I wasn't planning on a full refactor when adding these additional functions that a user requested.

The "progress" wasn't visible to extension users before, right?
If this is the case, now is the right time to change it because after this update it won't be possible anymore without a breaking change.

tristanbob:
> It don't understand. It doesn't need the property CenterPointX and CenterPointY to be visible in the configuration, does it?

This is the only way the user can move an object with this behavior.  If they try moving the object, it gets instantly teleported to it's path based on the center value.  The only way to remove this is to only apply deltas (changes in position) instead of absolute position.  I might be able to figure out a way to do this, but I don't have any ideas at the moment.  (This is how the object shake extension tracks movement)

> The "progress" wasn't visible to extension users before, right? If this is the case, now is the right time to make the change because after this update it won't be possible anymore without a breaking change.

Yes, I agree, however, the speed and distance (amplitude) parameters did exist, so we can't change those.

D8H:
> > It don't understand. It doesn't need the property CenterPointX and CenterPointY to be visible in the configuration, does it?
> 
> This is the only way the user can move an object with this behavior. If they try moving the object using normal methods, the object gets instantly teleported along it's sine-path based on the center value. The only way to remove this is to only apply deltas (changes in position) instead of absolute position. I might be able to figure out a way to do this, but I don't have any ideas at the moment. (This is how the object shake extension tracks movement)

When this property is left to 0, it's set to the object position at the beginning. I'm just saying that it should be hidden to always do this. It doesn't change anything in the events.

> > The "progress" wasn't visible to extension users before, right? If this is the case, now is the right time to make the change because after this update it won't be possible anymore without a breaking change.
> 
> Yes, I agree, however, the speed and distance (amplitude) parameters did exist, so we can't change those.

Ok, the speed is expressed in angle per second. It should have been a duration for one cycle but well.
But, it doesn't forbid the "progress" to be between 0 and 1.

D8H:
I think we should make a breaking change to make it an ellipse (like it was initially) and allow to move the object this way:
- both amplitude != 0 => ellipse
- both amplitude != 0 + a force => loops
- one amplitude != 0 => back and forth
- one amplitude != 0 + force => waves

tristanbob:
> I think we should make a breaking change to make it an ellipse (with only one speed property) and allow to move the object this way:

I do not want to remove the ability to independently set the X and Y speed.  This feature provides the most flexibility for users to create a variety of movement types (back and forth, circle, figure 8, chaotic, etc)

> There could even be a property to rotate the object to face forward.

I thought about this, but I'd rather have people add the FaceForward extension to the object.

tristanbob:
> When this property is left to 0, it's set to the object position at the beginning. I'm just saying that it should be hidden to always do this. It doesn't change anything in the events.

Yes, I will hide this property but allow users to set it using events.

> Ok, the speed is expressed in angle per second. It should have been a duration for one cycle but well. But, it doesn't forbid the "progress" to be between 0 and 1.

I'll investigate and try to implement this recommendation.

D8H:
> > I think we should make a breaking change to make it an ellipse (with only one speed property) and allow to move the object this way:
> 
> I do not want to remove the ability to independently set the X and Y speed. This feature provides the most flexibility for users to create a variety of movement types (back and forth, circle, figure 8, chaotic, etc)

With `ProgressX` and `ProgressY`, how can users make sense of the actual progress?
A full cycle (both progress to 0) may even never happen.
A chaotic movement should be the role of another extension, a noise-based ShakeObject perhaps. If people want a strange rose movement, they can still add this extension twice or do this very specific thing themselves.
I bet that this extension is used with `speedX = speedY`, 90% of the time. Why make things harder for the most?

With only one speed, the "progress" can just be a angle which is easy to picture.
People could use an action "Turn the other way" that would toggle the center on the left or right side of the object. Repeated every 180°, it allows to do nice curves like this:
Arc

D8H:
A made a version of the extension.

New features:
- The object starts from where it was on the scene.
- The initial angle and direction can be setup with properties
- The object can turn left and right
- The object can be move by another behavior (force, screen warping...)

I made a mini game in only 2 events to illustrate:

image

The build:
https://liluo.io/games/be98d949-c9a0-4972-86df-1cfb5a10fea6

Its project:
https://www.dropbox.com/s/g55dyp1vj5gtlm7/EllipseMovementExample.zip?dl=0

tristanbob:
This extension has been deprecated and replaced with a much more versatile version: `Ellipse movement` https://github.com/GDevelopApp/GDevelop-extensions/pull/591

PR #582, opened 2022-09-21
Upgrade Cordova IAP to 11.0.0

ClementPasteau:
Following 2 raised problems on the forum: 

- https://forum.gdevelop.io/t/inapppurchase-outdated/41962
- https://forum.gdevelop.io/t/inapppurchase-unsupported-version-of-play-billing/41926/16

I believe we can safely update the cordova plugin to handle the Play Billing v4
https://github.com/j3k0/cordova-plugin-purchase/releases

### DIFF
--- a/extensions/reviewed/InAppPurchase.json
+++ b/extensions/reviewed/InAppPurchase.json
@@ -8,7 +8,7 @@
   "name": "InAppPurchase",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Glyphster Pack/Master/SVG/Shopping and Ecommerce/Shopping and Ecommerce_wallet_money_cash.svg",
   "shortDescription": "Add items to buy directly in your game (\"In-App Purchase\"), for games published on Android or iOS.",
-  "version": "0.0.1",
+  "version": "1.0.0",
   "tags": [
     "purchase",
     "iap",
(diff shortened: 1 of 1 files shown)


ClementPasteau:
cc @arthuro555 - can you confirm the update can be done safely?

arthuro555 (approved):
I skimmed through the changes in the plugin and through the changelog, since there seems to be no breaking change this should be safe, yes!

extensions/reviewed/InAppPurchase.json
```diff
@@ -8,7 +8,7 @@
   "name": "InAppPurchase",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Glyphster Pack/Master/SVG/Shopping and Ecommerce/Shopping and Ecommerce_wallet_money_cash.svg",
   "shortDescription": "Add items to buy directly in your game (\"In-App Purchase\"), for games published on Android or iOS.",
-  "version": "0.0.1",
+  "version": "1.0.0",
```
arthuro555, line 11: No need for a major bump here, also the extension quality is pretty poor still as far as I remember, and a 0.x.x release indicate it is not fully production ready, so I would keep it to something like this instead:
```suggestion
  "version": "0.0.2",
```

4ian (approved):

PR #590, opened 2022-09-24
Rectangular movement rework

D8H:
## Changes
- The object starts from where it is and the rectangle location is evaluated from the "Initial position" property
- The rectangle dimension are set with width and height instead of distances from the center
- The rectangle can be changed from left, top, right and bottom instead of distance from the center.
- Use durations instead of speeds because speed is variable with easing and it's easier to picture.

## New features
- Easing function can be used.
- The rectangle follows the object when it moves from outside of the extension. It allow to combine several movement behaviors.
- The object can be teleported to a corner.

## Bug fix
- Changing the direction between clockwise and counter-clockwise no longer make the object jumps.

## Example
- https://github.com/GDevelopApp/GDevelop-examples/pull/386


D8H:
The speed is constant on the rectangle, but the easing function introduce acceleration. An easing on a smaller period mean more acceleration. It result to a feeling of quicker movement on smaller edges.
The extension should probably allows to set a duration for horizontal and vertical edges (like the previous extension did with 2 speed properties).
This could make more difficult to configure a constant speed with the "linear" easing but the linear easing don't look good and a small and users can solve this with a cross-multiplication.

"Horizontal/Vertical edge duration" property will allow user to choose a good trade of between speed and acceleration.
As you can see in this configuration, what I feel give a "same speed" feeling is not proportional to the dimensions and it's probably a bit different for each easing function.

image

HelperWesley:
Honestly, I like this a lot!

This levels up the previous rectangular movement behavior, and isn't complicated at all. 👍

(I have no idea how it's working under the hood, I'm just commenting on it's utility. 😅)

tristanbob (changes requested):
## Summary

First, let me thank you for updating this extension.  This is one of the more popular extensions, but it has a lot of limitations.  You can't easily move the object, which also means you cannot use "shake object".  With your version, this is now possible!

The old version had two movement styles, but yours offers all the easings! This offers a lot of different movement options.  

## Review

- Why did you change RectangularFloodFill to a community extension?  It's not related to this extension in any way.

- We should add a note in the description of the deprecated extension to recommend using this new extension

- Initial position: Consider adding the word "corner" to the string options.  For example `Top-right corner`

- "SetVerticalEdgeDuration" action is changing the horizontal duration property

- Edge durations do not appear to take the right amount of time.  It feels like they only take 1/4 of the duration specified.

- `Easing` is not a world that most people will recognize.  Can we add something descriptive to indicate what it does?  For instance: `Movement style (easing)`

D8H:
>     * We should add a note in the description of the deprecated extension to recommend using this new extension

At first, I wanted to do it but both extension have the same short description and actually I prefer that users only do the upgrade if they know what they are doing. They will find the new extension when they start a new project so the transition will happen progressively.

>     * Initial position: Consider adding the word "corner" to the string options.  For example `Top-right corner`
> 
>     * "SetVerticalEdgeDuration" action is changing the horizontal duration property
> 
>     * Edge durations do not appear to take the right amount of time.  It feels like they only take 1/4 of the duration specified.

Thanks, I've made the changes.

>     * `Easing` is not a world that most people will recognize.  Can we add something descriptive to indicate what it does?  For instance: `Movement style (easing)`

There is no explanation in the Tween extension. I guess that users can pick another one and see what it does.

tristanbob (approved):

tristanbob:
I tested the new version of the example game and everything looks good.  Great job, @D8H!

PR #591, opened 2022-09-24
Ellipse movement rework

D8H:
Following this discussion:
- https://github.com/GDevelopApp/GDevelop-extensions/pull/554

## Changes
- The object starts from where it is and the circle center is evaluated from the "Initial direction angle" and "angular velocity" properties.
- There is only one speed property. Stick to an ellipse movement make the extension easier to understand.
- The ellipse dimensions are set with "radius" properties instead of "distances".

## New features
- The ellipse follows the object when it moves from outside of the extension. It allows to combine several movement behaviors.
- The object can turn the other way with an action which allow this behavior to be used for a player movement or an AI.

## Example
- https://github.com/GDevelopApp/GDevelop-examples/pull/387


tristanbob (commented):
## Summary

This extension is great! I think it will be more useful than the previous SineMovement.  I especially like how it tracks movement caused from outside the extension, so that complex movement (such as object shaking) can be combined with the elliptical movement provided by this extension.  We should look for opportunities to add this capability to all movement extensions!

## Suggested changes

`Move objects on ellipses or on one axis following sine waves.`
While 100% correct, this wording is overly complex and doesn't fit the goals of GDevelop.   My first attempt is `Move object in a circular shape or move smoothly back and forth in one direction` but I don't really like this wording that much.

`Change the in which side the object is turning (left or right).`
- Should be `Change turning direction (left or right).` 
- Consider using clockwise and counterclockwise instead of left/right

"Movement angle" expression+action do not have a group

Use the same Group names (some are just a few characters different)

D8H:
Actually most movement behaviors use forces so they already can be combine.

> `Move objects on ellipses or on one axis following sine waves.` While 100% correct, this wording is overly complex and doesn't fit the goals of GDevelop. My first attempt is `Move object in a circular shape or move smoothly back and forth in one direction` but I don't really like this wording that much.

I took your last part "smoothly back and forth in one direction" but kept "ellipse" because ellipses are not circular. I understand that it's important to use simple terms but if they are too imprecise, neither children nor adults will understand.
 
> `Change the in which side the object is turning (left or right).`
> 
>     * Should be `Change turning direction (left or right).`

Ok, I added a "the" before "turning direction". Do you think it's ok?

>     * Consider using clockwise and counterclockwise instead of left/right

"right" and "left" are not great but "clockwise" doesn't works because the speed can be negative. When the speed is negative, "left" and "right" still works, it's like with a car that go backward.

(sorry, I inverted the ship direction and the speed sign on this drawing)
EllipseMovement

> "Movement angle" expression+action do not have a group
> 
> Use the same Group names (some are just a few characters different)

This is expected, expressions have hierarchical groups where instruction groups are only one level.
I'll do a PR for GDevelop to automatically remove the extension prefix from the expression group to allow to use the same group without having a repetition in the group hierarchy (expressions has a root group with the extension name).

Functions without a group use the default group. They are not part of the configuration.

tristanbob (approved):

PR #601, opened 2022-09-26
Remove an image in action parameters of SoundVolumeBasedOnDistance.json

Bouh:
Don't show in changelog


Bouh:
@VegeTato Hi thx for this extension, but please don't add images in the description of a parameter.

### Why?
We don't want images because if one day the image becomes unreachable this lead to an error. And images cannot be read after the installation if the user is offline.

If a user need help to set the parameters they should click on the **Help** button for this extension.

The Help button is configurable in the extension properties panel. 
And you can redirect the user to the wiki or somewhere else you want to detail the parameters.

There is also a problem of design on your side, we're too permissive.

We agree that the image in the description of the extension is useful because if you can read the description of an extension it means that you have internet so it will load.
But there is still a problem if the image becomes unreachable.


So please don't add more images to the parameters.
because offline people see a broken image.
image

PR #623, opened 2022-10-05
[Reviewed][Sway] Update categorie and fix description

Bouh:
Fixes for these:
https://github.com/GDevelopApp/GDevelop-extensions/pull/468#issuecomment-1268342801


D8H:
The "How to use" should either be in instruction or properties description or in a help page if more details is needed.
The extension description only answer: "Do I need this extension?"

Bouh:
> The "How to use" should either be in instruction or property descriptions or in a help page if more details is needed. The extension description only answer: "Do I need this extension?" Coming back to the extension description for help is a bad UX.

The help page redirects to a Liluo profile. I agree this isn't the best thing.
Is there an auto-generated page on the wiki for this extension?


Also, the imgur image is broken I plan to edit the description like this:

```

Use the Sway behavior to make multiple instances of an object sway at different times. For example, this can give a grass object an animated look that is always changing and is different to the other grass objects.

Sway works by:
- Angling (pivoting) the object to the left and right while also changing its Y scale up and down
- The angle and Y scale tween durations will take a random amount of time from a range of values
- It's this random tween duration that creates the unique movement pattern for each instance of the object.

Please note:
The behavior works best if both the *Center* and *Origin* points are at the base of the object.

How to use:
- Set left and right angles (degrees) for the object to sway to
- For equal sway on each side set the left angle to the negative number of the right angle, eg. set left angle to -5 and right angle to 5

- Set a minimum and maximum Y scale amount for the object to stretch and shrink to
- Set to 1 for no change, more than 1 for stretching, and less than 1 for shrinking

- Set a range for the angle and Y scale tween times (seconds)
- The tweens will take a random amount of time between the minimum and maximum values
```

D8H:
For instance:
```
The Sway behavior makes objects sway at different times.

It can be used to:
- Animate grasses in a way that always change and is different from one another.
- Give a hit to player with a grass animation
- Get creative and animate other decorations
```
An example to back each case is a plus.

> The behavior works best if both the *Center* and *Origin* points are at the base of the object.

Is there an action description where this can be added?

> - Set left and right angles (degrees) for the object to sway to
> - For equal sway on each side set the left angle to the negative number of the right angle, eg. set left angle to -5 and right angle to 5
>
> - Set a minimum and maximum Y scale amount for the object to stretch and shrink to
> - Set to 1 for no change, more than 1 for stretching, and less than 1 for shrinking
>
> - Set a range for the angle and Y scale tween times (seconds)
> - The tweens will take a random amount of time between the minimum and maximum values

This look like a list of properties or parameters descriptions. These sentences should be visible where they are useful. But, I bet that a good property name is enough.

Bubble-one:
Great feedback, thank you so much. I'm definitely too wordy, very glad to get your comments.

I've changed the short and long description as well as the description in behavior properties - configuration. 
No other changes.
[Sway.json.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/9730107/Sway.json.zip)

**Configuration description**
was
Sway multiple instances of an object at different times - useful for random grass swaying.
now
`Sway objects (like grass) from side to side with randomized timing.`


**Short description**
was
A behavior to make objects (like grass) sway randomly.
now
`Sway objects (like grass) from side to side with randomized timing.`
(same as config desc)


**Description** (long)
Deleted the intro, how it works and how to use. I think the current names of the actions are descriptive for how it works and what to do. Added examples of how it can be used.

now

```
Use the Sway behavior to make multiple instances of an object sway at different times.

It can be used to:
- Simulate grass and plant movement swaying in the wind
- Animate seaweed in an underwater environment
- Give flames a flickering look 


Control: 
- Left and right rotation
- Y scale up and down
- Range of times for each movement

Set *Center* and *Origin* points to the center base of the object for best results.


Link to a gif showing an example of Sway in use:
[Imgur](https://i.imgur.com/dOyTgbN.gifv)
```





^^^
Thought about how I could include center and origin tip in the actions but couldn't find anywhere.

If I think any extra explanations are necessary I'll manually update the wiki entry. The link format for the gif that works in the description doesn't display correctly in the wiki so I'll update that too.

tristanbob:
Any ideas on what to do with this PR?  

The category was updated recently here: https://github.com/GDevelopApp/GDevelop-extensions/pull/762

D8H:
I changed the descriptions to:

```
Sway objects like grass in the wind.
```

```
This extension makes objects sway at different times.

It can be used to:
- Simulate grass and plant movement swaying in the wind
- Animate seaweed in an underwater environment
- Give flames a flickering look 

Set *Center* and *Origin* points to the center base of the object for best results.

Check out [this demo](https://liluo.io/bubble/swayexamples) to see what this extension can do.
```

Tell me what you think.

PR #635, opened 2022-10-10
New extension: Speed Restrictions

tristanbob:
Limit the maximum movement and rotation speed of an object from forces or the physics behavior.

How to use:
- Add the behavior that matches the method you are using to move the object (forces or physics)
- Set the maximum speed value in the behavior properties

Tips:
- This only affects movement caused by forces or physics behavior
- This may not affect movement caused by other behaviors (platformer, top-down, draggable, etc)

## Playable game

~https://liluo.io/victrisgames/extension-enforce-max-speed~

## Project files

~[Extension - Enforce max speed.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/9742989/Extension.-.Enforce.max.speed.zip)~

## Video

https://user-images.githubusercontent.com/8879811/194789391-1fe7f622-027c-4a6a-bd63-e805e236773b.mp4


D8H:
> I think I'm going to add a Minimum speed to this extension also. This way users can simply add acceleration and deceleration controls and let this extension handle the limits.

An object with a speed = 0 doesn't have any direction. It could make it difficult to do and I guess a minimum speed won't be very useful.

tristanbob:
> > I think I'm going to add a Minimum speed to this extension also. This way users can simply add acceleration and deceleration controls and let this extension handle the limits.
> 
> An object with a speed = 0 doesn't have any direction. It could make it difficult to do and I guess a minimum speed won't be very useful.

I would set up the minimum speed to only apply when the current speed is > 0 (it's the only way to know the direction).

I agree this is not going to be commonly used, but it could be useful in some cases.  For instance, a weapon that pushes back enemies at different rates.  The weapon could stop a small enemy, but the stronger ones would only be slowed (yes, this could be done in other ways).

It will only take me a minute to implement.

tristanbob:
> It will only take me a minute to implement.

Oh, how silly I was.  :)  

I am not going to add MinimumSpeed to this PR, but perhaps I should change the name to SpeedRestrictions so that more restrictions could be added in the future?

tristanbob:
I renamed this to "Speed Restrictions" and added "Max rotation speed".

D8H:
I gave a quick look, it looks good.
A max speed = 0 could be a valid use-case to stop the object.

tristanbob:
> I gave a quick look, it looks good. A max speed = 0 could be a valid use-case to stop the object.

I agree, I removed that condition.

I used it as an optimization for when the user didn't set the max speed, but I think it's more important to have the flexibility you described.

PR #641, opened 2022-10-15
[Reviewed] [MultitouchJoystick] Added conditions for 8-way movement

tristanbob:
Thanks to @jerikothomas!

194347655-e158ec70-fee3-4838-99cc-17a5d5347c5a

## Submission

https://github.com/GDevelopApp/GDevelop-extensions/issues/629

## Project file

[Example_8way_TouchJoystick.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/9792691/Example_8way_TouchJoystick.zip)

## Playable game (touchscreen only)

https://liluo.io/instant-builds/fdc4cde8-ae5b-4e7d-b410-16fd5bc9b1c8

### DIFF
--- a/extensions/reviewed/MultitouchJoystick.json
+++ b/extensions/reviewed/MultitouchJoystick.json
@@ -4,7 +4,7 @@
   "description": "Users can interact with the multitouch joystick to specify angle and force values.  These values can be used to control other objects in the scene such as movement and rotation, such as for twin-stick shooter games.\n\nMulitouch buttons can be used whenever a game allows the user to press multiple locations at once.\n\nHow to use:\n\n- Add the joystick behavior to a sprite that will be the joystick \n- Place the joystick object on the scene\n- Run the \"Activate joystick\" action on every frame and specify the thumb object\n- The joystick thumb object will automatically be created and moved\n\nTips:\n\n- Use \"Simulate a touch\" functions to provide mouse and gamepad controls\n- More than one joystick or button can be used at the same time\n- Joystick and thumb objects should have all sides the same length\n- Thumb object must be smaller than the joystick object",
   "extensionNamespace": "",
   "fullName": "Multitouch joystick and buttons",
-  "helpPath": "",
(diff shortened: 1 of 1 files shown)


D8H:
It could also be implemented with a division:
directionIndex = floor(mod(angle, 360) + 360 / 8 / 2) / (360 / 8)
The directionIndex can be use to look into an array to get the direct string.
It's a lot shorter and less prone to typos.

tristanbob:
> It could also be implemented with a division: directionIndex = floor(mod(angle, 360) + 360 / 8 / 2) / (360 / 8) The directionIndex can be use to look into an array to get the direct string. It's a lot shorter and less prone to typos.

I understand your proposed idea, and it would work great.  But I think it is better to keep this logic expanded for the benefit of readability.  I would not easily be able to decipher that code after some time has passed.

D8H:
> I understand your proposed idea, and it would work great. But I think it is better to keep this logic expanded for the benefit of readability. I would not easily be able to decipher that code after some time has passed.

It can always be commented if needed and hard coded values won't be easier to understand later either. Someone could wonder why -112.5° was chosen where a division make it clear what the logic is.

But, I don't mind if you keep the current implementation. There are some conditions that can be simplified a bit: some bounds are cut in half when it not needed (something like 0 <= x < 3 or 3 <= x < 5).

image

image

tristanbob:
> But, I don't mind if you keep the current implementation. There are some conditions that can be simplified a bit: some bounds are cut in half when it not needed:

Great catch! I have combined these into a single range for both 4-way and 8-way conditions.

I also modified the 4-way movement to use 90-degree arcs (it previously used over-lapping 180-degree arcs).
```
4-way movement = 90 degree arcs (non-overlapping, no missing angles)
8-way movement = 45 degree arcs (non-overlapping, no missing angles)
```
If people need precise values like an analog controller, there is always an expression that provides the exact angle of the joystick.

image

PR #647, opened 2022-10-21
Gamepad - Bug Fix and Enhancement

krunkster:
Bug Fix: The onScenePostEvents was not resetting the pressed state to false for unpressed buttons which made the C_button_released call return true always after releasing a button once.

Enhancement: Updated the onSceneFirstLoaded and onScenePostEvents to keep track of the last activated controller id. Also added a condition to get the last activated controller id.

This makes it possible to dynamically switch between gamepads and just use whichever controller the user picks up instead of having to designate a static controller/player id. Sorry for the whitespace differences.

I'm not a big github user.


krunkster:
@Bouh I wasn't sure this was the correct way to share changes with you, but I hope you will consider them.

Bouh:
This look good!
That mean the **C_button_released** condition return true only once when the button has been released. And it doesn't stay on false.
This is a breaking change in a good way because this mimic how it's done with the key released condition.
Thanks for this fix :+1:

extensions/reviewed/Gamepads.json
```diff
@@ -345,6 +345,26 @@
       ],
       "objectGroups": []
     },
+    {
+      "description": "Return the index of the last last gamepad activated.",
```
Bouh, line 349:
```suggestion
      "description": "Return the index of the last gamepad activated.",
```

Double **last**.
Also during my tests, the expression returned the last one plugged and the last gamepad used by using it.
I'm afraid that **activated** was understood only as the last gamepad plugged?

extensions/reviewed/Gamepads.json
```diff
@@ -345,6 +345,26 @@
...
+      "fullName": "Last activated controller (id)",
+      "functionType": "Expression",
+      "group": "",
+      "name": "LastControllerID",
+      "private": false,
+      "sentence": "",
+      "events": [
+        {
+          "type": "BuiltinCommonInstructions::JsCode",
+          "inlineCode": "const gamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads() : []);\r\n\r\n//Get last activate controller\r\nconst controllerId = gdjs._extensionController.lastActiveController;\r\n\r\n// Check if controller is active\r\nconst gamepad = gamepads[controllerId];\r\nif (gamepad == null) {\r\n    eventsFunctionContext.returnValue = -1;\r\n} else {\r\n    // Return active controller id\r\n    eventsFunctionContext.returnValue = controllerId + 1;\r\n}\r\n",
```
Bouh, line 359:
```suggestion
          "inlineCode": "const gamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads() : []);\r\n\r\n//Get the last activated controller\r\nconst controllerId = gdjs._extensionController.lastActiveController;\r\n\r\n// Check if controller is active\r\nconst gamepad = gamepads[controllerId];\r\nif (gamepad == null) {\r\n    eventsFunctionContext.returnValue = -1;\r\n} else {\r\n    // Return active controller id\r\n    eventsFunctionContext.returnValue = controllerId + 1;\r\n}\r\n",
```

extensions/reviewed/Gamepads.json
```diff
@@ -729,12 +745,11 @@
...
...or.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads() : []);\n\n//Get function parameter\nlet countPlayers = Object.keys(gdjs._extensionController.players).length;\n\n//Repeat for each players\nfor (let i = 0; i < countPlayers; i++) {\n    let gamepad = gamepads[i]; // Get the gamepad of the player\n\n    //we need keep this condition because when use have not yet plug her controller we can't get the controller in the gamepad variable.\n    if (gamepad == null) {\n        continue;\n    }\n\n    for (let b = 0; b < Object.keys(gamepad.buttons).length; b++) { //For each buttons\n        if (gamepad.buttons[b].pressed) { //One of them is pressed\n            gdjs._extensionController.players[i].lastButtonUsed = b; //Save the button pressed\n\n            //Save the state of the button for the next frame.\n            gdjs._extensionController.players[i].previousFrameStateButtons[b] = { pressed: true };\n\n            // Update Last Active Controller\n            gdjs._extensionController.lastActiveController = i;\n        } else {\n            gdjs._extensionController.players[i].previousFrameStateButtons[b] = { pressed: false };\n        }\n    }\n}\n\n\n",
```
Bouh, line 752:
```suggestion
          "inlineCode": "//Each time a player press a button i save the last button pressed for the next frame\n/** @type {Gamepad[]} */\nconst gamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads() : []);\n\n//Get function parameter\nlet countPlayers = Object.keys(gdjs._extensionController.players).length;\n\n//Repeat for each players\nfor (let i = 0; i < countPlayers; i++) {\n    let gamepad = gamepads[i]; // Get the gamepad of the player\n\n    //We have to keep this condition because if the user hasn't plugged in his controller yet, otherwise we can't get the controller in the gamepad variable..\n    if (gamepad == null) {\n        continue;\n    }\n\n    for (let b = 0; b < Object.keys(gamepad.buttons).length; b++) { //For each buttons\n        if (gamepad.buttons[b].pressed) { //One of them is pressed\n            gdjs._extensionController.players[i].lastButtonUsed = b; //Save the button pressed\n\n            //Save the state of the button for the next frame.\n            gdjs._extensionController.players[i].previousFrameStateButtons[b] = { pressed: true };\n\n            // Update Last Active Controller\n            gdjs._extensionController.lastActiveController = i;\n        } else {\n            gdjs._extensionController.players[i].previousFrameStateButtons[b] = { pressed: false };\n        }\n    }\n}\n\n\n",
```

Bouh: Changed the comment you added to:
`We have to keep this condition because if the user hasn't plugged in his controller yet, otherwise we can't get the controller in the gamepad variable.`

Bouh:
Also if you can edit the version from `0.3.0` to `1.0.0`.

extensions/reviewed/Gamepads.json
krunkster: Maybe I should rename the method to be called "Last gamepad used"? The intention was that this would represent the playerId of the gamepad that mostly recently had a button touched.

krunkster:
I think there still might be a lingering bug where calling "Any Gamepad button was released" or "Any gamepad button was pressed" before called "Was button X released?" will prevent getting an accurate response.
There are also rare times where button presses are not detected.
I have not done problem isolation or debugging on either of these, so just letting you know for now.

Bouh:
Indeed I found an issue with released condition.
I fixed it by adding a case in `C_any_button_released` where if the button is released or not I save the appropriate value in the `previousFrameStateButtons` of the button.

You can try it with this example.
Let me know if you see something else.
[gamepad 1.0.0 - example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/9851196/gamepad.1.0.0.-.example.zip)

tristanbob:
This seems like something that should get merged.  

@krunkster were you able to test @Bouh latest update of the example game?

Bouh:
The merge will be made into a cleaner branch here:
https://github.com/GDevelopApp/GDevelop-extensions/pull/692

PR #648, opened 2022-10-31
Fix URLTools attr parameter reading

AlexandreSi:
Resolves https://github.com/4ian/GDevelop/issues/4447

### DIFF
--- a/extensions/reviewed/URLTools.json
+++ b/extensions/reviewed/URLTools.json
@@ -109,7 +109,7 @@
           "disabled": false,
           "folded": false,
           "type": "BuiltinCommonInstructions::JsCode",
-          "inlineCode": "eventsFunctionContext.returnValue =\n    (new URL(eventsFunctionContext.getArgument(\"URL\")))[eventsFunctionContext.getArgument(attr)];\n",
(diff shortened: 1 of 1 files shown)


extensions/reviewed/URLTools.json
```diff
@@ -109,7 +109,7 @@
           "disabled": false,
           "folded": false,
           "type": "BuiltinCommonInstructions::JsCode",
-          "inlineCode": "eventsFunctionContext.returnValue =\n    (new URL(eventsFunctionContext.getArgument(\"URL\")))[eventsFunctionContext.getArgument(attr)];\n",
+          "inlineCode": "eventsFunctionContext.returnValue =\n    (new URL(eventsFunctionContext.getArgument(\"URL\")))[eventsFunctionContext.getArgument(\"attr\")];\n",
```
ClementPasteau, line 112: Should you increase the version number so that users are aware there's a new version?

AlexandreSi: Oh yes you're right!

ClementPasteau (approved):
👍 after version update

PR #649, opened 2022-10-31
Panel sprite button

D8H:
This GDevelop build is required to review this extension:
- https://www.dropbox.com/s/n9ovh2p3h5uhdfo/gdevelop-ebo-beta149-2022-10-31.7z?dl=1

## Example project
- https://www.dropbox.com/s/gbd9bygacvl63dx/EventBasedObjectPrototypeDemo.zip?dl=1

### DIFF
--- a/extensions/reviewed/PanelSpriteButton.json
+++ b/extensions/reviewed/PanelSpriteButton.json
@@ -1,0 +1,1857 @@
+{
+  "author": "",
+  "category": "User interface",
+  "description": "The button can be customized with a background for each state and a label. It handles user interactions and a simple condition can be used to check if it is clicked.",
+  "extensionNamespace": "",
+  "fullName": "Panel sprite button",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


tristanbob:
## Things I like:
- You can easily edit the text inside the button
- Text is automatically centered horizontally and vertically
- Text stays in the right place even when the button moves
- State transitions are handled automatically
 
## Suggestions:
- Provide default button images for each state.  Most people are going to see this:
image

- The new default button images should be high-quality and usable in many games (many users will not change the images).  I don't like the images in your example project, they do not look clean or fun.  @HelperWesley can help find better button images to use.

- I think it could be helpful to add a little more info on the big titles of sections: ```"Text label", "Idle state", "Hovered state", "Pressed state"```

- `Label offset on Y axis when pressed` would make more sense inside the `Pressed state` properties. It's currently the first thing people see, and it's not apparent what it does.

- `Padding` would make sense inside `Text label` properties

- Could I have an option to `Shrink text size to fit inside the button`, `Increase the size of the button to fit text`, `Do not adjust the size of text or button`

- Can you offer more button options?  This is an animation-based button.
How about a "Shape painter button" (each state has its own shape-painter properties)
How about an "Effects-based button" (each state has its own object effect properties)
How about a "Size-based button" (each state has its own scale value and a tween easing with duration)

- The tiled "Repeat borders..." properties are quite imposing to a new user. Can you hide them when the button is unchecked?  (And leave the button unchecked by default)

- `Change the text of the button label` action in the event sheet is double the height expected (the error highlighting is offset a half-line, it seems)

image

D8H:
> Provide default button images for each state.

It's the asset-store role. No object have default assets.

> The new default button images should be high-quality and usable in many games (many users will not change the images).  I don't like the images in your example project, they do not look clean or fun.  @HelperWesley can help find better button images to use.

It's a slightly modified Kenney asset from the rolling pack. I like it because:
- It's plain so it can be used in many context without being wierd.
- It's simple so even if I have little knoledge about drawing, I can edit it to make variants.

But, of couse, alternatives are welcomed.

> I think it could be helpful to add a little more info on the big titles of sections: `"Text label", "Idle state", "Hovered state", "Pressed state"`

I think that "Label" better describe it than "Text label" because it's the whole configuration of the label object not just its text.
I'm not against adding suffixes but it will look repetitive and the precise title of the objects would be something like "Idle state background" which is a bit long.

> Along these lines, each state should have a description field to explain what it is used for and how it is triggered.

It's an interesting idea. No description can be set on child-object but It's something we should add.

> `Label offset on Y axis when pressed` would make more sense inside the `Pressed state` properties. It's currently the first thing people see, and it may not be apparent what it does (I understand what it does)

It can't be moved because parent properties all are at the beginning, but I'll add a description for it.

> `Padding` would make sense inside `Text label` properties

The properties of the label are the property of TextObject. They can't be change just for the button.

> `Increase the size of the button to fit text`

First I did go for this, but I changed it after talking about it with Evaluz and I think she is right. I don't remember every details of the conversation but I think the bets are:
- Users want to set the size in the scene editor and expect the buttons to keep the size they chose.
- Users can choose a button height to have 2 lines if necessary (even if it's probably better to avoid it)

> `Shrink text size to fit inside the button`

It would result to different text sizes on every botton and probably some of them won't be readable.

> Could users select which mouse button will be used to interact with the button?

Anything other than the left-click won't work on mobile so I would prefer not to.

> How about a `Shape painter button` (each state has its own shape-painter properties)

Shape painters can't use assets. We won't know what to display in the scene editor.

> How about an `Effects-based button` (each state has its own object effect properties)

A shader might not be the more efficient way to animate a button.

> How about a `Size-based button` (each state has its own scale value and a tween easing with duration)

Do you have an example of this?
I have the feeling that it would look cheap. I suppose it's an easy way to animate a button and I can understand people choosing to do it, but can it look good?

> The tiled "Repeat borders..." properties are quite imposing to a new user. Can you hide them when the button is unchecked?  (And leave the button unchecked by default)

I'm not sure I understand. How can user check it if it's hidden?
This is a 9-patch object property It can't be removed only for buttons.

> `Change the text of the button label` action in the event sheet is double the height expected (and the error highlighting is offset a half-line, it seems)

This is a bug in GDevelop any string parameter field is this way.

> Conditions:
> 
> * `Check if the button is clicked` => `Check if the button was just clicked.`
> 
> * `Is focused` description is very confusing. I'm not sure what Focused state is.
> 
> * `Check if the button is hovered` => `Check if the mouse cursor is hovered over the button`
> 
> * `Check if the button is pressed.  This...` => `Check if the button is currently being pressed with mouse or touch`

ok

HelperWesley:
https://user-images.githubusercontent.com/72082749/199263712-cd7d1106-68e5-4a8c-94c4-ca5ca649fd46.mp4

Scaling buttons instead of using an animation is pretty common actually, Ball challenge 2 uses it for it's level select.

D8H:
I see. We could maybe do another extension for icon buttons with an optional label a bit like the ones on the top right corner.
What do you think about opening a discussion where we can list every kind of button that could be made an extension?

HelperWesley:
I'm good with that. Realistically we don't need a single button that does everything you'd ever want a button to do.

We could have multiple objects that do different things. Maybe one called "button with animations", and another for "button with scaling", and so on. 🤔

HelperWesley:
My biggest wish for this, is that the label offset and padding values were inside of the label box within the object, so the first thing you saw were label options. That would just make the options a little more obvious, and show the user the thing they care most about first.

We'll of course need alternative button image options for users, and it would be great if we could bundle these with a font that suits the button style. (Pixel art, for example.😅) And ideally, in the future, we could make other types of button objects that work differently.(As discussed above)

____________________________
I did just notice this though. 😅

image

PR #655, opened 2022-11-04
[Reviewed] [Swipe gesture] Detect swipe gestures based on their distance and duration

tristanbob:
Note: This replaces the existing SwipeDetector extension.

## Playable game

https://liluo.io/victrisgames/candy-shake

## Project files

https://github.com/GDevelopApp/GDevelop-examples/pull/400

## Video

https://user-images.githubusercontent.com/8879811/199889100-ef9859a5-dfc2-4ed5-8af7-126d48e41a65.mp4


D8H:
What happens when 2 fingers are used?
It might be interesting to do an example with the pinch extension to see how they work together.

tristanbob:
> What happens when 2 fingers are used? 

Here are the results of my testing:
If a swipe is in progress, the second finger will "steal" the position of the swipe.  
Then the swipe will follow whichever finger is moving. 
If both fingers are moving, the swipe follows the touch that started first.
If either touch is released, the swipe is ended (even if a touch is still in progress).

I did not consider multitouch when I built this extension.  Arthuro made a version that collected data about multitouch, but it only returned data on the most recent touch.  Let me know if you have ideas on how we can make this user-friendly and multitouch.

> It might be interesting to do an example with the pinch extension to see how they work together.

They would be independent of each other.  What are you thinking about?

D8H:
> > It might be interesting to do an example with the pinch extension to see how they work together.
> 
> They would be independent of each other. What are you thinking about?

I was wondering if it will be easy to toggle between the 2 gestures when there is 1 or 2 touches. I'm not sure if we should restrict this extension to one touch only but I think users could use an example that use both extensions together (and it would be an opportunity to check everything is ok).

tristanbob:
> I was wondering if it will be easy to toggle between the 2 gestures when there is 1 or 2 touches. I'm not sure if we should restrict this extension to one touch only but I think users could use an example that use both extensions together (and it would be an opportunity to check everything is ok).

Do we have an expression that returns the number of touches?  If so, we could do:

```
Condition:  Number of touches = 1
Action:
Disable Pinch detection (if there is something like this)
Enable Swipe detection

Condition:  Number of touches > 1
Action:
Disable Swipe detection
Enable Pinch (if there is something like this)
```

D8H:
> > I was wondering if it will be easy to toggle between the 2 gestures when there is 1 or 2 touches. I'm not sure if we should restrict this extension to one touch only but I think users could use an example that use both extensions together (and it would be an opportunity to check everything is ok).
> 
> Do we have an expression that returns the number of touches? If so, we could do:

Maybe we could do a small extension for it.

> BTW, when Swipe detection is disabled, I don't stop the existing swipe, I only stop detecting new swipes. Perhaps I should force the swipe to be stopped?

I think that the Draggable behavior release objects when it's deactivated.

D8H:
I don't know the old extension but this new one seems good overall.

I still have some suggestions:
* I guess it could be interesting to wait the next release (Monday) to use actions with operator without any compatibility to ensure.
* IsSwipeValid: It could be just one event. The group and comments don't help to understand.
* Why do you test the layer variable with `"0"`?
* SetInputButton: I'm not sure it's useful because swiping with a right click is very uncommon I guess.
* I think it's not necessary to keep the previous swipe when a new one starts. `IsSwipeInProgress` and `IsSwipeValid` already allow users to know if they are getting values for a completed swipe or not.
* The timer could be used to avoid the `SwipeInProgress` variable.
* "Start timer" is not a good comment given the action sentence is "Start (or reset) the timer". It doesn't explain anything.
* The distance and angle should probably not be calculated every frame for efficiency reason but maybe it's easier this way. It could be done inside the expression when the swipe is in progress.
* `ShapePainterObject.OutlineSize()` can be used instead of the parameter.
* The shape painter configuration can be changed without any condition it's easier to read and have no impact on efficiency.
* What do you think about "DispacementX" instead of "DistanceX"

Interesting idea to give 4 and 8 way direction conditions.

tristanbob:
> * I guess it could be interesting to wait the next release (Monday) to use actions with operator without any compatibility to ensure.

Good idea. I've converted all the non-boolean expressions to use the new "expression and condition".

I tried to use "action with operator" but it seems to always include `_PARAM0`, so I assume it's only meant to work as a behavior function?  (and not a standalone function)

> * IsSwipeValid: It could be just one event. The group and comments don't help to understand.

This is just a cosmetic preference, but I changed it to a single event.

> * Why do you test the layer variable with `"0"`?

Good question... I'm not sure why I did that.  I have removed this.

> * SetInputButton: I'm not sure it's useful because swiping with a right click is very uncommon I guess.

I agree, I have removed the option to select an input button.  Swiping is mostly done with touch anyway (which is detected as the left button). 

> * I think it's not necessary to keep the previous swipe when a new one starts. `IsSwipeInProgress` and `IsSwipeValid` already allow users to know if they are getting values for a completed swipe or not.

Nice catch, I will remove these unnecessary events.

> * The timer could be used to avoid the `SwipeInProgress` variable.

I don't see how this would work.  I pause the timer when the swipe is done so the swipe duration is available even after the swipe is completed.

> * "Start timer" is not a good comment given the action sentence is "Start (or reset) the timer". It doesn't explain anything.

Ok, I removed that comment. (I'll try not to make overly simple comments).

> * The distance and angle should probably not be calculated every frame for efficiency reason but maybe it's easier this way. It could be done inside the expression when the swipe is in progress.

The previous version did this, but I wasn't sure if it was more optimal. For instance, what if the creator used the expression multiple times?  I can easily change this if we agree it is best.

> * `ShapePainterObject.OutlineSize()` can be used instead of the parameter.

Interesting idea... I have changed to your recommendation and added a note in description for the users.

> * The shape painter configuration can be changed without any condition it's easier to read and have no impact on efficiency.

Really? Do the actions already check if the value needs to be changed?  I thought that actions can have multiple steps so it was good to check if they are needed with a single condition.  In the meantime, I have removed the conditions based on your recommendation.

> * What do you think about "DisplacementX" instead of "DistanceX"

It's a good name, but I don't see a big value in changing it.  More people will understand distance than displacement.

> Interesting idea to give 4 and 8 way direction conditions.

It was easy because I had just fixed/added them on the virtual joystick.  Speaking of which, I still need to do the same to the gamepad controller extension.

D8H:
> > * The timer could be used to avoid the `SwipeInProgress` variable.
> 
> I don't see how this would work. I pause the timer when the swipe is done so the swipe duration is available even after the swipe is completed.

Ok, I didn't see this.

> > * The distance and angle should probably not be calculated every frame for efficiency reason but maybe it's easier this way. It could be done inside the expression when the swipe is in progress.
> 
> The previous version did this, but I wasn't sure if it was more optimal. For instance, what if the creator used the expression multiple times? I can easily change this if we agree it is best.

You can save the result if you want but inside the expression to avoid to do it completely if users don't need it (a value can represent an out of date value). Unless your extension already needs it to check the end.

> > * The shape painter configuration can be changed without any condition it's easier to read and have no impact on efficiency.
> 
> Really? Do the actions already check if the value needs to be changed? I thought that actions can have multiple steps so it was good to check if they are needed with a single condition. In the meantime, I have removed the conditions based on your recommendation.

In this case, it's just like setting a property. A read, a test and a jump is probably less efficient than setting the same value.

tristanbob:
> You can save the result if you want but inside the expression to avoid to do it completely if users don't need it (a value can represent an out of date value). Unless your extension already needs it to check the end.

I tried doing this, but I couldn't get it to work exactly the same way.  The optimization is not a big deal in this case, so I'll leave it in PreStepEvents.

> In this case, it's just like setting a property. A read, a test and a jump is probably less efficient than setting the same value.

That makes sense.

Any other thoughts on this extension?  

My only question is if we should set some default validation properties?  I never found a best practice on defaults for distance and duration of a swipe.  Let m know if anyone has info on this.

D8H:
> > * What do you think about "DisplacementX" instead of "DistanceX"
> 
> It's a good name, but I don't see a big value in changing it. More people will understand distance than displacement.

It can be negative which is unexpected of a distance.

extensions/reviewed/SwipeGesture.json
```diff
@@ -0,0 +1,2423 @@
...
+      ],
+      "expressionType": {
+        "type": "expression"
+      },
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Check if the swipe is valid based on distance and time restrictions (if they are used).",
+      "fullName": "Is the swipe valid",
```
D8H, line 1255:
```suggestion
      "fullName": "Swipe is valid",
```

extensions/reviewed/SwipeGesture.json
```diff
@@ -0,0 +1,2423 @@
...
+            }
+          ]
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Check if a swipe is currently in progress.",
+      "fullName": "Is a swipe in progress",
```
D8H, line 1351:
```suggestion
      "fullName": "Swipe is in progress",
```

extensions/reviewed/SwipeGesture.json
```diff
@@ -0,0 +1,2423 @@
...
+        "type": "expression"
+      },
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Start point of the swipe X position.",
+      "fullName": "Start point of the swipe X position",
+      "functionType": "ExpressionAndCondition",
+      "name": "StartPointX",
```
D8H, line 1061:
```suggestion
      "name": "StartX",
```

extensions/reviewed/SwipeGesture.json
```diff
@@ -0,0 +1,2423 @@
...
+        "type": "expression"
+      },
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Start point of the swipe Y position.",
+      "fullName": "Start point of the swipe Y position",
+      "functionType": "ExpressionAndCondition",
+      "name": "StartPointY",
```
D8H, line 1089:
```suggestion
      "name": "StartY",
```

extensions/reviewed/SwipeGesture.json
```diff
@@ -0,0 +1,2423 @@
...
+        "type": "expression"
+      },
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "End point of the swipe Y position.",
+      "fullName": "End point of the swipe Y position",
+      "functionType": "ExpressionAndCondition",
+      "name": "EndPointY",
```
D8H, line 1145:
```suggestion
      "name": "EndY",
```

extensions/reviewed/SwipeGesture.json
```diff
@@ -0,0 +1,2423 @@
...
+        "type": "expression"
+      },
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "End point of the swipe X position.",
+      "fullName": "End point of the swipe X position",
+      "functionType": "ExpressionAndCondition",
+      "name": "EndPointX",
```
D8H, line 1117:
```suggestion
      "name": "EndX",
```

tristanbob:
>

tristanbob:
> > It's a good name, but I don't see a big value in changing it. More people will understand distance than displacement.
> 
> It can be negative which is unexpected of a distance.

I return the absolute value of the difference in X values, so it cannot be negative. 

> > My only question is if we should set some default validation properties? I never found a best practice on defaults for distance and duration of a swipe. Let m know if anyone has info on this.
> 
> I wonder if the validation is really useful. Users can use the conditions on the duration and distance directly if they need to. It seems easier to understand than an opaque condition (IsValid) where users have to understand which actions to use to configure it. 

I agree that "valid swipe" does not have an obvious meaning.  Is there a better wording that we can use?
In my research on swipe gestures, the minimum distance and maximum duration are often used.
I think we should leave this extension without any default constraints to prevent tripping a user who doesn't understand what it means.

> Also, It would also be easier to add new constraints (on the straightness for instance) as it's just 1 function to add instead of 3.

Can you explain this more?  I'm not sure how to determine straightness and I don't know what the 1 and 3 functions are.

tristanbob:
I think I figured out what you mean.  

- Remove the "is swipe valid" condition
- Users can check the duration, distance, and any future tests simply by adding the extension conditions we provide.

I will remove the "valid" swipe condition.

tristanbob:
FYI - This is ready for another review (no rush).

PR #657, opened 2022-11-07
[Reviewed] [DrawPathfinding] The first path element is now drawn.

D8H:
The bug has almost no impact with the current path finding version but the next one will have longer path elements:
https://liluo.io/games/47f7973b-f3bc-4779-80be-d4baed26fef6?dev=true

It also changes some wording.

Example:
https://www.dropbox.com/s/tv5c0egzltxud69/PathfindingSmoothingTest.zip?dl=1


D8H:
@Entr0py404, do you want to do a review?

Entr0py404:
looks great thank you

PR #658, opened 2022-11-07
Update BoidsMovement Extension

planktonfun:
Fix objects going towards 0,0 by default when the objects are outside boids cohesion radius, simulating a fish sleeping instead


tristanbob (dismissed):
This PR passed my basic functionality tests.

tristanbob:
@planktonfun (Warning: Scope creep, feel free to say no.) 

How do you feel about adding actions that can be used to place an object in a ?

- "Cancel movement from targets" 
- "Cancel movement from obstacles"

I'm not sure if this is the best way to stop movement, so feel free to suggest other wording.

tristanbob (approved):
(Sorry for the false alarm, my bug was caused by having Physics Behavior at the same time.)

This extension passed my functionality tests.
I even tested stopping and starting the behavior, and it worked great.

https://user-images.githubusercontent.com/8879811/200460809-d0c503f1-dfd8-4bf1-9ca5-e200bb14c0ab.mp4

D8H:
I wonder if it would be better to keep the last `cohesionDirection`. As an animal would try to keep up with the group by continuing in the last known direction.

I suggest this:
```JS
    if (count > 0) {
      sumX /= count;
      sumY /= count;

      // A vector pointing from the position to the target
      this.cohesionDirection.set(
        sumX - object.getCenterXInScene(),
        sumY - object.getCenterYInScene()
      )
      this.cohesionDirection.normalize();
    }
    return this.cohesionDirection;
```

planktonfun:
> I wonder if it would be better to keep the last `cohesionDirection`. As an animal would try to keep up with the group by continuing in the last known direction.
> 
> I suggest this:
> 
> ```js
>     if (count > 0) {
>       sumX /= count;
>       sumY /= count;
> 
>       // A vector pointing from the position to the target
>       this.cohesionDirection.set(
>         sumX - object.getCenterXInScene(),
>         sumY - object.getCenterYInScene()
>       )
>       this.cohesionDirection.normalize();
>     }
>     return this.cohesionDirection;
> ```

Tested that works too.

D8H:
Cool, if you are ok with this solution, you can commit it and I will merge.

planktonfun:
Updated

D8H:
The comment must be where I put it in my suggestion. After the normalization, it's no longer true.

tristanbob:
I tested the latest version pushed by @planktonfun and it seems to work great.  

We should get this PR merged ASAP so people don't get frustrated when they try this extension.  

@D8H what do you think?  I didn't understand your last comment.

D8H:
> @D8H what do you think? I didn't understand your last comment.

A comment got misplaced. I would change it myself but I'm not sure I can modify the branch as it's a fork.

D8H (approved):

planktonfun:
> > @D8H what do you think? I didn't understand your last comment.
> 
> A comment got misplaced. I would change it myself but I'm not sure I can modify the branch as it's a fork.

Ohhh, you can directly with github, files changed tab > view file > edit > commit

PR #690, opened 2022-12-10
[Reviewed] [Gamepad] Add behaviors for platformer characters and top-down movement

D8H:
### Examples
- https://github.com/GDevelopApp/GDevelop-examples/pull/414


Bouh:
The list of inputs is limited, we can't put the triggers LT BR RT RB. You should put all the keys. 
image

Bouh:
I was in the Platformer behavior, and I saw this What is the propertie here?
I use the webapp, maybe it's not a version that is not update to date for testing?
image
image


I recommend to use groups, because what mean Use left stick I don't know what it refer, the group name helps:
image

D8H:
I started from the wrong version 😭

extensions/reviewed/Gamepads.json
```diff
@@ -1684,7 +1684,7 @@
                           "parameters": [
                             "",
                             "Object.Behavior::PropertyGamepadIdentifier()",
-                            "\"LEFT\"",
+                            "\"Left\"",
```
Bouh, line 1687: There is two Left, **Right** is missing?

Bouh (approved):

D8H:
I tested the last commit on both example.

PR #693, opened 2022-12-12
[Reviewed] [Multitouch joystick] Add a multitouch controller mapper behavior for platformer characters.

D8H:
### Changes
- Add properties on the stick and button behaviors to identify a controller.
- Add free condition and expression for these controllers (a bit like the gamepad extension).
- Add a behavior to map a controller to a platformer character.

### Example

### DIFF
--- a/extensions/reviewed/MultitouchJoystick.json
+++ b/extensions/reviewed/MultitouchJoystick.json
@@ -50,62 +50,6 @@
   "dependencies": [],
   "eventsFunctions": [
     {
-      "fullName": "",
-      "functionType": "Action",
-      "name": "onFirstSceneLoaded",
-      "sentence": "",
-      "events": [
-        {
-          "disabled": true,
-          "type": "BuiltinCommonInstructions::Standard",
-          "conditions": [
-            {
-              "type": {
-                "inverted": true,
-                "value": "BuiltinCommonInstructions::Always"
-              },
-              "parameters": [
-                ""
-              ]
-            },
-            {
-              "type": {
-                "value": "BuiltinCommonInstructions::CompareNumbers"
-              },
-              "parameters": [
-                "MultitouchJoystick::JoystickForce(0,\"\")",
-                "=",
-                "0"
-              ]
-            },
-            {
-              "type": {
-                "value": "BuiltinCommonInstructions::CompareNumbers"
-              },
-              "parameters": [
-                "MultitouchJoystick::JoystickAngle(0,\"\")",
-                "=",
-                "0"
-              ]
-            },
-            {
-              "type": {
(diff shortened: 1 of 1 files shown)


extensions/reviewed/MultitouchJoystick.json
```diff
@@ -2511,45 +2784,30 @@
...
         },
         {
-          "description": "Simulate a touch on joystick (based on position).  Can be used for mouse controls.",
+          "description": "Simulate a touch on joystick (based on position).  Can be used for mouse controls. This action is deprecated.",
           "fullName": "Simulate a touch on joystick (based on position)",
           "functionType": "Action",
-          "group": "",
           "name": "SimulateTouch_Position",
-          "private": false,
+          "private": true,
```
D8H, line 2810: I hide these actions.
What should they be used for?
I don't understand what "mouse controls" have to do with touch control. Games should hide the virtual joystick on a computer or if a gamepad is connected to the phone so I don't see a usage for it.

tristanbob: My original inspiration was from the platformer behavior that has "Simulate jump key pressed" actions.  
However, I found these actions very helpful when testing touch controls from a PC.

My preference is to leave these actions in the extension (not hidden, not deprecated), but this is not critical to me.

D8H: If this is just for testing. They must be unhidden only during tests to avoid to confuse users.

tristanbob: It is useful for testing the game controls when you are making a game on PC for a mobile device.

D8H: I doubt it's useful to test it this way because it only tests something like 4 events where actions are already shared with keyboard events. What is important to test is playability and it can only be done with the real device.

tristanbob:
Can you please leave the simulate touch actions visible? 

I use these functions inside the game I'm working on now, and if I update this extension then my game would break.
https://liluo.io/victrisgames/marble-race

D8H:
> Can you please leave the simulate touch actions visible?
> 
> I use these functions inside the game I'm working on now, and if I update this extension then my game would break. https://liluo.io/victrisgames/marble-race

The game will still work because the functions are only hidden not deleted.
Can you take a small screenshot of the events that use these functions? Because, I really can't picture how they can be used.

tristanbob:
> Can you take a small screenshot of the events that use these functions?

image

D8H:
I see, it's because the extension doesn't handle the mouse. We can keep the action visible and come back to it later when the mouse/touch system is reworked in the engine.

PR #709, opened 2023-01-02
[Fire Bullet] [Reviewed] Update: Added the ability to specify the layer that bullets are created on

tristanbob:
Test this change using the updated example game:
https://github.com/GDevelopApp/GDevelop-examples/pull/425

### DIFF
--- a/extensions/reviewed/FireBullet.json
+++ b/extensions/reviewed/FireBullet.json
@@ -1,7 +1,6 @@
 {
   "author": "@4ian",
   "category": "Game mechanic",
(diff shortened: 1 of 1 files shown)


D8H (approved):
Do you think we should do another update (a breaking one) to fire on the object layer?

The extension description needs to be sum-up (not mandatory for this update).

tristanbob:
> Do you think we should do another update (a breaking one) to fire on the object layer?

No (unless a lot of people ask for it). At least users can now change the layer with an action.

> The extension description needs to be sum-up (not mandatory for this update).

If we put the detailed info in the description, then it's accessible both in the app and on the wiki (auto-generated).  The description field is currently the easiest way for extension creators to share documentation with the user.

PR #710, opened 2023-01-03
[Follow multiple objects with camera] [Reviewed] Changed calculations to use AABB (bounding boxes)

tristanbob:
Changed the calculations to use AABB instead of trying to estimate the proper width/height.

This problem was noticeable with very long objects, but now it works perfectly!  (Try dragging the long blue objects)

## Playable game
https://liluo.io/games/9afd7a87-e2ac-4598-b387-713f6bbee373

## Project files

[MoveAndZoomCameraToKeepObjectsOnScreen.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/10334026/MoveAndZoomCameraToKeepObjectsOnScreen.zip)

## Video 

https://user-images.githubusercontent.com/8879811/210291535-88f7b148-a544-4763-b123-2dd62f0fae1e.mp4

### DIFF
--- a/extensions/reviewed/FollowObjectsWithCamera.json
+++ b/extensions/reviewed/FollowObjectsWithCamera.json
@@ -1,7 +1,6 @@
 {
   "author": "Jeroen Erkens, Tristan Rhodes (https://victrisgames.itch.io/)",
   "category": "Camera",
-  "description": "This action is meant to be run on every frame (i.e: usually you would use it without conditions, or just conditions picking the appropriate objects to track).\n\n- Use an object group to track multiple objects.\n- Only run on one layer at a time.  If multiple layers need to move the same way, create events that copy CameraX, CameraY, and CameraZoom values.\n- This can also be used on a single object instance to move the camera smoothly.",
   "extensionNamespace": "",
   "fullName": "Follow multiple objects with camera",
   "helpPath": "",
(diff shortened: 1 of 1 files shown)


D8H:
Thanks for updating this extension.

I have some suggestions:
- `__FollowObjects.RightMostObjectWidth` and `__FollowObjects.RightMostObjectHeight` are no longer used.
- `__FollowObjects.MinX` can be calculated with `min()` instead of a condition (it avoids to do a picking).
- What happens after `MinZoom` is reached? Is it usable? I guess the action doesn't do what it says it does. I think this parameter should be removed.
- Could `Layer` and `Camera` be guessed from objects (it can be one of them arbitrarily as if they are from various layers the action is doom to fail anyway).

tristanbob:
> * `__FollowObjects.RightMostObjectWidth` and `__FollowObjects.RightMostObjectHeight` are no longer used.

Good catch, removed.

> * `__FollowObjects.MinX` can be calculated with `min()` instead of a condition (it avoids to do a picking).

Good optimization, I have added it!

> * What happens after `MinZoom` is reached? Is it usable? I guess the action doesn't do what it says it does. I think this parameter should be removed.

Wow, I hadn't thought of that. It currently shows the center of the objects, but there may not be any objects on the screen!  The downside of removing it is that it would renumber parameters (which is a breaking change).  Even so, I'm OK with making the change.

> * Could `Layer` and `Camera` be guessed from objects (it can be one of them arbitrarily as if they are from various layers the action is doom to fail anyway)?
 
Yes, this is a very good idea!

D8H:
> The downside of removing it is that it would renumber parameters (which is a breaking change). Even so, I'm OK with making the change.

To ensure compatibility, you can keep the existing action hidden and make the new action call the old one with `MinZoom = 0`.

D8H:
> > To ensure compatibility, you can keep the existing action, hide it and make the new action call the old one with `MinZoom = 0`.
> 
> That is very interesting because it could allow new and old to work at the same time (similar to how it is done in the core).
> 
> However, I think I'd prefer to leave it with the breaking change because we simplified the parameters (by removing 3). What do you think?

Compatibility is important because users will see it as a bug of the extension if they update it and their game no longer work as expected. Sometimes, it's too hard to unsure compatibility and we have to do a breaking change but in this case it's easy because it's just a function with only one action inside.

tristanbob:
OK, I added the old function back and pointed it to the new function.  How does it look now?

D8H:
May I ask for an example PR to quickly check it out? (I guess, it could even be added to examples).

PR #715, opened 2023-01-08
New extension: Object Slicer

tristanbob:
Slice an object into small fragments that match the color of the original object

## To do

- [x] Pick the name of the extension (I need your ideas)
- [x] Update variables to use the correct name

## Project files

[PixelDestruction.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/10367574/PixelDestruction.zip)

## Playable game

https://liluo.io/victrisgames/object-slicer

## Video

https://user-images.githubusercontent.com/8879811/211177906-106460f6-b725-4e68-9038-8f0bee0c94fe.mp4


tristanbob:
What is the problem with using _readPixels?  I simply copied the functions from the ReadPixel extension.

planktonfun:
> What is the problem with using _readPixels? I simply copied the functions from the ReadPixel extension.

there's a bug in the builder that contradicts the documentation, there's an exception file that needs to be changed for gdjs._methods or
you can also use window._readPixels as a temporary fix

Bubble-one:
Amazing. I love it.


<br>

> *needs a better name... I need your ideas
<br>
If the word 'explode' is in the title, I think it will give people an idea of what to expect. Some examples:

- Pixelated object explosion
- Object pixel explosion
- Exploding object pixelation


words:
pixel, pixelate, pixelation
explode, exploding, explosion
<br>

Also, I'm curious about the readpixels object outline color. What is it for? At first I thought it was going to be the outline color of the exploding pieces (I haven't looked at the project, only the example game).

tristanbob:
This is now ready for a review!  :)

arthuro555:
> there's a bug in the builder that contradicts the documentation

That is, in fact, not a bug, and this is not an exception. It is done exactly for such cases: here, this extension would override the functions of the read pixel extension, which can lead to all kinds of issues if the read extension is modified in the future. The namespace name should be changed here, or even better totally removed and integrated into the JS events of the expressions directly.

---

It might be interesting to take a look at PIXI particles container for the objects, as it is optimized for displaying a lot of simple objects (its how individual particles of a particle emitter are rendered). This could make this much more performant.

D8H:
> > It might be interesting to take a look at PIXI particles container for the objects, as it is optimized for displaying a lot of simple objects (its how individual particles of a particle emitter are rendered). This could make this much more performant.
> 
> That is a neat idea, but I'm not sure if rendering is the biggest slowness in this example. Physics behavior is probably the biggest cause.

It would also mean a 2nd implementation completely in JS. Currently, I guess each fragment is an instance. Events can't move things inside one renderer because 1 instance = 1 renderer.

tristanbob:
@D8H and @arthuro555 does this look correct now?  (I changed the namespace name)

planktonfun:
sounds more like "its not a bug but a feature" excuse, automated build shouldn't require human intervention.
If a developer followed the documentations perfectly there shouldn't be an error, otherwise the documentation is wrong.

D8H:
> @D8H and @arthuro555 does this look correct now? (I changed the namespace name)

It seems ok

PR #746, opened 2023-01-25
 Update extension: Record

Add00:
I accidentally closed the previous [PR ](https://github.com/GDevelopApp/GDevelop-extensions/pull/732)instead of re-basing it. Hopefully this pull request will resolve that issue.


tristanbob:
Since there is a known bug with the existing version, I'm going to approve this.  We can always update it again if we need to.  Thanks for your hard work on this, @Add00!

PR #747, opened 2023-01-25
[Reviewed] [Multitouch joystick] Add choices for the joystick name.

D8H:
* The name can no longer be a custom value.

### Demo
- https://github.com/GDevelopApp/GDevelop-examples/pull/440


4ian:
Thank you!

PR #748, opened 2023-01-25
Make the clipboard extension ✨true async✨

arthuro555:
This adds a new action for the clipboard extension that is an async action. 
Relies on https://github.com/4ian/GDevelop/pull/4819

Additionally:
 - Changed the parameter from a string to an actual variable field on the new action
 - Privated the old one to discourage further use
 - Added `[DEPRECATED]` in front of the original clipboard read expression that did not work cross-platform

### DIFF
--- a/extensions/reviewed/Clipboard.json
+++ b/extensions/reviewed/Clipboard.json
@@ -1,7 +1,6 @@
 {
   "author": "@Bouh, @arthuro555",
   "category": "User interface",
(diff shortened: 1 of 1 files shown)


Bouh:
Once the auto update is enabled for the latest version of the app, we will be good to merge.

Bouh:
We can merge now, the GDeveloppers are almost all in 5.1.157.
Thank you for the update, and the async feature!

PR #764, opened 2023-02-04
DragCameraWithPointer with scroll limitation

fishme:
### Updates

New behaviour for the "DragCameraWithPointer" extension.
Added the possibility to add it for top, bottom, left and right limitation for the dragging of the camera.

### Configuration
Screenshot 2023-02-04 at 10 17 33
Screenshot 2023-02-04 at 10 17 20

### Project files
[Example App](https://github.com/fishme/DragCameraWithPointer_Demo_Limitation)

### DIFF
--- a/extensions/reviewed/DragCameraWithPointer.json
+++ b/extensions/reviewed/DragCameraWithPointer.json
@@ -1,7 +1,6 @@
 {
   "author": "@ddabrahim",
   "category": "Camera",
-  "description": "Move a camera by dragging the mouse (or touchscreen). \n\nHow to use:\n- Run this action on every frame that you want the camera to be movable by the mouse (or touchscreen)\n- Select the mouse button to use (choose \"left\" for touchscreen)\n- Select which directions the camera will move (vertical, horizontal, or both) \n- Select the layer that will be moved\n\nTips:\n- If no parameters are selected, the camera will move in both directions when the left mouse button (or touchscreen) are dragged.",
   "extensionNamespace": "",
   "fullName": "Drag camera with the mouse (or touchscreen)",
   "helpPath": "",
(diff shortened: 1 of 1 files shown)


D8H:

Did you try to call this action after the "drag camera" action? Was there any issue?

image

fishme:
I added into my demo repo an example with your question. There is not an error, but the behaviour isn't different to my change of 

Screenshot 2023-02-05 at 06 15 49


https://github.com/fishme/DragCameraWithPointer_Demo_Limitation
(see scene "without limitation and limitation by object")

fishme:
I added into my example app https://github.com/fishme/DragCameraWithPointer_Demo_Limitation your request. There is no error, it is also working but not so smooth like with my additional behaviour. 

see scene "Enforuce camera boundaries"

D8H:
The action must be in this order to work properly:

image

If someone else want to check @fishme example, here the link to open it with the web-app:
- https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/fishme/DragCameraWithPointer_Demo_Limitation/develop/game.json

>If you want I could also adapt my action to the entry format of "Enforce camera boundaries" so that the entry mask is similar.

1 action that do the same as 2 actions give less freedom to users so I don't think it's necessary.

fishme:
ok, up to you. From my side the full screen is jumping back, with the 2 actions. That is for the UX not good from my perspective. The behaviour for the user is different. ;)
Then I'll keep it in my project.

D8H:
> ok, up to you. From my side the full screen is jumping back, with the 2 actions. That is for the UX not good from my perspective. The behaviour for the user is different. ;) Then I'll keep it in my project.

I don't see any jumping back. Are you sure you put the 2 actions in the right order like in the screenshot of my previous message?

fishme:
ok the order was the issue. well ok works. Then let's remove my PR.

D8H:
I'm closing this PR, but we love to see new contributors, keep it up.

PR #768, opened 2023-02-05
[New Extension] [Reviewed] Joint connector: Create and manage physics joints between two objects

tristanbob:
Weld joint:  Glue two objects together.

Rope joint: Two objects can get closer to each other but they cannot exceed their starting distance.

Distance joint:  Two objects are forced to stay at a set distance from each other.  Adjust the frequency property to change the stiffness of the spring.

Revolute joint:  Two objects are connected but can rotate on the joint.  Angles can be restricted and joints can have motors that spin the objects.

## Playable game

https://liluo.io/victrisgames/extension-joint-connector

## Project files

https://github.com/GDevelopApp/GDevelop-examples/pull/396

## Video

https://user-images.githubusercontent.com/8879811/216799846-c890bc50-caed-45b8-93ca-b5258c6016f1.mp4


tristanbob:
@D8H I'm not sure what is happening here.  I think this is typescript element.  Should it be removed or updated?
```
  disallowedProperty: 'Physics2RuntimeBehavior',
```
```
/** @type {gdjs.Physics2RuntimeBehavior} */
const physicsBehavior = firstObject.getBehavior(behaviorName);
```

tristanbob:
What action(s) do I need to take to make this PR pass CI?

1) Find a way to not use these properties (I need guidance)
2) Add the 3 properties to the list of exceptions (I can do this)
3) Rename the properties so they don't collide with anything else (I need guidance)

tristanbob:
@D8H  I realized that I did not have all the functions to get and set every property.  Tomorrow I'll try using the new condition+expression methods you built. 

Can anyone answer my questions about CI failing?

D8H:
> What action(s) do I need to take to make this PR pass CI?
> 
>     1. Find a way to not use these properties (I need guidance)
> 
>     2. Add the 3 properties to the list of exceptions (I can do this)
> 
>     3. Rename the properties so they don't collide with anything else (I need guidance)

- `Physics2RuntimeBehavior` is fine.
- `evtTools.linkedObjects` should be used instead of `gdjs.LinksManager` if possible.
- I don't know how `gdjs.behaviorsTypes` is used but I guess it should be avoided.

tristanbob:
> * `evtTools.linkedObjects` should be used instead of `gdjs.LinksManager` if possible.

My reference for how to link to objects in Javascript was this forum post:
https://forum.gdevelop.io/t/solved-javascript-link-two-object/29150/6?u=tristanrhodes

I see similar info in the docs: 
https://docs.gdevelop.io/GDJS%20Runtime%20Documentation/classes/gdjs.LinksManager.html

I did find the method you recommended, but I'm not sure how to use it (and can't find examples): 
https://docs.gdevelop.io/GDJS%20Runtime%20Documentation/modules/gdjs.evtTools.linkedObjects.html#linkObjects

> * I don't know how `gdjs.behaviorsTypes` is used but I guess it should be avoided.

I have no idea what it is used for, but removing it breaks the extension. Any hints on how to replace it?

extensions/community/JointConnector.json
```diff
@@ -0,0 +1,2029 @@
...
+      "sentence": "Connect overlapping _PARAM1_ with a revolute joint located at the center of _PARAM3_",
+      "events": [
+        {
+          "type": "BuiltinCommonInstructions::JsCode",
+          "inlineCode": [
+            "const getProperty = function (propertyName) {",
+            "    return this._behaviorData[propertyName];",
+            "}",
+            "",
+            "gdjs.behaviorsTypes.items['JointConnector::RevoluteJoint'].prototype.getProperty = getProperty;",
```
arthuro555, line 62: This is not a very good idea...

- This changes the protoype which is not great optimisation-wise
- this._behaviorData would not exist anymore if we were to mark it as private in the source file, as the minifier would be able to minify it if it knows nothing outside should be able to access it (and it should not be)
- Generally, this accesses an _ marked property (`_behaviorData`), which is completely disallowed by the extensions best practices (not, adding an indirection through a function call on the prototype does not make this magically OK)

Instead, you can use `jointBehavior._get<PropertyName>()`, e.g. `jointBehavior._getMotorEnabled()`. Those are the only exceptions to the `_` marked variables, as they are marked as such to tell only _users of the extension & creators of other extensions_ not to use it, but the extension that makes the behavior is still allowed to access it, as the naming convention is otherwise stable and guaranteed.

tristanbob: Thank you very much for helping me; I am truly out of my depth at this point.  Here is what I tried (it seems to work until the last line):
```
const behaviorName = eventsFunctionContext.getBehaviorName("Physics");
const jointObjects = eventsFunctionContext.getObjects("Joint");
const linkManager = gdjs.LinksManager.getManager(runtimeScene);

for (const jointObject of jointObjects) {
    // Save properties of each joint
    const jointBehavior = eventsFunctionContext.getBehaviorName("RevoluteJoint");
   // The line below causes "not a function" error
    const enableLimit = jointBehavior._getPropertyAngleLimitsEnabled();
```
Error:
```
Uncaught TypeError: jointBehavior._getPropertyAngleLimitsEnabled is not a function
```

Can you provide an example of the *right* way to do this?  Then I will be able to replicate it on all functions of this extension.

arthuro555: That was almost it, you only needed to remove the `Property` part. I went ahead and pushed the correction for you. I also auto-formatted the white-spaces (Shift-Alt-F in the code editor 😉) and removed a useless and likely forgotten console.log.

tristanbob:
Thank you to everyone who helped make this a reality:

@D8H
@planktonfun
@arthuro555

Your contributions on this extension mean a lot to me.  :)

PR #787, opened 2023-02-25
[Community] [Dynamic water] Fixes bug when multiple water objects are used. Thanks @Alios5!

tristanbob:
@Alios5 let me know if you want any additional changes to your extension.

I would love to get a short list of the changes if you can provide it.


Alios5:
I fixed a bug that creates a confusion when you create several water area, at the level of the period of the waves because it was managed by only one variable of scene for all the dynamic water object

PR #797, opened 2023-03-11
[Community] [Top-down movement animator] Change the animation according to the top-down movement direction

D8H:
### Demo
- https://github.com/GDevelopApp/GDevelop-examples/pull/485
- https://github.com/GDevelopApp/GDevelop-examples/pull/483
- https://github.com/GDevelopApp/GDevelop-examples/pull/484


extensions/reviewed/TopDownMovementAnimator.json
```diff
@@ -0,0 +1,898 @@
...
+          "group": "Animation",
+          "extraInformation": [],
+          "hidden": false,
+          "name": "IsScalingAnimation"
+        },
+        {
+          "value": "Walk",
+          "type": "String",
+          "label": "Animation name",
+          "description": "Animations must be called \"Walk0\", \"Walk1\"... for left, down...",
```
AlexandreSi, line 834: Could we add more details here? I think it has to be exhaustive to be really helpful

extensions/reviewed/TopDownMovementAnimator.json
```diff
@@ -0,0 +1,898 @@
...
+              "name": "Behavior",
+              "supplementaryInformation": "TopDownMovementAnimator::TopDownMovementAnimator",
+              "type": "behavior"
+            }
+          ],
+          "objectGroups": []
+        },
+        {
+          "description": "Check if the animation is scaled according to speed.",
+          "fullName": "Scale animation according to speed",
```
AlexandreSi, line 433: I don't understand what this means

AlexandreSi:
image
Looking at this, I wonder:
- What is the top-down movement selector? (I don't know if we can hide it since I think there is no use in changing it for the user)
- Comparing to the platformer character animator:
image

  I understand why you did it like that (easier with numbers I guess) but it seems a bit obscure from a non-programmer point of view I would say. Do you think it could be possible to have a similar interface as the platformer one?

extensions/reviewed/TopDownMovementAnimator.json
D8H, line 834:
```suggestion
          "description": "Animations must be called \"Walk0\", \"Walk1\"... for right, down...",
```

D8H: It gives the start and the direction. I think giving more won't help. Especially because there are not necessarily only 4 directions.

extensions/reviewed/TopDownMovementAnimator.json
D8H: It use this action
image

D8H:
> Looking at this, I wonder: - What is the top-down movement selector? (I don't know if we can hide it since I think there is no use in changing it for the user) - Comparing to the platformer character animator:

It's a required behavior. It allows users to understand why this extension was added automatically and why they can't remove it while keeping this behavior.

> I understand why you did it like that (easier with numbers I guess) but it seems a bit obscure from a non-programmer point of view I would say. Do you think it could be possible to have a similar interface as the platformer one?

It's not possible because 8 is not a hard limit. Analog sticks allow any direction so games could need 64 animations.
There is the same issue with path finding where more that 8 directions is the canonical case and I think it's best to have the same interface for both.

It's not easy to understand by just reading the description but looking at an example or a tutorial make it obvious.

D8H:
In the case of this extension the movement angle is used which is kind of the same as using the acceleration vector. Characters rotate as soon as a key is pressed even if there inertial make them move in the opposite direction.

For other movements with indirect player control, the speed can be used. We could probably do a generic extension that is based on the change of position. It would make teleportation less intuitive though so it may not be a good idea. But, in another hand there is a lot of movement behaviors so a generic behavior could be interesting.

An animator behavior can be useful with:
- Pathfinding
- Forces
- Phisics2
- Predetermined paths
  - Curved movement
  - Ellipse movement
  - Rectangular movement
  - Boids
  - ...

D8H:
I'm merging it in the community list as it's a question that comes back from time to time.

PR #807, opened 2023-03-19
Extension proposition : Authorized Platform Validation

Oxey405:
This is a small extension made to check where a game exported to web is executed to insure it is not being published and played on an unauthorized platform.
Some platforms scraps websites such as itch.io (and gd.games) in the search to republish small but relatively popular games on their website without the owner's content and then display ads on the page to make profit out of a game they did not even have the right to publish !
This extension allows for gamedevs to define with events which platforms are allowed and can use the "The game is running on a valid context" conditon to act accordingly for example, displaying a warning message or block the game entirely.

### DIFF
--- a/extensions/reviewed/AuthorizedPlatformsValidation.json
+++ b/extensions/reviewed/AuthorizedPlatformsValidation.json
@@ -2,10 +2,10 @@
   "author": "",
   "category": "Network",
   "extensionNamespace": "",
-  "fullName": "Validate Execution Context",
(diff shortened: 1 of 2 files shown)


Oxey405 (commented):
Rereading my own extension I realize I need to specify a namespace and it needs a bit more of housekeeping... I'll let the rest of the team know when these changes happen but I think you could expect them for tomorrow

arthuro555 (approved):

PR #828, opened 2023-03-24
[Reviewed] [Gamepad] Handle advanced vibrations

Bouh:
Update the Gamepads extension, from 0.4.1 to **0.5.0**.

This add new features:
- A new action: Advanced gamepad vibration which use the strong and weak magnitude.
- A new action to edit the magnitude in real time.
- An expression to get the **strong** magnitude.
- An expression to get the **weak** magnitude.

Thanks to @lhphr & @ZachjuKamashi

The project example with the extension:
[Extension.Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/11063356/Extension.Example.zip)


ZachjuKamashi:
I also just noticed a typo for the condition that checks which button was pressed. "BAck" should be "Back" instead
Screenshot (273)

Bouh:
Right, nice catch!
I made the changes you asked for.

extensions/reviewed/Gamepads.json
```diff
@@ -1672,6 +1841,48 @@
...
+          "useStrict": true,
+          "eventsSheetExpanded": false
+        }
+      ],
+      "expressionType": {
+        "type": "expression"
+      },
+      "parameters": [],
+      "objectGroups": []
     }
```
D8H, line 1886: Why is this expression shared by all gamepad ?

Bouh: I think there is no need to make these values unique for each players.

extensions/reviewed/Gamepads.json
```diff
@@ -1390,12 +1387,171 @@
...
+            "    duration: duration * 1000,",
+            "    weakMagnitude: weakRumbleMagnitude,",
+            "    strongMagnitude: strongRumbleMagnitude",
+            "  });",
+            "}",
+            "",
+            "gdjs._extensionController.rumble.duration = duration;",
+            "gdjs._extensionController.rumble.elapsedTime = 0;",
+            "gdjs._extensionController.rumble.weakMagnitude = weakRumbleMagnitude;",
+            "gdjs._extensionController.rumble.strongMagnitude = strongRumbleMagnitude;"
```
D8H, line 1451: These attributes are affected in an action that focus on 1 gamepad, but they override the same attribute for any gamepad. It should use an array with `playerId` as index.

Bouh:
I made the edit I hope it's what you talking about.
The extension is still functional.
I don't have two controllers so I couldn't test it.

extensions/reviewed/Gamepads.json
```diff
@@ -1390,12 +1387,171 @@
...
+          "type": "BuiltinCommonInstructions::JsCode",
+          "inlineCode": [
+            "/** @type {Gamepad[]} */",
+            "//Vibration work only on game in browser.",
+            "const gamepads = navigator.getGamepads ? navigator.getGamepads() : (navigator.webkitGetGamepads ? navigator.webkitGetGamepads() : []);",
+            "",
+            "//Get function parameters",
+            "const playerId = eventsFunctionContext.getArgument(\"Player_ID\") - 1;",
+            "const elapsedTime = gdjs._extensionController.rumble.elapsedTime || 0;",
+            "const originalDuration = gdjs._extensionController.rumble.duration || 1;",
```
D8H, line 1499: Should the playerId used here too?

D8H (approved):
I didn't tested it, but it LGTM.

Bouh:
Thanks!

@lhphr & @ZachjuKamashi if you have a feedback or not let me know, then I merge your new functionalities.

extensions/reviewed/Gamepads.json
```diff
@@ -1465,6 +1629,12 @@
             "            16: \"PS_BUTTON\",\r",
             "            17: \"CLICK_TOUCHPAD\"\r",
             "        }\r",
+            "    },\r",
+            "    rumble: {\r",
+            "        duration: 0,\r",
+            "        elapsedTime: 0,\r",
+            "        weakMagnitude: 0,\r",
+            "        strongMagnitude: 0\r",
```
lhphr, line 1637: Since the rumble was moved into `gdjs._extensionController.players`, this is now redundant.
Other than that, the code seems good enough to me. For now, it's probably best to wait for ZachjuKamashi to give feedback on whether or not the rumble works as expected with multiple controllers.

ZachjuKamashi:
Alright I've tested it with two controllers and I discovered no issues. Both controller can be vibrated separately and changed separately. I discovered no issues with expressions either so I think it's good to go! 👍

Bouh:
Awesome, thanks everyone!
This is now merged, and is available in few minutes in the app.
Again thank you for these new functionalities and tests.

I taking care of updating examples with this new version.

PR #832, opened 2023-03-26
[Reviewed] [Inventories] A rewrite of the extension with events

D8H:
It has the same features of the one integrated to GDevelop but it's written with events so users can modify it if they need to.

It requires:
- https://github.com/4ian/GDevelop/pull/5220

### Demo
- https://github.com/GDevelopApp/GDevelop-examples/pull/500

### DIFF
--- a/extensions/reviewed/Inventories.json
+++ b/extensions/reviewed/Inventories.json
@@ -692,13 +692,11 @@
           "actions": [
             {
               "type": {
-                "value": "Inventories::CopySceneVariable"
+                "value": "CopyVariableToArgument"
               },
               "parameters": [
-                "",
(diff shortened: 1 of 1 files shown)


4ian:
Very nice 🔥🔥🔥

Bouh:
Note there is an existing extension named `Extra inventory functions` that could be merge with this update of the Inventory extension.
image

Bouh (approved):
After a review of the example of the extension of this PR, you added the functions from the other extension. 👍 

**The [example](https://github.com/GDevelopApp/GDevelop-examples/pull/500) and the extension LGTM.**

PR #844, opened 2023-04-03
[Reviewed] [AdvancedProjectile] Add "AdvancedProjectile" extension

tristanbob:
## Advanced Projectile
- Control how a projectile object moves including speed, acceleration, distance, and lifetime.

## Playable game
https://gd.games/victrisgames/extension-advanced-projectile

## Video
https://user-images.githubusercontent.com/8879811/229410580-9b90add5-2602-46f7-989b-63f16fa57a32.mp4

See the example game PR:
https://github.com/GDevelopApp/GDevelop-examples/pull/506

### DIFF
--- a/extensions/reviewed/AdvancedProjectile.json
+++ b/extensions/reviewed/AdvancedProjectile.json
@@ -1,0 +1,1279 @@
+{
+  "author": "",
+  "category": "Movement",
+  "extensionNamespace": "",
+  "fullName": "Advanced projectile",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


D8H:
> ## PreEvents
> 
> * Check if "Distance from start" or "Lifetime" have been exceeded.  If so, mark for deletion (with a boolean object variable)
> 
> ## PostEvents
>
> * Delete objects marked for deletion (this allows conditions "Is lifetime/distance exceeded" to detect when a limit is exceeded)

Sounds good (but it probably don't need a boolean).

> * Initialize object (starting speed, timer, etc).  This intentionally overrides existing speed (such as from "Fire Bullet" action)

It can't be done in onCreate because fire bullet adds a "force"? Is it a one time thing?

> * Enforce speed limits. If a user applies additional forces in events, we want to fix speed before rendering movement.

I guess if users add instant "forces", they want something very specific and the extension should let them do it. If they do, they should probably not use the extension anyway so I think the extension should not do a special case for it.

tristanbob:
@D8H Thanks for the help in chat!  I have made several improvements:

- Removed unnecessary boolean variable to check if the object exceeded lifetime/distance values
- Simplified min/max speed logic
- Added TimeDelta() to acceleration
- Use existing movement angle when the object is moving, otherwise use the direction the object is facing
- Added tips for users in description

tristanbob:
@D8H I've thought more about what you said. Do you think it would be best to remove the "Delete object" functionality from this extension?  It will allow creators to decide what happens when the Lifetime or Distance values are exceeded.

D8H:
> @D8H I've thought more about what you said. Do you think it would be best to remove the "Delete object" functionality from this extension? It will allow creators to decide what happens when the Lifetime or Distance values are exceeded.

For a projectile, it can make sense to delete it automatically but maybe a condition on the Lifetime or Distance values is easier to understand than a "Will be deleted this frame" condition. I don't know, I'm 50-50 on this.

tristanbob:
> > @D8H I've thought more about what you said. Do you think it would be best to remove the "Delete object" functionality from this extension? It will allow creators to decide what happens when the Lifetime or Distance values are exceeded.
> 
> For a projectile, it can make sense to delete it automatically but maybe a condition on the Lifetime or Distance values is easier to understand than a "Will be deleted this frame" condition. I don't know, I'm 50-50 on this.

I already have two conditions for "Lifetime exceeded" and "Max distance (from starting position) exceeded".

I think I will add two Boolean properties:
- Delete when lifetime is exceeded (default: yes)
- Delete when max distance is exceeded (default: yes)

I'm also going to add two helper functions
- Restart Lifetime timer
- Set starting position to current position

This should provide a wide variety of use cases, but with default settings, the objects will be deleted when exceeding limits.

PR #872, opened 2023-05-08
Add "Start dragging object" to enable custom picking of object to be dragged

tristanbob:
Default dragging requires the cursor to be over an object. This is not great for fabrics of small objects, because there is a lot of space that will not trigger a drag to start.  This change adds a new action to manually start dragging, so game creators can choose when and which objects to grab.

To enable this logic, I added a new option to the Mouse Button property: "None (disable automatic dragging)".
Do you like this implementation, or should I make a separate property, such as: "Enable automatic dragging"?

image

## Example game

https://gd.games/victrisgames/building-monsters

### DIFF
--- a/extensions/reviewed/DraggablePhysics.json
+++ b/extensions/reviewed/DraggablePhysics.json
@@ -106,8 +106,7 @@
                     "textG": 0,
                     "textR": 0
                   },
-                  "comment": "Create mouse joints and save the joint ID",
-                  "comment2": ""
+                  "comment": "Create mouse joints and save the joint ID"
                 },
                 {
                   "type": "BuiltinCommonInstructions::Standard",
@@ -214,8 +213,7 @@
                     "textG": 0,
                     "textR": 0
                   },
-                  "comment": "Delete the mouse joint",
-                  "comment2": ""
+                  "comment": "Delete the mouse joint"
                 },
                 {
                   "type": "BuiltinCommonInstructions::Standard",
@@ -262,8 +260,7 @@
                     "textG": 0,
                     "textR": 0
                   },
-                  "comment": "Update the position of the mouse joint to follow the mouse",
-                  "comment2": ""
+                  "comment": "Update the position of the mouse joint to follow the mouse"
                 },
                 {
                   "type": "BuiltinCommonInstructions::Standard",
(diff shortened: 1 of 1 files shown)


tristanbob:
Note: I'm repeating this question so it is more obvious that I am seeking feedback.

I added a new option to the `Mouse Button` property: `None (disable automatic dragging)`
Do you like this implementation, or should I make a separate property, such as: `Enable/Disable automatic dragging`?

tristanbob:
Any feedback on my question?  Which is the best implementation:

- Add a new option to the `Mouse Button` property: `None (disable automatic dragging)`
OR
- Add new property `Enable/Disable automatic dragging`

D8H:
> Any feedback on my question? Which is the best implementation:
> 
>     * Add a new option to the `Mouse Button` property: `None (disable automatic dragging)`
>       OR
> 
>     * Add new property `Enable/Disable automatic dragging`

I'm not sure the 1st option is possible because the "mouse button" parameter type can't have this new choice in the action to change the property.

tristanbob:
> > Any feedback on my question? Which is the best implementation:
> 
> > 
> 
> >     * Add a new option to the `Mouse Button` property: `None (disable automatic dragging)`
> 
> >       OR
> 
> > 
> 
> >     * Add new property `Enable/Disable automatic dragging`
> 
> 
> 
> I'm not sure the 1st option is possible because the "mouse button" parameter type can't have this new choice in the action to change the property.

This is actually what my commit does. I'm using a "list of strings".

D8H:
> This is actually what my commit does. I'm using a "list of strings".

Yes, but, ideally the "mouse button" parameter type should be used because the editor can make a better UI with translations and icons for instance.
The same types could be added to properties in the future as it seems useful for input mapping behaviors.

tristanbob:
> > This is actually what my commit does. I'm using a "list of strings".
> 
> Yes, but, ideally the "mouse button" parameter type should be used because the editor can make a better UI with translations and icons for instance. The same types could be added to properties in the future as it seems useful for input mapping behaviors.

OK, I think what you are saying is:

1) There are no current "mouse button" parameter types, but in the future, there might be.
2) Changing the list of strings to match the standard "mouse button" format will make it easier to transition to the above.
3) Therefore, I should create a new property to enable/disable automatic dragging 

Did I understand correctly?

image

tristanbob:
I've changed this to a new property and removed the option on the mouse buttons.

Next, I will create an update to the existing example game for "draggable physics" to test this change.

tristanbob:
I tested this on the default example, but I couldn't think of a good way to add it to the game that made sense.

PR #873, opened 2023-05-08
Update GitHub Action

st-Wook:
1. Bumps [actions/checkout](https://github.com/actions/checkout) from v2 to v3
2. Bumps [peter-evans/create-pull-request](https://github.com/peter-evans/create-pull-request) from v4 to v5
3. Bumps [xt0rted/pull-request-comment-branch](https://github.com/xt0rted/pull-request-comment-branch) from v1 to v2

### DIFF
--- a/.github/workflows/auto-pr.yml
+++ b/.github/workflows/auto-pr.yml
@@ -113,7 +113,7 @@ jobs:
 
       - name: Create Pull Request
         id: cpr
-        uses: peter-evans/create-pull-request@v4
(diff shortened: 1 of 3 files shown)


st-Wook:
What do you think about including npm dependency caching?

ClementPasteau:
Thanks for this PR @st-Wook !

Tbh, the whole CI takes only 30 seconds, so even if I think it's a good idea it's probably not required yet

PR #889, opened 2023-05-19
[Reviewed] [Mouse pointer lock] Add handling of touches

D8H:
Listen `pointermove` instead of `mousemove` to make it work on with touch events too.


D8H:
@PANDAKO-GitHub Do you think this change is ok?

PANDAKO-GitHub:
Yes, I think it's fine.

PR #906, opened 2023-06-08
Upgrade in-app-purchase version for Google Play Billing v5

ClementPasteau:
Not tested yet, should allow upgrading to Google Play Billing v5 (see https://github.com/j3k0/cordova-plugin-purchase/pull/1308)

Should test with example: https://github.com/GDevelopApp/GDevelop-examples/tree/main/examples/in-app-purchase


tristanbob:
Here is some more info about this required change:

Google Play Billing Library version deprecation
https://developer.android.com/google/play/billing/deprecation-faq

selected-pixel-jameson:
Has this been implemented? We are now required to use Billing Library V5 or higher?

ClementPasteau:
See #1017

PR #930, opened 2023-07-10
Add behavior and object headers

D8H:
It add extension behavior and objet in the header to allow the IDE to list behaviors instead of extensions.

### DIFF
--- a/scripts/types.d.ts
+++ b/scripts/types.d.ts
@@ -111,7 +111,7 @@ export interface EventsBasedBehaviors {
   fullName: string;
   name: string;
   objectType: string;
-  private: boolean;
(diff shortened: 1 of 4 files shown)


scripts/types.d.ts
```diff
@@ -1,41 +1,89 @@
...
   category: string;
   previewIconUrl: string;
 }
 
+type ExtensionTier = 'community' | 'reviewed';
+
+/**
+ * An extension, behavior or object.
+ */
+export interface RegisteryItem extends ItemExtensionHeaderFields {
```
ClementPasteau, line 16: I believe there is a typo here, this should be Registry

PR #934, opened 2023-07-13
Better top behavior list handling

D8H:


scripts/generate-extensions-registry.js
```diff
@@ -288,11 +288,17 @@ const readExtensionsFromFolder = async (folderPath, tier) => {
       extensionShortHeaders,
       behavior: {
         headers: behaviorShortHeaders,
+        firstIds: views.default.firstBehaviorIds,
       },
       object: {
         headers: objectShortHeaders,
+        firstIds: views.default.firstObjectIds,
```
4ian, line 295: I wonder if it's worth adding `views: { default: { ` to mimic what was done for extensions, and maybe one day have a different list for another "view" (like a view for some kind of games, etc...)

4ian (approved):

scripts/generate-extensions-registry.js
D8H: ok, it makes sense, I didn't understand what `views` meant.

PR #935, opened 2023-07-14
Improve Automatic PR Creation GitHub Action and Command GitHub Action.

st-Wook:
1. Fixed the issue where error comments occur when a comment is posted during the upload of a zip file. (It seems that the action will again trigger when the upload is complete.) ([problem ran action](https://github.com/GDevelopApp/GDevelop-extensions/actions/runs/5536079994/jobs/10103262105))

<details>
  <summary> Run info </summary>

  ```bash
  curl -L $(echo $BODY | grep -ioEh 'https?://\S+.zip' -m1) -o /tmp/ext.zip
  unzip -o /tmp/ext.zip -d ./extensions/community
  shell: /usr/bin/bash -e {0}
  env:
    BODY: I've made the requested fixes and uploaded the updated extension file. Please review the file and make sure that it contains the changes you expect.
  
  !update
  [Uploading CountdownTimer.json.zip…]()
  ```
  
 </details>

2. Added a clearer error comment for when there are no changes in Git.
3. Fix small typo
4. Implement example file validation (Cannot update example)

### DIFF
--- a/.github/workflows/auto-pr.yml
+++ b/.github/workflows/auto-pr.yml
@@ -1,7 +1,7 @@
 name: Extension issue to Pull Request conversion
 on:
   issues:
-    types: [opened, labeled, edited]
+    types: [labeled, edited]
 
 jobs:
   generate-pr:
--- a/.github/workflows/commands.yml
+++ b/.github/workflows/commands.yml
@@ -69,9 +69,20 @@ jobs:
           ref: ${{ steps.comment-branch.outputs.head_ref }}
 
       - name: Download extension
+        id: download-extension
         env:
           BODY: ${{ github.event.comment.body }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          FILEURL=$(echo $BODY | grep -ioEh 'https?://\S+.zip' -m1)
+          FILENAME=$(basename $FILEURL)
+
+          if [[ "${FILENAME,,}" == *"sample"* ]]; then
+            echo "sample=true" >> $GITHUB_OUTPUT
+            gh pr comment ${{ github.event.issue.number }} --body "✅ Cannot update the sample via command. Please feel free to upload the updated sample zip file in a comment!"
+            exit 0
+          fi
+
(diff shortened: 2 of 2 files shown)


.github/workflows/auto-pr.yml
```diff
@@ -1,7 +1,7 @@
 name: Extension issue to Pull Request conversion
 on:
   issues:
-    types: [opened, labeled, edited]
+    types: [labeled, edited]
```
st-Wook, line 4: Fixing the issue where GitHub Actions are triggered at once with 'opened' and 'labeled' conditions when an issue is opened.

.github/workflows/commands.yml
```diff
@@ -1,5 +1,8 @@
 name: GDevelop Extensions PR commands
-on: issue_comment
+on:
+  issue_comment:
+    types: [created, edited]
+
```
st-Wook, line 5: Specify the type to prevent triggering when a comment is deleted.

.github/workflows/auto-pr.yml
st-Wook: The issue arises when the '✨ New extension' label is automatically applied as the issue is opened.

arthuro555 (approved):

.github/workflows/commands.yml
```diff
@@ -66,10 +69,21 @@ jobs:
...
+          FILEURL=$(echo $BODY | grep -ioEh 'https?://\S+.zip' -m1)
+          FILENAME=$(basename $FILEURL)
+
+          if [[ "${FILENAME,,}" == *"example"* ]]; then
+            echo "example=true" >> $GITHUB_OUTPUT
+            gh pr comment ${{ github.event.issue.number }} --body "✅ Cannot update the example via command. Please feel free to upload the updated example zip file in a comment!"
+            exit 0
+          fi
+
+          curl -L $FILEURL -o /tmp/ext.zip
```
arthuro555, line 86: This check seems a bit fragile, it could still allow for an example to be given or for a valid extension to be rejected, let's reverse it. 

Instead, we should try to refactor the update command to use the extract/verify extension functions used by the auto-pr workflow, since these provide a robust check against using an example file since #1034

PR #939, opened 2023-07-17
[FPS] Update: Provide more accurate values for frames per second

tristanbob:
- Uses a sliding window to show exactly the number of frames in the previous second
- Backwards compatible with the previous extension

## Playable game

https://gd.games/victrisgames/load-testing

## Game example

https://github.com/GDevelopApp/GDevelop-examples/pull/544

## Video

https://github.com/GDevelopApp/GDevelop-extensions/assets/8879811/fb5d5953-b3fe-48f2-887b-047cbada839d

### DIFF
--- a/extensions/reviewed/FPS.json
+++ b/extensions/reviewed/FPS.json
@@ -10,10 +10,12 @@
   "shortDescription": "Calculate and display the frames per second (FPS) of the game.",
   "version": "1.1.0",
   "description": [
-    "Provides an expression to get the current FPS and a behavior to simplify displaying the current FPS.",
+    "Provides an expression to get the current FPS and a text object behavior to display the current FPS.",
     "",
     "Frames Per Second (FPS) describes how many times in the last second your game logic was executed. ",
-    "This includes running behaviors and events, and then rendering the new game state into the game window. The higher the FPS, the more fluid and performant your game looks."
+    "This includes running behaviors and events, and then rendering the new game state into the game window. The higher the FPS, the more fluid and performant your game looks.",
+    "",
+    "Note: Use the Performance Profiler (inside in the Debugger) for detailed performance information about your game."
   ],
   "origin": {
     "identifier": "FPS",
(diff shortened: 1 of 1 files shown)


extensions/reviewed/FPS.json
```diff
@@ -19,82 +28,263 @@
...
+              ]
+            }
+          ]
+        }
+      ],
+      "parameters": [],
+      "objectGroups": []
+    },
+    {
+      "description": "Average frames per second (FPS) during the last update period.",
```
D8H, line 147:
```suggestion
      "description": "Average frames per second (FPS) during the last update period. When measuring a game performance, rendering time should be used instead of FPS because going from 100 to 80 FPS is 2,5 ms difference where going from 40 to 20 FPS is 25 ms difference.
```

arthuro555:
Hmmm I think the implementation could be done better. The UpdatePeriod is bad since the FPS expression does not show the current FPS in real time but the last measured one among other things.

Instead, you could push a current timestamp onto an array every frame, and pop all the timestamps more than 1 second old from the array. That way, the array always contains the list of frames captured within the last second, and you just need to get the length of the array to know the amount of frames within the last second.

For correctness sake, a frame should also not be counted before it has fully finished. when onStepPreEvents runs, there are still objects, behaviors, and rendering that may run and are not counted as part of the frame. GDevelop does not offer a way to run code at the end of the events loop, but luckily we can use JavaScript for that. [`queueMicrotask`](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask) will run a function when GDevelop's javascript has finished doing all that it is doing to render a frame and gives control back to the browser. That's the moment where we should record a frame, preferably with [`performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) for getting a precise performance measuring timestamp.

Here is my updated version (game example with updated extension): [gdevelop-game(2).zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12096291/gdevelop-game.2.zip)

extensions/reviewed/FPS.json
```diff
@@ -19,82 +28,263 @@
...
+      "expressionType": {
+        "type": "expression"
+      },
       "parameters": [],
       "objectGroups": []
     },
     {
       "description": "Return your game's current FPS count, with more control over rounding.",
-      "fullName": "Precise FPS (Frames per second)",
+      "fullName": "Precise FPS (Frames per second) [Deprecated]",
```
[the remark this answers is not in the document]
tristanbob: The previous expression wasn't really working as intended.  I kept this function for backward compatibility, but I would be happy to remove it.

extensions/reviewed/FPS.json
tristanbob: I guess we could do a similar "AverageRenderTime" that works similar to FPS (taking an average of the measuring period).  Do you think that would be useful?

D8H: No, you can't measure the render time with events because it measure the waiting time too.
GDevelop or browser profilers must be used.

tristanbob:
Thanks @arthuro555 and @D8H for your help reviewing this!

PR #941, opened 2023-07-18
[Reviewed] Add AdvancedHTTP extension

arthuro555:
An advanced HTTP client with supports for headers, cache control, built-in opt-in CORS proxy (hosted on the edge via Cloudflare Workers).

Everything has been thoroughly tested via a mock server.

All events can be seen in this example screenshot:

HZ0pDSgQww

### DIFF
--- a/extensions/reviewed/AdvancedHTTP.json
+++ b/extensions/reviewed/AdvancedHTTP.json
@@ -722,7 +722,7 @@
       "objectGroups": []
     },
     {
-      "description": "CORS prevents most external websites from being queried with the browser's HTTP client, since the browser may be authenticated on that website and as such another website would be able to impersonate the player on that other website.\n\nWhen the CORS Bypass is enabled, the request will be made from a server that is not authenticated anywhere and as such is not blocked by CORS, and it eill share the response with your game. Note that as such, authentication cookies are ignored! If you own the REST API you are requesting, add CORS headers to your server instead of using this CORS Bypass.",
+      "description": "CORS prevents most external websites from being queried with the browser's HTTP client, since the browser may be authenticated on that website and as such another website would be able to impersonate the player on that other website.\n\nWhen the CORS Bypass is enabled, the request will be made from a server that is not authenticated anywhere and as such is not blocked by CORS, and it will share the response with your game. Note that as such, authentication cookies are ignored! If you own the REST API you are requesting, add CORS headers to your server instead of using this CORS Bypass.",
       "fullName": "Enable CORS Bypass",
       "functionType": "Action",
       "group": "Request settings",
--- a/extensions/reviewed/AdvancedHTTP.json
+++ b/extensions/reviewed/AdvancedHTTP.json
@@ -1,0 +1,1333 @@
+{
+  "author": "",
+  "category": "Network",
+  "extensionNamespace": "",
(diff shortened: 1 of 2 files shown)


AlexandreSi:
I have not had a look to the content yet but I saw this in your screenshot:
image

Even for developers, I think that setting the query string parameters directly in a URL is not a good practice.
Seeing how you built the action, I guess it could be easy to have an action such as:

```
Set the parameter to template Mew key: a, value 1 (set permanently in the template yes/no)
```
And you could use URLSearchParams behind the scenes. Wdyt?

arthuro555:
> Even for developers, I think that setting the query string parameters directly in a URL is not a good practice.

I was thinking that people would be able to use my already existing URL Tools extension to mutate the URL query string parameters and that it would be a bit out of scope to reimplement all the URL mutating work in this extension.

tristanbob:
@arthuro555 can you provide an example project that I can use for testing?

4ian:
Do you want to go for community while a review is on going?

arthuro555:
I'm good, I'll wait for the review 👍 I'd rather keep the flexibility of being able to have breaking changes if the review requires some.

arthuro555:
@tristanbob Here's the example I used in the screenshot: [Fetch.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12211696/Fetch.zip)

I believe I used all conditions, expressions and actions in there.

extensions/reviewed/AdvancedHTTP.json
```diff
@@ -0,0 +1,1329 @@
...
+          "description": "Request template name",
+          "name": "Request",
+          "supplementaryInformation": "sceneRequest",
+          "type": "identifier"
+        }
+      ],
+      "objectGroups": []
+    },
+    {
+      "description": "CORS prevents most external websites from being queried with the browser's HTTP client, since the browser may be authenticated on that website and as such another website would be able to impersonate the player on that other website.\n\nWhen the CORS Bypass is enabled, the request will be made from a server that is not authenticated anywhere and as such is not blocked by CORS, and it eill share the response with your game. Note that as such, authentication cookies are ignored! If you own the REST API you are requesting, add CORS headers to your server instead of using this CORS Bypass.",
```
tristanbob, line 725:
```suggestion
      "description": "CORS prevents most external websites from being queried with the browser's HTTP client, since the browser may be authenticated on that website and as such another website would be able to impersonate the player on that other website.\n\nWhen the CORS Bypass is enabled, the request will be made from a server that is not authenticated anywhere and as such is not blocked by CORS, and it will share the response with your game. Note that as such, authentication cookies are ignored! If you own the REST API you are requesting, add CORS headers to your server instead of using this CORS Bypass.",
```

tristanbob (approved):
I have reviewed this extension and chatted with @arthuro555 about it.  
The concerns I had were:

1) If CORS bypass is enabled, requests go through a middle server (by design) and could be a privacy issue.  I think this concern is addressed by the text in the "Enable CORS bypass" action, and that CORS bypass is opt-in (not enabled by default).

2) If CORS bypass is enabled and the service goes down, it could cause an outage for games.  This scenario is unlikely since the service is provided by Cloudflare workers.

Thanks @arthuro555 for once again providing high-quality power tools for advanced GDevleop users!

PR #955, opened 2023-08-05
[Flash object] Add ability to flash color tint, object effect, and opacity (fade)

tristanbob:
## New features
- Flash color tint
- Flash effect
- Flash opacity (fade)

## Improvements
- When "Flash object" action is used during an pre-existing flash, simply extend the duration of the flashing 
  - Previously, it always started with "hide object", potentially causing the object to be hidden too much

## Playable game

https://gd.games/victrisgames/flash-object-extension-example

## Project files

https://github.com/GDevelopApp/GDevelop-examples/pull/554

## Video

https://github.com/GDevelopApp/GDevelop-extensions/assets/8879811/401d5d73-c13f-4e94-96be-05afb6ff7b55


tristanbob:
@D8H my only remaining question on this extension is this:
Should we allow users to make flashing repeat forever? 

If so, should we:
- Add a new boolean property "Infinite duration" (or similar)
- Use "Flash duration = 0" to signal this state

tristanbob:
> @D8H my only remaining question on this extension is this: Should we allow users to make flashing repeat forever?
> 
> If so, should we:
> 
> * Add a new boolean property "Infinite duration" (or similar)
> * Use "Flash duration = 0" to signal this state

I decide to use the second method. Duration >= 0 will keep flashing until stopped.

D8H:
Great job, it should be useful.

You copied a deprecated function that was replaced by "onDestroy". But, I bet the extension would work without it. As it only uses object timers and properties which are already destroyed with the object by the engine.
I guess it can be important to remove it to avoid to confuse users who would take this extension as en example.

D8H:
> > @D8H my only remaining question on this extension is this: Should we allow users to make flashing repeat forever?
> > If so, should we:
> > 
> > * Add a new boolean property "Infinite duration" (or similar)
> > * Use "Flash duration = 0" to signal this state
> 
> I decide to use the second method. Duration >= 0 will keep flashing until stopped.

I like to make 2 actions in this case:
- "Start flashing"
- ""Flash"

I think it's more user-friendly than a field value that must not be used (or set to 0).

PR #956, opened 2023-08-05
[Rectangle Movement] Update: Add option to select the position on the border (inside, center, outside)

tristanbob:
## Playable game

https://gd.games/victrisgames/rectangle-movement-positions

## Project files

[Rectangle Movement Positions.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12268653/Rectangle.Movement.Positions.zip)

## Video

https://github.com/GDevelopApp/GDevelop-extensions/assets/8879811/e419143f-b4b0-4b81-8166-bbb80f091b4a

### DIFF
--- a/extensions/reviewed/RectangleMovement.json
+++ b/extensions/reviewed/RectangleMovement.json
@@ -697,7 +697,7 @@
           "type": "objectList"
         },
         {
-          "description": "",
+          "description": "Position on border",
           "name": "PositionOnBorder",
           "supplementaryInformation": "[\"Inside\",\"Center\",\"Outside\"]",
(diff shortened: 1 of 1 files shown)


D8H:
Great job, it should be useful.

You copied a deprecated function that was replaced by "onDestroy". But, I bet the extension would work without it. As it only uses object timers and properties which are already destroyed with the object by the engine.
I guess it can be important to remove it to avoid to confuse users who would take this extension as en example.

tristanbob:
I'm glad you mentioned that.  I had not seen it before so I assumed it was a new lifecycle function.  I'll remove it.

PR #974, opened 2023-08-14
New extension SelectionTools.

ayushbhardwas:
The extension provides functionality to draw different types of selections in order to select object instances present on the scene. 

#### How to use:

- Create a shape painter object. It would be used to draw the selection on the screen.
- Add the `Setup selection tools` action to an event.
- Add the setup action for the required selection. Choose from: `Set rectangular selection`, `Set polygonal selection`, and `Set lasso selection`.

---

#### Example

- Project: [SelectionTools Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12333279/SelectionTools.Example.zip)
- Build: https://gd.games/instant-builds/008aa4fb-a228-4eeb-9543-e4004d96d04b

### DIFF
--- a/extensions/community/SelectionTools.json
+++ b/extensions/community/SelectionTools.json
@@ -12,18 +12,17 @@
   "description": [
     "How to use:",
     "",
-    "- Create a shape painter object. It would be used to draw the selection on the screen.",
-    "- Add the `Setup selection tools` action to an event.",
-    "- Add the setup action for the required selection. Choose from: `Set rectangular selection`, `Set polygonal selection`, and `Set lasso selection`.",
+    "- Activate SelectionTools using the `De/Activate selection tools` action.",
+    "- Create a shape painter object and add the `Selection painter` behavior to it. Create a single instance of this object in the scene.",
+    "- Set the type of selection using one of these actions: `Set rectangular selection`, `Set polygonal selection`, and `Set lasso selection`.",
     "",
     "Tips:",
     "",
     "- The `Set rectangular selection`, `Set polygonal selection`, and `Set lasso selection` actions are used for both setting properties and changing the current selection type.",
-    "- Actions used to set things (i.e. \"Setup selection tools\" and the set action for each of the different selections) aren't required to be called each frame. They should ideally be added and triggered only once at the beginning of the scene. But they can be used again to modify specific properties or change the current selection type.",
+    "- None of the action require you to call them each frame, it's adviced that you use them with `At the beginning of the scene` condition or a `Triggred Once`.",
     "",
     "Actions:",
     "",
-    "- `Setup selection tools`",
     "- `De/Activate selection tools`",
     "- `Set rectangular selection`",
     "- `Set polygonal selection`",
@@ -34,7 +33,12 @@
     "",
     "- `Select`",
     "- `Selection activated`",
-    "- `Current selection type`"
+    "- `Current selection type`",
+    "",
+    "Behavior:",
+    "",
+    "- `Selection painter`",
+    ""
   ],
   "tags": [
     "select",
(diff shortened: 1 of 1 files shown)


ayushbhardwas:
It seems I am not allowed to access `gdjs.evtTools.input` and `runtimeScene.getObjects()`. I can maybe do a workaround for input. But I am not quite sure what to do about the getObjects() function, it is being used to get access to the shape painter object to draw the selection. The only other way I can think of is if I make this into a behavior so that I have access to the shape painter object at all times. This is an option except I was trying to decouple the selection feature from the shape painter object, to save the user from the hassle of having to pass the object in all the actions/conditions. But maybe I am wrong in thinking that and maybe the right move is to make it a behavior. I'd appreciate your feedback and/or advice.

4ian:
Hi @ayushbhardwas - this looks nice! :) 

1. For `gdjs.evtTools.input`, it's indeed best to avoid it. Can you use InputManager instead (https://docs.gdevelop.io/GDJS%20Runtime%20Documentation/classes/gdjs.RuntimeGame.html#getInputManager). Get it through `getGame()` ont the `runtimeScene`
2. For `runtimeScene.getObjects`, as you mentioned it's usually I think a good idea to make a behavior because it means you will have access to the shape painter. And for the objects to select/unselect, then usually they are passed as parameter of the actions/conditions.

I think it would be nice to try to fix both of these. I understand it's a fair amount of rework to convert this to a behavior, but please give a try and give a try to avoiding getObjects too :) 

If you struggle too much, we'll add exceptions so we can merge your extension... but I think it would be future proof to fix both of these 👍

Thank you again for this nice work! 👍

ayushbhardwas:
Thanks a lot, @4ian
I'm incredibly grateful that you liked it and I will surely work on fixing the issues. 😊

ayushbhardwas:
####Changes made:

- Added `Selection painter` behavior.
- Migrated events related to shape painter to the new behavior eliminating the need to use `runtimeScene.getObjects`.
- Removed the `Setup selection tools` action. Since the painter is now identified via the behavior thus eliminating the need for this action.
- Rewrote the shape drawing logic using events (earlier js) since now, the shape painter object is not being accessed through `runtimeScene`. 
- The previous change also eliminates the need for `gdjs.evtTools.input`.

---

####How to use:

- Activate SelectionTools using the `De/Activate selection tools` action.
- Create a shape painter object and add the `Selection painter` behavior to it. Create a single instance of this object in the scene.
- Set the type of selection using one of these actions: `Set rectangular selection`, `Set polygonal selection`, and `Set lasso selection`.

---

####Example

- Project: [SelectionTools Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12338101/SelectionTools.Example.zip)
- Build: https://gd.games/instant-builds/f306d6eb-a2f6-43ed-a21f-bbe65f893d7e

tristanbob:
I haven't looked at the extension, but in your example game, units are not highlighted until the release happens.  
Is there a "preselected" state that can be used to determine which objects would be selected when the mouse releases? 
(RTS Unit Selection provides this)

ayushbhardwas:
@tristanbob No, as you speculated, there isn't a "preselected" state. I'm not sure why you'd want that, maybe you can help me understand that better. 

The extension does not use object variables as flags, it just filters the current context. Furthermore, it's slightly expensive to run the checks for polygons (lasso and polygonal selection) for each frame, depending on how big the polygon is and how many total instances of the object(s) we want to select are present in the scene. 

I remember our conversation about merging this extension with RTS Unit Selection, which when I looked into it is a very different sort of tool. I think it adds a lot that may be useful for a certain type of game I just don't see myself using it that often. This is supposed to be a more barebone extension. But I haven't forgotten about it, I think RTS Unit Selection would benefit immensely by having a lasso select, which if you allow I would definitely want to implement. On a similar note, you'd agree that you never really use polygonal selection to select units in an RTS so I don't think it needs that, not that it can not be implemented, just that it's not commonly seen in RTS games.

ayushbhardwas:
#### Changes made:

- Added the ability to `Pre select` i.e. select instances before closing the shape. Activate or deactivate it using the `Pre select` boolean parameter in the `Select` condition.

---

#### Example

- Project: [SelectionTools Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12342009/SelectionTools.Example.zip)
- Build: https://gd.games/instant-builds/75a87656-a517-4b3c-8077-6ba6e664c908

ayushbhardwas:
#### Changes made:

- Added the ability to choose if the polygonal selection is represented by a `True polygon`. If the selection is a `True polygon`, dragging while the mouse button or touch is down will not place new vertices. This is especially useful for devices with a touch screen where there is no distinction between hover and pressed state, making it difficult to preselect without dragging the touch around which will also place new vertices. Something that you might not want. With `True polygon` the vertices get placed when the touch or mouse button is released making it possible to drag across a touch screen and preselect object instances. 
- Bug fixes
- Changes to the description and condition names
- Refactored pre-selection events to eliminate redundant checks.

---

#### Example

- Project: [SelectionTools Example.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12354796/SelectionTools.Example.zip)
- Build: https://gd.games/instant-builds/75a87656-a517-4b3c-8077-6ba6e664c908

ayushbhardwas:
I intend to make no further changes to the extension before the merger unless someone reports an issue on this PR. Consequently, I eagerly await the extension team's review and approval.

4ian:
@ayushbhardwas Many thanks again for the work on this and sorry for the delay here. I think it's fine to go ahead and quite flexible already! If you have a way to ensure this work on touch screen that would be perfect. In the meantime, because it's already well made and you've spent time on this, I'm merging this.

Many thanks and keep up the great work 👍

PR #980, opened 2023-08-17
Update PanelSpriteButton.json

gemmomoh:
Add disabled state.

A disabled state is visible when the interactions are deactivated.

image

### DIFF
--- a/extensions/reviewed/PanelSpriteButton.json
+++ b/extensions/reviewed/PanelSpriteButton.json
@@ -686,6 +686,44 @@
         {
           "fullName": "",
           "functionType": "Action",
+          "name": "onActivate",
+          "sentence": "",
+          "events": [
+            {
+              "type": "BuiltinCommonInstructions::Standard",
+              "conditions": [],
+              "actions": [
+                {
+                  "type": {
+                    "value": "PanelSpriteButton::ButtonFSM::ResetState2"
+                  },
+                  "parameters": [
+                    "Object",
+                    "Behavior",
+                    ""
+                  ]
+                }
+              ]
+            }
+          ],
+          "parameters": [
+            {
+              "description": "Object",
+              "name": "Object",
+              "type": "object"
+            },
+            {
+              "description": "Behavior",
+              "name": "Behavior",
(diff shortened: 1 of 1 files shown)


D8H:

Did you consider applying an effect on the button after disabling it?

gemmomoh:
No, I did not add an effect. I think people usually have a disabled button sprite/image.

Bouh:
As the other states are already in images, it's a good to have this new state as an image too, to be consistent.
The effects are also rendered at each frame, which is a minor hit on the GPU, an additional cost that can be avoided.

Bouh (approved):

D8H:
> As the other states are already in images, it's a good to have this new state as an image too, to be consistent. The effects are also rendered at each frame, which is a minor hit on the GPU, an additional cost that can be avoided.

Adding a new state means that every assets from the store must be updated with an image for this state. Also, I think it's not very practical to use an image when, for most buttons, it will just be a grey-ish version of the idle state.

Effects performance is another matter. There is no need to evaluate a shader each frame when the image stays the same.

D8H (changes requested):
Please, extension team members, do not merge this PR.

4ian:
I think we should start thinking about having different variations of buttons :)

Bouh:
> I think we should start thinking about having different variations of buttons :)

What did you suggest, kept the object in this version, and duplicate the a new object with improvment on a disabled state?
It will duplicate the object buttons just for that difference...

>Also, I think it's not very practical to use an image when, for most buttons, it will just be a grey-ish version of the idle state.

I agree.

>Effects performance is another matter. There is no need to evaluate a shader each frame on a still image.

I think the shaders from effects are calculated at every frame, I'm not sure. But if we can calculate the effect one time on the object its a wonderful solution.
Except the effect doesn't work on children objects in a EBO like I saw and shared on discord, so we can't use them for the moment.

romw314:
This PR should have the 'Extension updated' tag

arthuro555:
As per the thread above, the reviewers rejected these changes. Closing PR now.

PR #985, opened 2023-08-22
[Reviewed] [Platformer character animator] Handle 3D models

D8H:
### Demo
- https://github.com/GDevelopApp/GDevelop-examples/pull/562


4ian:
It's not finished nor released yet, but just want to say that this will be incredible :)

PR #994, opened 2023-08-28
New Extension: Key Press 

TheGemDev:
An extension that checks how many times a keyboard key was pressed in a period of time ( maximum delay ), works just like the double click extension but for keyboard keys.

image

### DIFF
--- a/extensions/community/KeyPress.json
+++ b/extensions/community/KeyPress.json
@@ -1,0 +1,360 @@
+{
+  "author": "",
+  "category": "Input",
+  "extensionNamespace": "",
+  "fullName": "Key Press",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


tristanbob:
Does this also have a function to detect how long a key has been depressed?  (for charging spells/attacks)

Condition: "Button X has been pressed for Y seconds"

I also wonder if it should have an expression that shows the percent complete?

Action: Start an X second long charge named "SuperAttack"

Expression: ChargePercent("SuperAttack")

Condition: "SuperAttack" is charging
Condition: "SuperAttack" is fully charged

Action: Cancel charging "SuperAttack" (the user can use this for key released, or perhaps if the player gets hit while charging)

4ian:
Just commenting to say that this is a great idea and should simplify the work of many people if this can be flexible enough 👍

TheGemDev:
> Does this also have a function to detect how long a key has been depressed? (for charging spells/attacks)
> 
> Condition: "Button X has been pressed for Y seconds"
> 
> I also wonder if it should have an expression that shows the percent complete?
> 
> Action: Start an X second long charge named "SuperAttack"
> 
> Expression: ChargePercent("SuperAttack")
> 
> Condition: "SuperAttack" is charging Condition: "SuperAttack" is fully charged
> 
> Action: Cancel charging "SuperAttack" (the user can use this for key released, or perhaps if the player gets hit while charging)

Sure, i can try adding a feature like that.

TheGemDev:
this works

image

TheGemDev:
I was wondering what data type is required here?. the parameter type is a "keyboard key", but typing the key as a string doesn't seem to work.

image

D8H:
> I was wondering what data type is required here?. the parameter type is a "keyboard key", but typing the key as a string doesn't seem to work.

The screenshot doesn't show the action sentence. I don't have enough context to answer.

TheGemDev:
Okay, I need to take a look at the extension again, I will send an update later in the day.

tristanbob:
> Okay, I need to take a look at the extension again, I will send an update later in the day.

Do you want to work on this more?  It is probably close to being merged when you are ready.

gemmomoh:
yes, i will still work on it.

gemmomoh:
@D8H This was my question, in order to get the key pressed duration I need to get the key user is checking for I used this 
image
image

the parameter type is a keyboard key 
image

but it still returns an error.

D8H:
> @D8H This was my question, in order to get the key pressed duration I need to get the key user is checking for I used this 
> 
> the parameter type is a keyboard key
> 
> but it still returns an error.

It's really hard to follow with the events being cut.
This parameter type don't accept an expression. If you need call private actions use a text parameter.

gemmomoh:
okay

TheGemDev:
image

D8H:
oh, ok, I guess the key parameter type can't be used for expression parameter. The UI should probably not show it in the list in this case.

PR #1002, opened 2023-09-03
Update face forward extension

TheGemDev:
The face forward extension is excellent but sometimes fails to get the default direction, adding a default direction property fixes the issue. 

image

### DIFF
--- a/extensions/reviewed/FaceForward.json
+++ b/extensions/reviewed/FaceForward.json
@@ -142,7 +142,7 @@
                       },
                       "parameters": [
                         "Object",
-                        "Object.Behavior::PropertyMovementDirection() - Object.Behavior::PropertyDirection()",
+                        "Object.Behavior::PropertyMovementDirection() - Object.Behavior::OffsetAngle()",
                         "abs(Object.Behavior::PropertyRotationSpeed())",
                         ""
                       ]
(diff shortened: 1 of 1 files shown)


TheGemDev:
I should have named it named it default angle.

Bouh:
> I should have named it named it default angle.

- Is "Offset angle" is more appropriate?

- Please remove the dot at the end of the short description.

extensions/reviewed/FaceForward.json
```diff
@@ -421,15 +384,27 @@
...
           "description": "",
+          "group": "",
           "extraInformation": [],
           "hidden": false,
           "name": "RotationSpeed"
         },
+        {
+          "value": "0",
+          "type": "Number",
+          "label": "Default direction.",
```
Bouh, line 395: Remove the dot here.

extensions/reviewed/FaceForward.json
```diff
@@ -421,15 +384,27 @@
           "type": "Number",
           "label": "Rotation speed (degrees per second).  Use \"0\" for immediate turning.",
```
Bouh, line 385:
```suggestion
          "label": "Rotation speed (degrees per second).  Use \"0\" for immediate turning",
```

extensions/reviewed/FaceForward.json
Bouh, line 395:
```suggestion
          "label": "Default direction",
```

TheGemDev:
fixed

tristanbob:
@TheGemDev great idea to add this feature!  I'm merging this now so people can start benefiting from it. :)

PR #1015, opened 2023-09-17
[Reviewed] [Marching squares] Fix the mode that only draw the visible part for portrait screens

tristanbob:


extensions/reviewed/MarchingSquares.json
```diff
@@ -1981,7 +1979,7 @@
           "objectGroups": []
         },
         {
-          "description": "Cap every value of the field to a range.",
+          "description": "Restrict every value of the field to a range.",
```
D8H, line 1982: I chose "cap" because the values are cut. "Restrict" could mean anything, for instance something like a scaling.
Is "cap" too slang?

extensions/reviewed/MarchingSquares.json
```diff
@@ -2613,9 +2603,9 @@
                 "",
                 "    const minX = Math.max(0, Math.floor(behavior.coordConverter.convertToGridBasisX(screenLeft)));",
                 "    const minY = Math.max(0, Math.floor(behavior.coordConverter.convertToGridBasisY(screenTop)));",
-                "    // I don't know why the + 1 is needed",
+                "    // Add 1 to ensure area can cover the entire screen",
```
D8H, line 2606: I think a comment is no longer needed here.
```suggestion
                "",
```

D8H (commented):
Thanks a lot for the fixes and the typos!

extensions/reviewed/MarchingSquares.json
tristanbob: EIther word works, I will keep your version.

D8H:
Note that the tags order is `[Reviewed]` then the name. Otherwise, it's hard to read the changelog.

D8H:
Great job, merging it now!

PR #1016, opened 2023-09-17
[Reviewed] [Object Picking Tools] Add actions and conditions to pick objects with the highest or lowest value of an object variable

tristanbob:
## Project files

~~[PickHighestLowestValue.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12643851/PickHighestLowestValue.zip)~~

~~[PickHighestLowestValue.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12644627/PickHighestLowestValue.zip)~~

[PickHighestLowestValue.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/12698815/PickHighestLowestValue.zip)

## Playable game

https://gd.games/victrisgames/object-picking-tools

## Video

https://github.com/GDevelopApp/GDevelop-extensions/assets/8879811/f07cdc37-8e99-4208-986c-4e13f786dc8f


tristanbob:
I also want to propose promoting this extension to Reviewed status.  It's very useful and our users would benefit from having it in the Reviewed status.

tristanbob:
@arthuro555 I have updated the new functions so that all instances that match the highest (or lowest) value are picked.

I strongly feel this is the best behavior because it matches what other conditions in GDevelop do.  This logic enables users to combine multiple conditions to filter picked items.

For example, I am making a tower defense game and I want to shoot the objects that are farthest along the path.  I can use two conditions for this:

```
Conditions:
- Pick "Enemy" objects with highest variable "TargetWaypointID" (this will often be multiple objects)
- Pick "Enemy" objects with lowest variable "DistanceToNextWaypoint"

Actions:
- Fire bullet at "Enemy" objects
```

If you agree, I will rework the other functions in this extension and change the version to 2.0.0.

tristanbob:
I have promoted this extension from the Community to the Reviewed folder.

I also updated Zorder picking to return all instances that match the highest/lowest value.
Since this is a different behavior than before, I updated this to version 2.0.0 and added a note in the extension description.

PR #1017, opened 2023-09-19
Upgrade in app purchase

ClementPasteau:
tested with example: https://github.com/GDevelopApp/GDevelop-examples/pull/574

Mainly followed https://github.com/j3k0/cordova-plugin-purchase/wiki/HOWTO:-Migrate-to-v13 and API of functions for upgrade.

Screenshot_20230919-142408
Screenshot_20230919-145614

### DIFF
--- a/extensions/reviewed/InAppPurchase.json
+++ b/extensions/reviewed/InAppPurchase.json
@@ -282,7 +282,7 @@
       "objectGroups": []
     },
     {
-      "description": "Returns true if the item with ID is owned, false otherwise. (can return false until the store is initialized)",
+      "description": "Returns true if the item with ID is owned, false otherwise. (can return false until the store is initialized).",
       "fullName": "Check if item is owned",
       "functionType": "Condition",
       "name": "IsItemOwned",
--- a/extensions/reviewed/InAppPurchase.json
+++ b/extensions/reviewed/InAppPurchase.json
@@ -1,7 +1,6 @@
 {
   "author": "",
   "category": "Ads",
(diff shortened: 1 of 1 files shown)


4ian (approved):
Overall lgtm!

extensions/reviewed/InAppPurchase.json
```diff
@@ -140,90 +153,106 @@
...
+            "// Automatically finish a receipt when verified.",
+            "store.when().verified(receipt => receipt.finish())",
+            "",
+            "store.when().finished((transaction) => {",
+            "    // If we're watching this event, call the callback.",
+            "    if (eventName === 'finished') {",
+            "        transactionCallback(transaction);",
+            "    }",
+            "    // Nothing left to do when finished.",
+            "})",
```
arthuro555, line 233: I have received multiple reports of the extension being broken & orders going unfulfilled after this update.

Taking a look now, this looks extremely wrong to me. 
1. This adds a global event handlers for every event type every time this action is called? This is completely unnecessary, harmful to memory and performance, and will almost certainly lead to multiple calls of `.verify()`/`.finish()`, which is not valid and likely will cause unintended behavior!

2. This forces a forbidden behavior upon users: once this action has been called, and the purchase of the product is started, it will automatically do the steps to finalize the purchase. However, it is absolutely not how it is supposed to be done! Not letting the creator handle verification themselves is arguably not that big of a deal, but we cannot mark the purchases as finalized automatically. A purchase may only ever be marked as finished once the game code has successfully delivered the product that was bought, at this point we know for sure the item has not been delivered yet. Heck, this code will mark as finished *ANY* product that was bought, even ones for which the creator has not even put up this event listener for!
3. It is a very error prone implementation. If this action does not get called, a purchase of a product will never be processed since those event handlers will never have been set. The action is inconspicuous, doesn't seem and doesn't say in any way that it is crucial for the extension to function!

ClementPasteau: Thanks for the feedback @arthuro555 
I agree the implementation looks incomplete and was done in haste, to prevent creators from being stuck in publishing their app on the stores because of the Google Play deadline.
The tests I had done were fine on the example, but it may have been not enough.
I feel bad about having published this update now.

As you know well how this works, would you have time to improve it and make the extension less error-prone?

PR #1020, opened 2023-09-24
Array tools v2

arthuro555:
A long overdue overhaul of the array tools extension!

- Use Scene Variable parameters instead of string parameters (allows using nested arrays)
- Add Global and Object variable variants
- Change the icon to a "list" icon, rather than a `[ ]` icon, for clarity
- Rename some expressions to fit best practices
- Clean up the code a lot, replaced abbreviations with good and consistent variable names
- Add categories to be able to navigate the different functions more easily

This goes without saying, but this is a breaking change!

### DIFF
--- a/extensions/reviewed/ArrayTools.json
+++ b/extensions/reviewed/ArrayTools.json
@@ -1,15 +1,32 @@
 {
   "author": "Arthur Pacaud (arthuro555)",
   "category": "General",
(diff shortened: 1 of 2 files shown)


4ian (approved):
I think this is in any case a major improvement over what we have now. Merging this now!
If there are improvements or comments by other people, this can be addressed in another PR :)

PR #1055, opened 2023-10-16
[Community] [3D Tween] Fix typos descriptions

AlexandreSi:


AlexandreSi:
Also, I think that the ScaleVolume action should be fixed (sentence, description and fullName are the ones of the StretchZ action) but I was not sure what to write instead:

https://github.com/GDevelopApp/GDevelop-extensions/blob/main/extensions/community/Tween3D.json#L313-L318

D8H:
> Also, I think that the ScaleVolume action should be fixed (sentence, description and fullName are the ones of the StretchZ action) but I was not sure what to write instead

Actually, you saw them because you read the json directly, but ActionWithOperator functions are using the associated expression declaration. I must have started with an action, changed the type and copy-pasted it.

D8H (approved):

PR #1075, opened 2023-11-05
Update DoubleClick.json

Neko-MB:
Updated to work for GDevelop version 5.3 and up (hopefully) Also added "Back/Forward" buttons incase one use so with mouse.

### DIFF
--- a/extensions/reviewed/DoubleClick.json
+++ b/extensions/reviewed/DoubleClick.json
@@ -1,7 +1,6 @@
 {
   "author": "Silver-Streak",
   "category": "Input",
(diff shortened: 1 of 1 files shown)


D8H:

> Updated to work for GDevelop version 5.3

What do you mean? I tried the existing extension with GDevelop 5.3.180 and didn't see any issue compared to older released.

> Also added "Back/Forward" buttons in case one use so with mouse.

Indeed, this extension was not updated when these 2 choices were added.

There is also an issue when the condition is used several times in the events. Only the first one is working.

image

Do you want to help solve this issue? One solution could be to clear the state only at the end of events in this life-cycle function:

image

PR #1076, opened 2023-11-06
[Community] [WithThreeJs] Supports GDevelop v5.3.180

PANDAKO-GitHub:
Supports GDevelop v5.3.180
Added custom object to always enable built-in 3D rendering.


4ian (approved):
Thank you!

PR #1115, opened 2023-12-03
[Reviewed] [Object Spawner] Added "Random Position" checkbox and deprecated "Offset" properties

tristanbob:
Offsets were originally designed to enable random positions, but this new implementation is much cleaner. The offset method can still be used by adding a value to the position of objects as a subevent to the "Periodically spawn objects".  This was not possible before GDevelop started picking objects created by extensions.

Playable game:
https://gd.games/instant-builds/cbf53f3e-449a-45e2-b7c1-011f257c7343

Project file:
[ObjectSpawnerRandomPositions.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/13540079/ObjectSpawnerRandomPositions.zip)

### DIFF
--- a/extensions/reviewed/ObjectSpawner.json
+++ b/extensions/reviewed/ObjectSpawner.json
@@ -2,21 +2,25 @@
   "author": "",
   "category": "Game mechanic",
   "extensionNamespace": "",
-  "fullName": "Object spawner",
+  "fullName": "Object spawner area",
   "helpPath": "",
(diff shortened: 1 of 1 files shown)


extensions/reviewed/ObjectSpawner.json
```diff
@@ -1241,6 +1373,136 @@
...
+              "description": "Behavior",
+              "name": "Behavior",
+              "supplementaryInformation": "ObjectSpawner::ObjectSpawner",
+              "type": "behavior"
+            }
+          ],
+          "objectGroups": []
+        },
+        {
+          "description": "Change if use random positions inside the spawner. Useful for making large spawner areas.",
```
D8H, line 1426: Are areas rectangular? I guess distance based area could be useful too. 
Users will have to deal with big spawner objects. It's not always easy to manage in the scene editor.

extensions/reviewed/ObjectSpawner.json
```diff
@@ -1241,6 +1373,136 @@
...
+          "fullName": "Use random positions inside the spawner",
```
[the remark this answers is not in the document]
D8H: There is also a condition I guess.

tristanbob:
I updated the wording, and I like it a lot better.  What do you think?

extensions/reviewed/ObjectSpawner.json
tristanbob: Another important aspect of providing a visual spawning area is that creators can make sure their objects are not accidentally spawned inside a wall or outside the playable area.

D8H:
It sounds good, thank for the update!

PR #1116, opened 2023-12-03
[Reviewed] [Flash object] Added get/set functions for half-time property

tristanbob:
- Changed an input parameter to type "effect"
- Minor wording changes

## Game example
https://github.com/GDevelopApp/GDevelop-examples/pull/603

### DIFF
--- a/extensions/reviewed/Flash.json
+++ b/extensions/reviewed/Flash.json
@@ -1,6 +1,6 @@
 {
   "author": "@4ian, Entropy, VegeTato",
-  "category": "",
+  "category": "Visual effect",
   "extensionNamespace": "",
   "fullName": "Flash object",
   "helpPath": "",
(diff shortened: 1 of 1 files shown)


extensions/reviewed/Flash.json
```diff
@@ -909,6 +913,87 @@
             }
           ],
           "objectGroups": []
+        },
+        {
+          "description": "the half period (time during which object is invisible) of the object.",
+          "fullName": "Half period (time during which object is invisible)",
```
D8H, line 919:
```suggestion
          "fullName": "Half period",
```
Names should not include description the description field is enough.

D8H: The issue is probably that the property name is also too long and the property description should have been used.
The generator automatically adds the property description at the end of the expression description.

extensions/reviewed/Flash.json
```diff
@@ -2635,6 +2886,89 @@
             }
           ],
           "objectGroups": []
+        },
+        {
+          "description": "the half period (time between flashes), in seconds of the object.",
+          "fullName": "Half period (time between flashes), in seconds",
```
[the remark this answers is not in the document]
tristanbob: Check out my latest commit.

PR #1145, opened 2024-01-21
[Community] 3D jump

D8H:
### Following
- https://github.com/GDevelopApp/GDevelop-extensions/pull/896
- https://github.com/GDevelopApp/GDevelop-extensions/pull/1048

### Demo
- https://github.com/GDevelopApp/GDevelop-examples/pull/621

### DIFF
--- a/extensions/community/Jump3D.json
+++ b/extensions/community/Jump3D.json
@@ -1,0 +1,2230 @@
+{
+  "author": "",
+  "category": "Movement",
+  "extensionNamespace": "",
+  "fullName": "3D jump",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


gabrielzv1233:
we may want to make the separate camera from platform function a behavior by making it check for collision of an object groups assigned via a behavior for the obstacle

gabrielzv1233:
here is a video demonstrating the problems
https://streamable.com/37ly6k

D8H:
>  if I put an obsticle above the player, that it doesn't collide with it

Yes, I wasn't sure if this was useful or not, but it should be easy to fix.

> collisions with slanted objects (image attached) are wonky

It's not ideal but expected.

image

gabrielzv1233:
gravity value and jump height are broken, solution, add platformer trajectory extension (seems like you used a function from platformer trajectory without the extension being installed, either add it as a dependency and into the example or take the function from the platformer trajectory extension and put it into the jump 3D extension to prevent the need of installing more than needed
suggestion, make max falling speed the same value as gravity and replace it will jump speed (how fast the player goes up when jumping, seems like the property is there but there is no functionality)
rename land on floor function to cancel jump, and make desc  something like "Cancel the jump of object causing it to fall" 
when on floor, the is on floor value fluctuates from true and false
[Screen recording 2024-01-21 11.18.21 AM.webm](https://github.com/GDevelopApp/GDevelop-extensions/assets/77300296/18265dd3-6caa-4086-ad62-b8422a19b6d0)
why is this a problem? if someone wants to say create a object when on the floor than they would use this event and a trigger once so it wouldnt keep doing it every tick/frame, the problem is, even if trigger once is on, it is constantly fluctuating, so it would just keep spamming objects if that makes sense

D8H:
Thanks a lot for your feedback. I fixed the issues and added an action to abort the jump.

gabrielzv1233:
> Thanks a lot for your feedback. I fixed the issues and added an action to abort the jump.

awesome, can you send another copy of the example so I can test the updated version?

gabrielzv1233:
nevermind, figured out that the link updated automatically when you put a new commit in, I do see that the problem with the fluctuating is on floor value is still there and wasn't mentioned to be fixed in the commits, if you want to know how I see if its still fluctuating, I just use this code and a lazaly slapped on textbox to the scene
image

D8H:
Do you see anything else?

D8H:
Thank you for your feedback, I'm moving ahead with this.
Please, open an issue if you want to talk future improvements.

PR #1149, opened 2024-01-27
[Reviewed] [Mouse pointer lock] Avoid errors in Safari on iOS.

PANDAKO-GitHub:
Avoid errors when running the "Request Pointer Lock" and "Exit pointer lock" actions in Safari on iOS.


D8H (approved):
Thank you for the update, it looks good to me.

PR #1162, opened 2024-02-08
3D Movement on local axis

PANDAKO-GitHub:
This extension provides behaviors that allow you to translate and rotate 3D objects along local axes.

Add the "Movement 3D on local axis" behavior to your 3D box or 3D model. The following actions will then be available:
- Translate by distance
- Translate by speed
- Rotate by angle
- Rotate by speed

### Example
https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/PANDAKO-GitHub/GDevelop-examples/3D-Movement-on-local-axis-Example/examples/3d-movement-on-local-axis/3d-movement-on-local-axis.json

### DIFF
--- a/extensions/community/Movement3dOnLocalAxis.json
+++ b/extensions/community/Movement3dOnLocalAxis.json
@@ -1,0 +1,281 @@
+{
+  "author": "",
+  "category": "Movement",
+  "extensionNamespace": "",
+  "fullName": "3D Movement on local axis",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


D8H:

Why do you need JavaScript for this extension?
I guess the same features can be done using something like forces but for 3D. Mah started to work on an extension for 3D forces:
https://discord.com/channels/258623956158906368/936338995875577926/1204096159279226940

About the rotation, I think it would need acceleration feature to worth an extension. Otherwise, it can be done by using built-in actions and `TimeDelta()`.

PANDAKO-GitHub:
Thank you for checking.

The feature of this extension is that it uses local axes.
It allows for control not possible (or very difficult) with built-in actions and other extensions that always use global axes.

For example, it will be easier to control an aircraft that can move and rotate 360 degrees, and the bullets fired from it.
This is very difficult to do with built-in actions and extensions designed for platformer games.

The reason I used Javascript is that it is inefficient to reproduce functionality that Three.js already has in place using events.

D8H:
> The feature of this extension is that it uses local axes. It allows for control not possible (or very difficult) with built-in actions and other extensions that always use global axes.
> 
> For example, it will be easier to control an aircraft that can move and rotate 360 degrees, and the bullets fired from it. This is very difficult to do with built-in actions and extensions designed for platformer games.

I see. Do you think there is other cases where the actions can be useful? Because if it's mostly used for spaceships/planes movement, I guess it could be easier for users to have a full behavior that handles acceleration and maximum speed.

PANDAKO-GitHub:
This extension can also be used when the user wants to have different speed performance for each axis of the object.
This extension can also be used when users want to perform relative object placement.

I created this thinking I need to provide basic actions to the user.
I think it's good that a separate extension is created that can handle acceleration and top speed.

D8H:
> I created this thinking I need to provide basic actions to the user.

That sounds fair. I'll wait to review the 3D force extension, because I guess that the translate actions could be one of the way to add forces.

PR #1163, opened 2024-02-08
3D Movement on local axis of camera

PANDAKO-GitHub:
This extension provides actions that allow you to move and rotate the camera along local axes.

The following actions are added to the camera category.
- Translate camera by distance
- Translate camera by speed
- Rotate camera by angle
- Rotate camera by speed

### Example
https://editor.gdevelop.io/?project=https://raw.githubusercontent.com/PANDAKO-GitHub/GDevelop-examples/3D-Movement-on-local-axis-Example/examples/3d-movement-on-local-axis/3d-movement-on-local-axis.json

### DIFF
--- a/extensions/community/Movement3dOnLocalAxisOfCamera.json
+++ b/extensions/community/Movement3dOnLocalAxisOfCamera.json
@@ -1,0 +1,256 @@
+{
+  "author": "",
+  "category": "Camera",
+  "extensionNamespace": "",
+  "fullName": "3D Movement on local axis of camera",
+  "helpPath": "",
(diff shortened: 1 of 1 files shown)


PANDAKO-GitHub:
Oops... am i doing any mistake? 🤔

D8H:

> Oops... am i doing any mistake?

It's because the extension uses `gdjs.scene3d` without declaring it in `scripts/lib/ExtensionsValidatorExceptions.js`

Can you explain what needs this extension answer?
It seems the example could be done with this extension:
- https://wiki.gdevelop.io/gdevelop5/extensions/third-person-camera/

PANDAKO-GitHub:
Thank you for checking.

The feature of this extension is that it uses local axes.
With the "Third person camera" extension, which uses global axes, it is not possible to fix the camera behind a rotating aircraft in 3D.

> It's because the extension uses gdjs.scene3d without declaring it in scripts/lib/ExtensionsValidatorExceptions.js

This extension avoids the error if 3D is not used, but is that not enough?

D8H:
The 2D layer position and angle are set to Three.js camera when the camera actions are used or when the game resolution changes:
https://github.com/4ian/GDevelop/blob/ba8d7f4e3824afabd18fbd960481ecd4dacad1e0/GDJS/Runtime/pixi-renderers/layer-pixi-renderer.ts#L470-L472

The extension probably needs to surround the rotation/position change by updates, something like this:
```JS
const layer = runtimeScene.getLayer(layerName);
const camera = layer.getRenderer().getThreeCamera();
if (camera) {
    camera.rotation.z = gdjs.toRad(layer.getCameraRotation());
    
    camera.rotateX(gdjs.toRad(angle));

    layer.setCameraRotation(gdjs.toDegrees(camera.rotation.z));
}
```

For instance, try to add 2 "Angle of camera" action to add 1 and subtract 1. Without the fix I guess it would mess the 3D camera rotation.

PANDAKO-GitHub:
@D8H , Thank you for your review.
I fixed the issue you pointed out.

PR #1174, opened 2024-02-18
Create DrawGrid.json

TheGemDev:
draw a virtual grid.

image

image

image

### DIFF
--- a/extensions/community/DrawGrid.json
+++ b/extensions/community/DrawGrid.json
@@ -39,43 +39,6 @@
           "conditions": [
             {
               "type": {
-                "value": "DepartScene"
-              },
-              "parameters": [
-                ""
-              ]
-            }
-          ],
-          "actions": [
-            {
-              "type": {
-                "value": "ModVarScene"
-              },
-              "parameters": [
-                "__posX",
-                "=",
-                "OffsetX"
-              ]
-            },
-            {
-              "type": {
-                "value": "ModVarScene"
-              },
-              "parameters": [
-                "__posY",
-                "=",
-                "OffsetY"
-              ]
-            }
-          ]
-        },
-        {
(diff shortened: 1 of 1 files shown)


TheGemDev:
Project - [Draw Grid.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/14323360/Draw.Grid.zip)

Extension - 
[Draw Grid Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/14323371/DrawGrid.zip)

Oxey405 (changes requested):
So it's a great idea but there are a few things to fix I think before this gets published
1. Performance
Performance is really bad because of how you draw the cells (using a loop)
At least you should warn people that it's gonna take a long time to render for fine-grain loops
2. Coloring
I think it would also be great to explain how to change the color of the grid for example in the description of the extension, the action or in a help page

Other than that it looks great :)

Oxey405:
> Project - [Draw Grid.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/14323360/Draw.Grid.zip)
> 
> Extension - [Draw Grid Extension.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/14323371/DrawGrid.zip)

Thanks

D8H:
> Performance is really bad because of how you draw the cells (using a loop)

The issue is not the loop. It's the instance creation.

TheGemDev:
fixed the performance and added a way to change colors in the action. Also, if no instance of the shape painter object exists,  the action creates one.

image

D8H:
What happens when the camera is moved?
Can extension users use the grid to debug path finding of a game with scrolling?
What do you think about making a behavior that updates the grid when needed?

PR #1210, opened 2024-03-08
NEW EXTENSION - Extended Animations

TheGemDev:
This extension provides:
Expression to return the total number of animations and frames,
Action to loop animations,
Action to set the time between frames,
Action to set origin and center points.

Screenshot 2024-03-08 at 14 30 03
Screenshot 2024-03-08 at 14 30 52
Screenshot 2024-03-08 at 14 30 38

Project File - 
[Archive.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/14538747/Archive.zip)

Extension - 
[ExtendedAnimations.json](https://github.com/GDevelopApp/GDevelop-extensions/files/14538753/ExtendedAnimations.json)

### DIFF
--- a/extensions/community/ExtendedAnimations.json
+++ b/extensions/community/ExtendedAnimations.json
@@ -29,7 +29,7 @@
       "description": "Returns the total number of animations in a sprite.",
       "fullName": "Total number of animations",
       "functionType": "Expression",
-      "name": "totalAnimation",
+      "name": "TotalAnimation",
       "sentence": "",
       "events": [
         {
@@ -61,7 +61,7 @@
       "description": "Returns the total number of frames in a sprite animation.",
       "fullName": "Total number of frames in an animation",
       "functionType": "Expression",
-      "name": "totalFrames",
+      "name": "TotalFrames",
       "sentence": "",
       "events": [
         {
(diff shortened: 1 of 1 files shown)


D8H:

> Expression to return the total number of animations and frames

- I guess it's rare to check the number of animations and a variable can be used if necessary.
- There is already an expression for the number of frames: `Sprite.AnimationFrameCount()`

> Action to loop animations,
> Action to set the time between frames,
> Action to set origin and center points.

Can you explain what is the purpose of these actions?
Do you use them in one of your project?

Members that starts with an underscore are private and extensions should avoid to use them because they may not exist anymore in future releases.

TheGemDev:
Sometimes i'm in a situation where i need to get the total number of animations, and i believe having an expression return that would be so much faster.

i forgot about the frames count.
An action to stop a loop is essential.

jumpingjh:
Hello,
it seems that the command
"object._animations[index].directions[0].frames[i].origin.x = x;"
generate an error. 
Thanks,
J

TheGemDev:
> Hello, it seems that the command "object._animations[index].directions[0].frames[i].origin.x = x;" generate an error. Thanks, J

hello it only works on version 195, as D8H said "Members that starts with an underscore are private and extensions should avoid to use them because they may not exist anymore in future releases"

I have an update though. which should work with the current version.

jumpingjh:
Thanks @TheGemDev  for your replay.

Reading theGDevelop projecyt I I tried with this  command
MyObject.animations[0].directions[0].sprites[0].originPoint.x = 10;
but it downs't work. 

I will wait for your new version extension.

If you need a tester, I can help you.

Thanks,
J

PR #1258, opened 2024-04-21
[Reviewed] [Transition] Fix the animation for zoomed layers

D8H:
### Demo
- [Transition.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/15051358/Transition.zip)

### Changes
- Fix the animation for zoomed layers
- Always keep the last frame when `"Forwoard"` is chosen
- Simplify the events

### Todo
- Add a property for the smoothing radius
- Use the shape painter color and opacity instead of the parameters
- Use easing
  - Make 3 actions:
    - Fade in
    - Fade out
    - Fade in and out
      - with 2 durations and 2 easing and a pause

### DIFF
--- a/extensions/reviewed/FlashTransitionPainter.json
+++ b/extensions/reviewed/FlashTransitionPainter.json
@@ -8,7 +8,7 @@
   "name": "FlashTransitionPainter",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/movie-filter.svg",
   "shortDescription": "Behavior for shape painter allows you to paint a color all over the screen for period of time with an effect (useful for simulate flash and transition effect).",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": [
     "* __Paint effect:__ Action to paint a color all over the screen for a period of time with specific effect.",
     "effect type:",
@@ -72,11 +72,24 @@
                 },
                 {
                   "type": {
-                    "value": "PauseObjectTimer"
+                    "value": "FlashTransitionPainter::FlashTransitionPainter::SetPropertyType"
                   },
                   "parameters": [
                     "Object",
-                    "\"__FlashTransitionPainter_timerEffect\""
+                    "Behavior",
+                    "=",
+                    "\"Flash\""
+                  ]
+                },
+                {
+                  "type": {
+                    "value": "FlashTransitionPainter::FlashTransitionPainter::SetPropertyMaxOpacity"
+                  },
+                  "parameters": [
+                    "Object",
+                    "Behavior",
+                    "=",
+                    "255"
                   ]
                 }
               ]
@@ -114,7 +127,7 @@
                 "textG": 0,
                 "textR": 0
               },
-              "comment": "Avoid default parameter of painter that could make the extension doesn't work."
+              "comment": "Avoid default parameter of painter that could make the extension not work."
             },
             {
               "type": "BuiltinCommonInstructions::Standard",
@@ -148,35 +161,14 @@
                 },
                 {
                   "type": {
-                    "value": "ResetObjectTimer"
+                    "value": "PrimitiveDrawing::UseRelativeCoordinates"
                   },
                   "parameters": [
                     "Object",
-                    "\"__FlashTransitionPainter_timerEffect\""
-                  ]
-                },
-                {
-                  "type": {
-                    "value": "UnPauseObjectTimer"
-                  },
-                  "parameters": [
-                    "Object",
-                    "\"__FlashTransitionPainter_timerEffect\""
+                    ""
                   ]
                 }
               ]
-            },
-            {
-              "type": "BuiltinCommonInstructions::Comment",
-              "color": {
-                "b": 109,
-                "g": 230,
-                "r": 255,
-                "textB": 0,
-                "textG": 0,
-                "textR": 0
-              },
-              "comment": "Initialise position of painter. \nIncrement or decrement \"_TimeProgressionEffect\" depending on direction."
             },
             {
               "type": "BuiltinCommonInstructions::Standard",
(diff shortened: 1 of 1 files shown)


VegeTato:
I took a look and it's amazing !
Forcing the paint to stay at the end is a good idea.
I added few things:
- Added the extension authors (was empty).
- Finished one of the to-do you mentioned (Use the shape painter color and opacity instead of the parameters)

Example including the extension update:
[Transition.zip](https://github.com/GDevelopApp/GDevelop-extensions/files/15051787/Transition.zip)

D8H:
> * Added the extension authors (was empty).

Are you sure it's the same user?

> * Finished one of the to-do you mentioned (Use the shape painter color and opacity instead of the parameters)

When the todo list will be done, we won't do breaking changes and we'll do everything in one update to avoid to deprecate the action several time.
This is why for now I just want your opinion and Tristan's one to be sure what will be done.

VegeTato:
> > * Added the extension authors (was empty).
> 
> Are you sure it's the same user?
> 
> > * Finished one of the to-do you mentioned (Use the shape painter color and opacity instead of the parameters)
> 
> When the todo list will be done, we won't do breaking changes and we'll do everything in one update to avoid to deprecate the action several time. This is why for now, I just want your opinion and Tristan's one so we decide together what will be done.

Oh ok, in that case, everything looks perfect in my opinion.
About the author name, i think so, that's the only registered name with WestBoy

the author GitHub name is **westboy31**
https://github.com/GDevelopApp/GDevelop-extensions/issues/84

tristanbob:
@D8H thanks for your work on this!  I have thoroughly reviewed and tested your example.

## Updates

- Great fixes and update. The events are much simpler.  Your math voodoo magic is powerful!
image

- I see you added two hidden properties: "Index" and "Value".  The purpose of these is not obvious from their name; perhaps we should rename "Index" to "SmoothingIndex"? "Value" seems to be related to "Progress", but I don't know a better name.

- I'm not sure how I feel about disabling the behavior when it's not active. It makes sense to reduce the CPU impact, but I worry about unintended consequences.  It might confuse users if they disable the behavior and then expect their actions to be disabled. Instead, this behavior activates and deactivates itself.  What do you think is right?

## Todo

- Add a property for the smoothing radius

I actually don't like smoothing, so I would set this to 0.  Most of the complexity in the extension is related to this smoothing, and I don't know how much value it adds.

If I were to create new effects to this extension, I'm not sure if I would implement smoothing.  Another consideration is that this would only apply to some effects.  For instance, the Fade effect does not use this property, so perhaps it should be an action parameter instead.

- Use the shape painter color and opacity instead of the parameters

We could offer start and stop values, similar to 

- Use easing

Yes, easings would be a very good addition to this extension
 
- Make 3 actions:
- Fade in
- Fade out
- Fade in and out
- with 2 durations and 2 easing and a pause

The word "fade" only applies to one of the effects.

My primary question is, how can we enable new effects to be easily added to this extension? 
For example, you can see the ones I made:

- https://x.com/VictrisGames/status/1482872044135936000
- https://games.gdevelop-app.com/game-d381d8c7-7d5d-422e-b551-df2e1f47e013/index.html

Perhaps it would work best with a set of actions per effect?  This way each effect can use different parameters in the action.
- Draw one-way transition effect (Parameters: Direction, Duration, Easing, etc) 
- Draw two-way transition effect (Parameters: Start Duration, Start Easing, End duration, End Easing, etc)

D8H:
> I see you added two hidden properties: "Index" and "Value".  The purpose of these is not obvious from their name; perhaps we should rename "Index" to "SmoothingIndex"? "Value" seems to be related to "Progress", but I don't know a better name.

The `Index` is like a local variable. Given that the loops are really simple I don't see the need for a long name. A longer name may make the events harder to read. A bit like when you do a for loop in JS: `for (let i = 0; i < count; i++) {}`.

> I'm not sure how I feel about disabling the behavior when it's not active. It makes sense to reduce the CPU impact, but I worry about unintended consequences.  It might confuse users if they disable the behavior and then expect their actions to be disabled. Instead, this behavior activates and deactivates itself.  What do you think is right?

It's how the extension worked. I want this update not to do too much breaking changes.

I was a bit surprised by this when I looked at the events but I guess users rarely use the "enable/disable behavior" action so I guess it doesn't matter much. We could search in the forum if someone complained about it.

> > * Add a property for the smoothing radius
> 
> I actually don't like smoothing, so I would set this to 0. Most of the complexity in the extension is related to this smoothing, and I don't know how much value it adds.

Looking at the comments, it was to do some kind of anti-aliasing on the circle. Now that there is a property for anti-aliasing on the shape painter, I guess we could remove the feature when we rework the extension and see if users ask for it. 

> > * Use the shape painter color and opacity instead of the parameters
> 
> We could offer start and stop values, similar to how particles and the shockwave extension work.

What use-cases do you see for it? I though this extension was mostly use with black or as a mask.

> The word "fade" only applies to one of the effects.

I see that some YouToube video call Star Wars transitions "wipe". Would "wipe in" and "wipe out" make sense?

> My primary question is, how can we enable new effects to be easily added to this extension? For example, you can see the ones I made:
> 
> * https://twitter.com/VictrisGames/status/1483220151889629191
> 
> * https://games.gdevelop-app.com/game-1fb34c7b-1093-4c93-a253-17187b4e540f/index.html
> 
> 
> Perhaps it would work best with a set of actions per effect? This way each effect can use different parameters in the action.
> 
>  * Draw one-way transition effect (Parameters: Direction, Duration, Easing, etc)
> 
> * Draw two-way transition effect (Parameters: Start Duration, Start Easing, End duration, End Easing, etc)

Yes, I think it makes sense. So to avoid too much actions, I guess we would keep a parameter to invert the transition. I guess we could make the action asynchronous. This way users can duplicate the action to make an out and in and even add a "wait" in the middle.

PR #1272, opened 2024-05-06
[Reviewed] [Boids Movement] Fix typo of VelocityX and VelocityY being swapped

tristanbob:
Should also fix potential problems with "avoid object" action by setting the correct TrajectorySpeed.

### DIFF
--- a/extensions/reviewed/BoidsMovement.json
+++ b/extensions/reviewed/BoidsMovement.json
@@ -8,7 +8,7 @@
   "name": "BoidsMovement",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Glyphster Pack/Master/SVG/Restaurant/Restaurant_restaurant_seafood_animal_fish.svg",
   "shortDescription": "Simulates flocks movement.",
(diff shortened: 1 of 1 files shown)


D8H:
It looks like only the labels and descriptions are inverted. Please, modify these instead of renaming the functions as it probably break a formula.

> Should also fix potential problems with "avoid object" action by setting the correct TrajectorySpeed.

Can you explain?

tristanbob:
> It looks like only the labels and descriptions are inverted. Please, modify these instead of renaming the functions as it probably breaks a formula.

Sure, I re-installed the previous version and changed the name and descriptions (instead of editing the functions).

> > Should also fix potential problems with "avoid object" action by setting the correct TrajectorySpeed.
> 
> Can you explain?

The VelocityX and VelocityY expressions are used by the "AvoidPosition" action.  I thought that perhaps these were used in reverse, but I don't understand the function.  If this looks correct to you, then you can ignore my comment about this.

image

D8H (approved):
Thank for this fix!

PR #1281, opened 2024-05-10
[Community] [Point and Orbit] Add the ability to point towards an angle (useful for gamepads)

tristanbob:
- Add the ability to point towards an angle (useful for gamepads)
- Converted events to use the new method to reference parameters directly

https://github.com/GDevelopApp/GDevelop-extensions/assets/8879811/91a5005d-15f2-4f0a-9456-46bec9514389

### DIFF
--- a/extensions/community/PointAndOrbit.json
+++ b/extensions/community/PointAndOrbit.json
@@ -7,15 +7,15 @@
   "iconUrl": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz48IURPQ1RZUEUgc3ZnIFBVQkxJQyAiLS8vVzNDLy9EVEQgU1ZHIDEuMS8vRU4iICJodHRwOi8vd3d3LnczLm9yZy9HcmFwaGljcy9TVkcvMS4xL0RURC9zdmcxMS5kdGQiPjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0ibWRpLWFycm93LWJvdHRvbS1yaWdodC1ib2xkLW91dGxpbmUiIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNOS44OCwxNS41NEw0LjIyLDkuODhMOS44OCw0LjIyTDE1LjU0LDkuODhMMTkuNzgsNS42NFYxOS43OEg1LjY0TDkuODgsMTUuNTRNMTcuNjYsMTAuNTlMMTUuNTQsMTIuNzFMOS44OCw3LjA1TDcuMDUsOS44OEwxMi43MSwxNS41NEwxMC41OSwxNy42NkgxNy42NlYxMC41OVoiIC8+PC9zdmc+",
   "name": "PointAndOrbit",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/arrow-bottom-right-bold-outline.svg",
-  "shortDescription": "Point an object toward a position or angle while orbiting that object around a position.",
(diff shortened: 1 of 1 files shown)


extensions/community/PointAndOrbit.json
```diff
@@ -1,15 +1,29 @@
...
   "helpPath": "",
   "iconUrl": "data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz48IURPQ1RZUEUgc3ZnIFBVQkxJQyAiLS8vVzNDLy9EVEQgU1ZHIDEuMS8vRU4iICJodHRwOi8vd3d3LnczLm9yZy9HcmFwaGljcy9TVkcvMS4xL0RURC9zdmcxMS5kdGQiPjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4bGluaz0iaHR0cDovL3d3dy53My5vcmcvMTk5OS94bGluayIgdmVyc2lvbj0iMS4xIiBpZD0ibWRpLWFycm93LWJvdHRvbS1yaWdodC1ib2xkLW91dGxpbmUiIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNOS44OCwxNS41NEw0LjIyLDkuODhMOS44OCw0LjIyTDE1LjU0LDkuODhMMTkuNzgsNS42NFYxOS43OEg1LjY0TDkuODgsMTUuNTRNMTcuNjYsMTAuNTlMMTUuNTQsMTIuNzFMOS44OCw3LjA1TDcuMDUsOS44OEwxMi43MSwxNS41NEwxMC41OSwxNy42NkgxNy42NlYxMC41OVoiIC8+PC9zdmc+",
   "name": "PointAndOrbit",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/arrow-bottom-right-bold-outline.svg",
-  "shortDescription": "Point an object toward a position and Orbit around a position.",
-  "version": "1.0.0",
+  "shortDescription": "Point an object toward a position or angle while orbiting that object around a position.",
+  "version": "1.1.0",
+  "description": [
+    "# User can",
```
D8H, line 13: The wiki uses `#` for page title. Titles in description should be `##`, but "User can" will look a bit silly in a summary. I think it's better to not use titles in descriptions.

Also, either use "you can" or "It allows" instead of "user can". You won't see "user" anywhere in the wiki (if you do, please report it).

D8H:
Why would someone make the object rotate at a given speed when it moves instantly. I guess it would look weird, wouldn't it?

tristanbob:
> Why would someone make the object rotate at a given speed when it moves instantly. I guess it would look weird, wouldn't it?

Good point! I copied your math from the Turret extension and got this working. Thanks for catching this oversight!
I also updated the wording of the extension.

Err... do you think we should merge this into the Turret extension?  
It has some similarities, but I guess turrets usually don't orbit another object.

https://github.com/GDevelopApp/GDevelop-extensions/assets/8879811/76fee3dd-0de4-4e20-905c-c58233b4bbbb

D8H:
> Good point! I copied your math from the Turret extension and got this working. Thanks for catching this oversight!
> I also updated the wording of the extension.

Sorry, my point was more about removing the speed parameter. To me, this extension purpose is just to squash "put around" and "change angle" to avoid to fill the direction 2 times.

I'm not prod of what I've done for the Turret extension. It will need some complete rework.

tristanbob:
@D8H do you mind if I keep this "speed" feature?  I think it is a fun and easy mechanism that could be useful in games.

Any other thoughts or comments before I merge this?

D8H:
I think it's over-complexifying the extension for maybe no value.

What uses-cases do you see for it?

D8H:
ok... nice discussing with you too... :confused:

extensions/community/PointAndOrbit.json
```diff
@@ -62,73 +77,188 @@
...
+        {
+          "description": "Angle",
+          "longDescription": "Point the gun object to this angle",
+          "name": "TargetAngle",
+          "type": "expression"
+        },
+        {
+          "description": "Pointing speed (in deg/second)",
+          "longDescription": "Set to 0 for an immediate pointing.",
+          "name": "PointSpeed",
```
D8H, line 261: Is this handled by your change?

PR #1284, opened 2024-05-11
[CrazyGame] Fix the action to display a video ads

Bouh:
Fix the action to display video ads, now a error is returned if the ad isn't ready (Thanks to MartijnKuilema)


extensions/reviewed/CrazyGamesAdApi.json
```diff
@@ -349,6 +349,7 @@
                 "    },",
                 "    adError: (error, errorData) => {",
                 "        gdjs._crazyGamesExtension.isPlayingVideoAd = false;",
+                "        gdjs._crazyGamesExtension.hasVideoAdJustFinishedPlaying = true;",
```
D8H, line 352: Is it to ensure that games won't be stuck if they forgot to handle errors?
Should there be a way for creator to still check if it has finished because of an error?

I guess it's to comply to this:

> Often, the request for a midgame ad will be unfilled (either because of timing restrictions, adblock, or low demand). In this case you receive an adError event. You should handle this case correctly and ensure that the game continues.

PR #1286, opened 2024-05-13
Update BoidsMovement.json

planktonfun:
Issue reference: https://forum.gdevelop.io/t/boids-extension-doesnt-work-anymore-after-deactivate-it/57526/3

Update 0.2.3:
- fix activate/deactivate behavior not working.

### DIFF
--- a/extensions/reviewed/BoidsMovement.json
+++ b/extensions/reviewed/BoidsMovement.json
@@ -8,11 +8,14 @@
   "name": "BoidsMovement",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Glyphster Pack/Master/SVG/Restaurant/Restaurant_restaurant_seafood_animal_fish.svg",
   "shortDescription": "Simulates flocks movement.",
-  "version": "0.2.2",
+  "version": "0.2.3",
   "description": [
     "Simulates swarms or flocks movement following the separation, alignment, cohesion principles. The flock can be attracted to a location or avoid some obstacles.",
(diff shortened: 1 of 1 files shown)


Kruger77:
Thanks for the fix 🙏

extensions/reviewed/BoidsMovement.json
```diff
@@ -743,7 +753,7 @@
                 "const object = objects[0];",
                 "const behaviorName = eventsFunctionContext.getBehaviorName(\"Behavior\");",
                 "const behavior = object.getBehavior(behaviorName);",
-                "runtimeScene.__boidsExtension.boidsManager.remove(behavior);"
+                "behavior.activated = 0;"
```
D8H, line 756: By removing this line, deactivated Boids will stays in the R-Tree and will continue to have side effect on other Boids.
A behavior already have a `activated()` accessor, but it's not needed. What should probably be done is to remove the Boids from `boids`.

D8H (changes requested):
Thank you for submitting an update.

extensions/reviewed/BoidsMovement.json
planktonfun: that's the error, by removing the boid from the R-tree it forgets its original behavior when reactivated, so my solution is to pause and then resume it when needed

D8H:
Is the issue still there in the 0.3.x?
- https://github.com/GDevelopApp/GDevelop-extensions/pull/1367

PR #1287, opened 2024-05-13
[Community] [WithThreeJs] Supports shadow

PANDAKO-GitHub:
Supports real-time shadow feature, etc.


extensions/community/WithThreeJS.json
```diff
@@ -8,7 +8,7 @@
   "name": "WithThreeJS",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/rotate-3d.svg",
   "shortDescription": "Create a simple 3D scene.",
-  "version": "1.2.2",
+  "version": "2.0.0",
```
D8H, line 11: When you increment the major, you need to explain to users what the breaking changes are and how they should modify their event to make their project work with the new version of the extension.

D8H (changes requested):

extensions/community/WithThreeJS.json
tristanbob: @PANDAKO-GitHub did you make this 2.0 because it causes breaking changes to existing games?  If so, Davy is asking to describe this in the description.  I think that is the only hold up.

PANDAKO-GitHub: Thank you for your review.
I have updated the version with optimizations for 3D object deletion and added explanations about compatibility.

D8H (approved):

D8H:
Thank you for this new update!

PR #1323, opened 2024-06-25
[Reviewed] [Screen shake] Fix camera Y drift

D8H:


tristanbob:
Thanks for this fix!  I can confirm it works much better now.  :)

PR #1363, opened 2024-08-29
[Community] [TweenZOrder] Support for GDevelop 5.4.x.

PANDAKO-GitHub:
- Support for GDevelop 5.4.x.  
- Changed duration unit from milliseconds to seconds.


D8H:
Thank you for the update.

I don't understand what this extension can be used for. An object is either in front or behind another object so I don't see how interpolation can help.
Can you give some use-cases in the description of the extension?

PANDAKO-GitHub:
Thanks for checking.
As you pointed out from the beginning, this extension is often used together with my 3D extension.
I understand that you don't want to add non-generic extensions to the community extensions.
If you think this extension is not suitable, feel free to remove it from the list. 😉
If not, please update this extension.

D8H:
If its only usage is for WithTreeJS extension users, it should be integrated to WithTreeJS directly.
If so, please update WithTreeJS and remove this one.

PANDAKO-GitHub:
How do I remove a community extension?
Do I just delete the files?

PR #1397, opened 2024-10-04
[Community] [WithThreeJs] Added drawing detection and improved raycasting.

PANDAKO-GitHub:
- Added "Check 3D Object Is Within 3D Camera FOV" condition.
- Added "Change 3D Object Raycast Detection" action.


D8H:
Thank you for this updated.

Does it include the Z-order tween action? Can the Z-order tween extension be removed?

PR #1418, opened 2024-10-30
[Community] [TweenZOrder] Remove the extension

PANDAKO-GitHub:
This has been removed as it is not generic.


D8H:
Is this feature now included in WithThreeJS directly?
I'm asking to write a migration note in the change log.

PANDAKO-GitHub:
No, I don't plan to add this functionality to WithThreeJS.
Since it's not used that often and there are alternatives, I'll distribute it on my website from now on.

PR #1421, opened 2024-10-31
[Community] [Navigation mesh pathfinding] Fixed problem with Rotation speed not being applied correctly.

PANDAKO-GitHub:
This PR includes fixes and suggestions.

- Fixed an issue where rotation speed was not being applied correctly.
- Added units to properties.
- Set actions to private that change parameters that cannot be changed later, such as "Change the Cell size."
- Added a note to the description of the "Draw navigation mesh" action.

@D8H , What do you think about these changes?

~~
This extension is great in that it works when added to a new tile map object.
It would be better if the navigation mesh could be updated when tiles are changed.

This current solution is not very smart.
image

### DIFF
--- a/extensions/community/NavMeshPathfinding.json
+++ b/extensions/community/NavMeshPathfinding.json
@@ -8,7 +8,7 @@
   "name": "NavMeshPathfinding",
   "previewIconUrl": "https://resources.gdevelop-app.com/assets/Icons/Line Hero Pack/Master/SVG/Maps and Navigation/Maps and Navigation_map_find_search.svg",
   "shortDescription": "Pathfinding allows to compute an efficient path for objects, avoiding obstacles on the way.",
-  "version": "0.2.3",
+  "version": "0.2.4",
   "description": [
     "In comparison to the built-in pathfinding behavior, this one aims to:",
     "- better respect obstacle shapes",
@@ -4368,7 +4368,7 @@
             "        var isometricRatio = manager.configuration._getIsometricRatio();",
             "        var owner = this.behavior.owner;",
             "        var angleOffset = this.behavior._getAngleOffset();",
-            "        var angularMaxSpeed = this.behavior._getMaxSpeed();",
+            "        var angularMaxSpeed = this.behavior._getAngularMaxSpeed();",
             "        var rotateObject = this.behavior._getRotateObject();",
             "        var timeDelta = owner.getElapsedTime(instanceContainer) / 1000;",
             "        this.pathFollower.step(timeDelta);",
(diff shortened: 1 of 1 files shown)


D8H:
> Set actions to private that change parameters that cannot be changed later, such as "Change the Cell size."

What make you think that it can't be changed?
The navmesh should update as soon as a new path is asked.

> It would be better if the navigation mesh could be updated when tiles are changed.

It watches the position and size. It would need a change in GDveleop to subscribe to tiles changes.

PANDAKO-GitHub:
> What make you think that it can't be changed?

Because I tried it and it didn't work.
Is this usage incorrect?
image

> It would need a change in GDveleop to be able to subscribe to tiles changes.

I think this problem can be solved just by adding an action to update the navigation mesh.

D8H:
> Because I tried it and it didn't work. Is this usage incorrect?

It should work, but there will be only 2x2 cells.

PANDAKO-GitHub:
> It should work, but there will be only 4 rastering cells. It won't be usable.
Does that mean the action has no effect?

D8H:
> > It should work, but there will be only 4 rastering cells. It won't be usable.
> 
> Does that mean the action has no effect?

I meant that it's the right actions, but the parameter values are off.
Did you choose random values just to take a screenshot?
What are the real values you used for your game?

PANDAKO-GitHub:
> What are the real values you used for your game?

I set the cell size to 10 in the scene properties and tried changing it to 16 in the actions, but the mesh didn't change.

D8H:
> I set the cell size to 10 in the scene properties and tried changing it to 16 in the actions, but the mesh didn't change.

It won't change unless you ask a new path and the changes will be very small.

PANDAKO-GitHub:
Even if the cell size is the same, the result will vary depending on the process of setting.





Events:


And just to be sure, I will tell you that the action to **change the Area bound** has no effect.

extensions/community/NavMeshPathfinding.json
```diff
@@ -5312,7 +5312,7 @@
           "objectGroups": []
         },
         {
-          "description": "Draw the navigation mesh used for the object.",
+          "description": "Draw the navigation mesh used for the object.\nThe \"Move to a position\" action must have been performed at least once before this action is executed.",
```
D8H, line 5315:
```suggestion
          "description": "Draw the navigation mesh used for the object.\nThis action must be used after \"Move to a position\".",
```

D8H:
> And just to be sure, I will tell you that the action to **change the Area bound** has no effect.

Ok, I guess it make sense not being able to change the area and cell size so hiding the actions seems to indeed be the right call.

Thank you for the fixes.

PR #1488, opened 2025-01-09
[Reviewed] [Two choices dialog box] Simplify the events

D8H:
### Changes
- Simplify the events
- Hide instructions and properties about button id because users can't know what it means.
  - if the feature is ask, it should probably use user-defined button labels. 

### Demo
- https://github.com/GDevelopApp/GDevelop-examples/pull/746


Bouh (approved):

Bouh:
doStepPostsEvents function is indeed simplier, thanks.

PR #1578, opened 2025-03-10
Fix typo in MarchingSquares.json

Bouh:
image


D8H:
It's better to increment the version even for typo because otherwise nobody knows there is a new version and someone may submit another update that reverts this change without realizing it (diff can be hard to check for extensions).

Bouh:
Got it, I'll keep this in mind for the next time 👍

PR #1622, opened 2025-03-21
Update a description to make it more meaningful

Bouh:
This lack was highlighted in user tests.


extensions/reviewed/PhysicsCar.json
```diff
@@ -2260,7 +2260,7 @@
           "type": "Number",
           "unit": "PixelAcceleration",
           "label": "Acceleration",
-          "description": "",
+          "description": "Object speed",
```
ClementPasteau, line 2263: but acceleration is not speed, that description is misleading in my opinion.
it should be more "How fast the object increases its speed"?

PR #1830, opened 2025-09-15
Fixed hasString and hasNumber for global and object arrays

manisto:
The `HasNumber` and `HasString` conditions for objects would always return true, regardless of whether the array contained the specified value or not. The reason was that the functions referenced a nonexistent variable named `__ArrayTools.Temp`.

This PR removes all occurrences of this prefix, and now uses the extension scene variable simply named `Temp`. 

An example project where the issue is documented has been attached to this PR.
[ArrayToolsTest.zip](https://github.com/user-attachments/files/22350566/ArrayToolsTest.zip)

### DIFF
--- a/extensions/reviewed/ArrayTools.json
+++ b/extensions/reviewed/ArrayTools.json
@@ -45,7 +45,8 @@
     "insert"
   ],
   "authorIds": [
-    "ZgrsWuRTAkXgeuPV9bo0zuEcA2w1"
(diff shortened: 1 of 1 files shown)


D8H:
Thank you for the fix. Please increment the revision in the version.

D8H:
Someone reported the issue and I forgot about this PR. I guess it's already fixed.
- https://github.com/GDevelopApp/GDevelop-extensions/issues/1867

D8H:
It’s an action name. The one without the `2` is deprecated.

PR #2026, opened 2026-02-10
Fix the case of the `GetURLParam` expression

D8H:


erikdubbelboer:
@4ian good idea, I'll make sure it gets included in there soon.

PR #2151, opened 2026-06-07
Update NavMeshPathfinding.json

Bouh:
Cache grid.dimX() and grid.dimY() calls into local variables (dimX, dimY, maxX, maxY) before nested loops, instead of recomputing them on every iteration. 
That reduces redundant function call overhead during pathfinding computation.


D8H:

Please, share profiling results that show the efficiency impact.

PR #2170, opened 2026-06-25
Fix InUseAttributeError in pointer lock

PANDAKO-GitHub:
Track pending state with a self-managed flag since pointerLockElement is null while a request is in flight. Catch the returned Promise to avoid the unhandled rejection.


extensions/reviewed/MousePointerLock.json
```diff
@@ -68,6 +69,7 @@
       "sentence": "",
       "events": [
         {
+          "disabled": true,
```
D8H: Please remove the duplicated JS event. It forbids to see what's changed.

PANDAKO-GitHub:
Hi @D8H, I’ve removed the duplicated JS event as requested in commit 949dca0. Could you please take another look when you have a chance? Thanks!

PR #2237, opened 2026-08-22
Update lights to be capped per light type and avoid shader recompilations

4ian:
See https://github.com/4ian/GDevelop/pull/8998/changes - this allows a bit more lights of different types (ensuring they don't go through the budget of the device) and also reduce the "churn" in shaders being forced to be recompiled because the amount of lights changed


extensions/reviewed/Light3D.json
```diff
@@ -10,7 +10,7 @@
   "name": "Light3D",
   "previewIconUrl": "https://asset-resources.gdevelop.io/public-resources/Icons/f237ae4e3b857c556846c7b2c0b132556fd4bcdeff217034b4d9c97dc1aab1d6_lightbulb-on-outline.svg",
   "shortDescription": "A collection of light object for 3D.",
-  "version": "1.0.2",
+  "version": "1.2.1",
```
D8H: It bumps the version at each commit.

4ian:
Summary of the whole change to 3D lights:

- Lights are now budgeted **per light kind** (point lights, spot lights, spot lights projecting a texture) instead of a single weighted total, so the numbers three.js compiles into its shaders stay stable and the mix of visible lights no longer forces shader recompilations (a stutter the player feels).
- The budget is **derived from the device's real WebGL capabilities** (texture units, varyings, uniform vectors) and shared between the kinds a scene actually uses — going past what the device can take previously made shaders fail to link, rendering everything black.
- The nearest lights and shadow casters are selected by distance to the camera; when a new light kind appears and the budget shrinks, lights that no longer fit are evicted immediately so a frame never renders over budget.
- The "max lights count" settings and descriptions now reflect that the caps apply per light kind, with the device enforcing a lower limit when needed.

---
_Generated by [Claude Code](https://claude.ai/code)_

